PR #31, opened 2016-01-13
Fix empty result sets

medvednikov:


medvednikov:
https://github.com/siddontang/go-mysql/issues/30

PR #105, opened 2017-01-06
support TLS/SSL authentication to the database

tylerkovacs:
This commit updates mysql client authentication to (optionally) support
TLS/SSL connections to the database. To activate, provide tls.Config
when calling client.Connect.

Changes to client/auth.go were copied from the official go-sql-driver.
https://github.com/go-sql-driver/mysql/blob/master/packets.go#L226

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1,8 +1,10 @@
 package client
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
+	"strings"
 	"testing"
 
 	. "github.com/pingcap/check"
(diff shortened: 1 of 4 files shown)


siddontang:
Thanks @tylerkovacs 

I think we can add a test with go-sql-driver :-)

tylerkovacs:
@siddontang `go-mysql-driver` has a single test case related to TLS (`TestTLS`) which verifies that a TLS connection can be made to the test database. 

I added a test to this PR that verifies that the provided tls.Config is used when connecting to the database. It verifies that the connection fails when the provided tls.Config is incorrect, however I'm having trouble writing a test for a successful connection.

`go-mysql-driver` does not appear to provide any mysql configuration to support TLS in CI. However, attempts to connect via TLS using the `go-mysql` instance of mysql in CI fail (with `unexpected EOF`). I'm not sure how to configure the CI instance of mysql to support TLS connections.

Do you think the test in this PR is sufficient since it proves that the provided tls.Config is used? If not, do you have any insight on why the travis instances of mysql for `go-mysql-driver` and `go-mysql` behave differently?

client/client_test.go
```diff
@@ -76,6 +78,23 @@ func (s *clientTestSuite) TestConn_Ping(c *C) {
...
 }
 
+func (s *clientTestSuite) TestConn_TLS(c *C) {
+	// Verify that the provided tls.Config is used when attempting to connect to mysql.
+	// An empty tls.Config will result in a connection error.
+	addr := fmt.Sprintf("%s:%d", *testHost, *testPort)
+	_, err := Connect(addr, *testUser, *testPassword, *testDB, func(c *Conn) {
+		c.TLSConfig = &tls.Config{}
+	})
+	if err == nil {
```
siddontang: We can use c.Assert(err, NotNil)

siddontang:
Hi @tylerkovacs 

> However, attempts to connect via TLS using the go-mysql instance of mysql in CI fail 

Can we use go-mysql to connect MySQL with TLS now? If we can, I think it is ok here.

siddontang:
Hi @tylerkovacs 

Any update? Is it ok to merge it now?

tylerkovacs:
Yes, I believe it is safe to merge. We've been TLS connections with this
code successfully since submitting this pull request.

tylerkovacs:
Thank you!

PR #168, opened 2017-09-26
BUGFIX: fix data race from closing canal

taylorchu:


taylorchu:
https://github.com/siddontang/go-mysql/issues/166

@siddontang

PR #372, opened 2019-03-08
add compatibility to json decode for generated column

amyangfei:
fix bug found in https://github.com/pingcap/tidb/issues/8761 , which is similar to a MySQL bug: https://bugs.mysql.com/bug.php?id=88791

add a switch to control whether we will ignore this error during jsonBinaryDecoder decode

### Side effects
When we decode a json binary and meet a `isDataShort` error, we don't have enough context to distinguish whether it is caused by this MySQL bug, so we may miss an unexpected error and return an invalid decode result.
However before we decode a json binary, the length of this value is known in advance from code [replication/row_event.go#L517](https://github.com/siddontang/go-mysql/blob/master/replication/row_event.go#L517) and we can believe the `isDataShort` error is definitely a MySQL bug or binlog data inconsistent.

### Tests 
 - Unit test

### DIFF
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -76,7 +76,10 @@ func (e *RowsEvent) decodeJsonBinary(data []byte) ([]byte, error) {
 	if len(data) == 0 {
 		return []byte{}, nil
 	}
-	d := jsonBinaryDecoder{useDecimal: e.useDecimal}
+	d := jsonBinaryDecoder{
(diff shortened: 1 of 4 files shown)


amyangfei:
PTAL @siddontang @GregoryIan

GregoryIan:
LGTM

GregoryIan:
@siddontang PTAL

PR #394, opened 2019-06-17
canal: skip query  event if parse failed

WangXiangUSTC:
fix issue #393 

will skip event if parse failed


WangXiangUSTC:
@GregoryIan PTAL

PR #409, opened 2019-08-15
replication: support datetime less than unix timestamp zero

amyangfei:
## What problem does this PR solve?
Golang standard library doesn't support datetime before `1970-01-01 00:00:00` very well, for example the following code snippet will get result
```
package main

import (
    "fmt"
    "time"
)

func main() {
    var (
        year   = 0
        month  = 0
        day    = 1
        hour   = 21
        minute = 30
        second = 42
    )
    t := time.Date(year, time.Month(month), day, hour, minute, second, 0, time.UTC)
    fmt.Printf("%s\n%d\n", t, t.Unix())
}
```

```
-0001-12-01 21:30:42 +0000 UTC
-62169820158
```
This will affect datetime type replication.

## What is changed and how it works?

Return the string format of a datetime directly if it is before `1970-01-01 00:00:00`

## Side effects
We may lose time zone conversion for datetime before `1970-01-01 00:00:00`. But the zero datetime (`0000-00-00 00:00:00`) still has this issue before this pr. Users should take their own risks to use these datetime values if upstream and downstream have different timezones. 

## Tests
Unit test

### DIFF
--- a/replication/row_event.go
+++ b/replication/row_event.go
@@ -754,6 +754,19 @@ func decodeDatetime2(data []byte, dec uint16) (interface{}, int, error) {
 	minute := int((hms >> 6) % (1 << 6))
 	hour := int((hms >> 12))
 
+	// DATETIME encoding for nonfractional part after MySQL 5.6.4
+	// https://dev.mysql.com/doc/internals/en/date-and-time-data-type-representation.html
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @siddontang @GregoryIan

PR #412, opened 2019-08-22
fix(canal): fix handleUnsigned use wrong col index

stutiredboy:
`i` is r.Rows index, not  r.Table.Columns index.


stutiredboy:
duplicate issue of https://github.com/siddontang/go-mysql/pull/405

PR #416, opened 2019-09-01
hotfix: fix saved gtid in replication

GregoryIan:
only save GTID when meeting XID and QUERY event. 

it duplicates with #414, but we really very need this fix, so after waiting three days, I decide to file a hotfix pr for it. Thanks  @ianzapolsky very much

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -707,6 +707,11 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
 	//skip OK byte, 0x00
 	data = data[1:]
 
+	var (
+		currentGTID string
+		err         error
+	)
+
 	needACK := false
 	if b.cfg.SemiSyncEnabled && (data[0] == SemiSyncIndicator) {
 		needACK = (data[1] == 0x01)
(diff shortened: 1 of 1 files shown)


GregoryIan:
@amyangfei PTAL

replication/binlogsyncer.go
```diff
@@ -707,6 +707,11 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
 	//skip OK byte, 0x00
 	data = data[1:]
 
+	var (
+		currentGTID string
```
amyangfei: should this value be cached in `BinlogSyncer` to be used in the following binlog event?

replication/binlogsyncer.go
```diff
@@ -729,27 +734,18 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
...
-			break
-		}
 		GTID := event.GTID
-		err := b.gset.Update(fmt.Sprintf("%d-%d-%d", GTID.DomainID, GTID.ServerID, GTID.SequenceNumber))
-		if err != nil {
-			return errors.Trace(err)
-		}
+		currentGTID = fmt.Sprintf("%d-%d-%d", GTID.DomainID, GTID.ServerID, GTID.SequenceNumber)
 	case *XIDEvent:
-		event.GSet = b.getGtidSet()
```
amyangfei: we can remove `getGtidSet` function, as it is not used any more

replication/binlogsyncer.go
GregoryIan: I think no need, it always retry from a gtid

amyangfei: but `currentGTID` is in the function `parseEvents` scope, `b.gset` is not correctly updated then.

GregoryIan: I dont know what's meaning of you

GregoryIan: You can give a binlog event temporal logic example

amyangfei: Does code `event.GSet, err = b.updateGtidSet(currentGTID)` in L743 and L745 want to update with the `currentGTID` that has the `gtid` value right in the same transaction?
Such as when processing the Query event for DDL `create database sharding1`, `b.gset` should be updated with `9afe121c-40c2-11e9-9ec7-0242ac110002:3`
```
| bin.000001 | 522 | Gtid           |       101 |         587 | SET @@SESSION.GTID_NEXT= '9afe121c-40c2-11e9-9ec7-0242ac110002:3'                                                                                                        |
| bin.000001 | 587 | Query          |       101 |         698 | create database `sharding1`
```

replication/binlogsyncer.go
```diff
@@ -729,27 +734,18 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
...
+		event.GSet, err = b.updateGtidSet(currentGTID)
 	case *QueryEvent:
-		event.GSet = b.getGtidSet()
+		event.GSet, err = b.updateGtidSet(currentGTID)
```
amyangfei: does this still have a problem with the `Query` event Begin in DML event sequences like following
```
| bin.000001 | 16597 | Gtid           |       101 |       16662 | SET @@SESSION.GTID_NEXT= '9afe121c-40c2-11e9-9ec7-0242ac110002:55'                                                                                                                                                                                                                                                                                                                                        |
| bin.000001 | 16662 | Query          |       101 |       16730 | BEGIN                                                                                                                                                                                                                                                                                                                                                                                                     |
| bin.000001 | 16730 | Table_map      |       101 |       16791 | table_id: 11993 (sharding1.t2)                                                                                                                                                                                                                                                                                                                                                                            |
| bin.000001 | 16791 | Update_rows    |       101 |       17277 | table_id: 11993 flags: STMT_END_F                                                                                                                                                                                                                                                                                                                                                                         |
| bin.000001 | 17277 | Xid            |       101 |       17308 | COMMIT /* xid=51083 */
```

replication/binlogsyncer.go
GregoryIan: Yep, then would it only receives a partial transaction?

GregoryIan:
the most correct way seems to be to not try again here, because we can't distinguish when we should save gtid, it's better to let caller handle it

ianzapolsky:
@gleonid has a good solution to this in #420. Let's go with that.

PR #421, opened 2019-09-04
Reduce memory allocation at several palces.

haosdent:


packet/conn.go
```diff
@@ -24,28 +48,35 @@ type Conn struct {
...
 	} else {
-		return buf.Bytes(), nil
+		result := make([]byte, buf.Len())
+		copy(result, buf.Bytes())
+		return result, nil
 	}
 }
 
-func (c *Conn) ReadPacketTo(w io.Writer) error {
+func (c *Conn) ReadPacketTo(w *bytes.Buffer) error {
```
siddontang: I prefer not breaking compatibility here, we can try to convert the Writer to Buffer and do the Grow below.

haosdent: Fixed

haosdent:
Without this change

image

After this change

image

PR #422, opened 2019-09-04
Wrap conn with bufio.Reader for normal case.

haosdent:
`bufio.Reader` is removed at 8b471135 previously. Because it improves
throughput a lot, we add it back in this change. But it only effective
for normal case. For TLS connection, we still read from `net.Conn`
directly.


siddontang:
please fix the CI 

@haosdent

haosdent:
Before this change

image


After this change

image

haosdent:
Looks still failed, I would take a look later.

haosdent:
Thx siddon and rim

PR #431, opened 2019-10-08
replication: support parsing event with header only and no checksum

amyangfei:

When parsing a binlog event that has event header only and no checksum data, go-mysql will return error `invalid event header, event size is 19, too small`

Ref to [MySQL doc](https://dev.mysql.com/doc/internals/en/stop-event.html), `STOP_EVENT` has event header only, so if binlog checksum is not enabled, there may exist binlog event with event size 19.

Besides the binlog event checksum [was disabled by default before MariaDB 10.2.1](https://mariadb.com/kb/en/library/binlog-event-checksums/), so this problem occurs often in MariaDB 10.1.x

Allow binlog event size equals to 19


amyangfei:
PTAL @siddontang @GregoryIan @csuzhangxc

PR #522, opened 2020-10-16
UUID clone

lance6716:
speed up GTID clone by use slice copy in `UUIDSet.Clone`. when application (TiDB DM, in my case) often call GTID clone, this helped a lot.

Benchmark shows 1166 ns/op -> 620 ns/op


lance6716:
@atercattus PTAL

atercattus (approved):

PR #524, opened 2020-11-10
ability to send multiple resultsets from handleQuery

GoWebProd:
For example, if query in HandleQuery equal to:

```
SELECT 0 UNION SELECT 1;
SELECT 1 AS col1, 2 AS col2, 3 AS col3 UNION SELECT 4, 5, 6;
```

we need to return two result sets, but current realisation not allows to make it


atercattus (approved):

skoef:
I would love to see more of this PR. While I don't get how this got approved in the first place (some parts are completely irrelevant to this PR and the new function `ChainResultSet` is not even used), but I implemented something similar myself as well, which I like to either convert to something that could be merged upstream or could convert to whatever this PR could lead to.

Especially reading the multiple results from the wire into a chained resultset is something I would like to wrap my head around!

dveeden:
@GoWebProd could you rebase this on master and/or resolve the conflicts?

dveeden:
Multiple resultsets can also happen when you call stored procedures.

PR #527, opened 2020-11-19
Fix fetchIndexesViaSqlDB/NewTableFromSqlDB for MySQL 8.0

RobinGeuze:
MySQL 8 has a different column count for SHOW INDEX FROM, causing issues with the use of Scan in fetchIndexesViaSqlDB


lance6716 (approved):

lance6716:
ptal @atercattus

atercattus (approved):

PR #569, opened 2021-04-13
Update method now may receive strings with GTID sets

mialinx:


lance6716:
Hi, I think your PR will enable supporting multiple GTID update, but will it happen in practice? Every binlog event should contains only one GTID

mialinx:
> Every binlog event should contains only one GTID

Hi!  

Sure, but GTIDSet types may be used for  general GTID calculations.
For instance, we may calculate full GTID stored on the replica as  `Retrieved_Gtid_Set + Executed_Gtid_Set`
Both may contain multiple GTIDs from different sources.

P.S. Actually, I'm going to open source for mysql HA tool we use in cloud.yandex.ru, it uses patched go-mysql for GTID calculation. So I thought it would be right to publish this patch and use upstream version of go-mysql.

lance6716 (approved):

PR #579, opened 2021-05-11
Parse Transaction Sql Should Continue It

HZMarico:
parse query(SAVEPOINT `trans2`) err line 1 column 9 near "SAVEPOINT `trans2`" , will skip this event

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -2,6 +2,7 @@ package canal
 
 import (
 	"fmt"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lance6716:
Hi, this project uses parser from https://github.com/pingcap/parser

And since SAVEPOINT is a valid mysql syntax https://dev.mysql.com/doc/refman/8.0/en/savepoint.html, would you like to contribute to the parser?

dveeden:
Should we close this? If not, what is the exact problem? Is `SAVEPOINT` breaking things? Is it causing logs to grow?

lance6716:
SAVEPOINT is implemented in upstream parser package https://github.com/pingcap/tidb/pull/34466 and we have upgraded the parser dependency version. The problem should be solved

PR #581, opened 2021-05-19
don't os.Exit(1) when meet an error

lance6716:
When upstream is not MySQL binlog compatible, return an error gracefully

reference: https://github.com/pingcap/dm/issues/1682#issuecomment-843891302


Ehco1996:
lgtm

lichunzhu (approved):
LGTM

PR #588, opened 2021-06-16
expose capability and charset of connections to server

skoef:
For a project of mine I need to be able to read the capabilities and default charset of a client connecting to my server. This PR exposes those 2 properties. Charset perhaps used to be decoded before into a type called `CollationId` but this doesn't exist (anymore) and I'm perfectly fine with the collation ID as an uint8.

I also took the liberty to refactor `readFirstPacket` a bit so it is easier to test its behaviour

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -17,6 +17,7 @@ type Conn struct {
 
 	serverConf     *Server
 	capability     uint32
+	charset        uint8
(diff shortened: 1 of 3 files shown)


lance6716 (approved):

skoef:
Awesome, thanks for your swift action!

PR #590, opened 2021-06-17
return EOF response on COM_SET_OPTION commands

skoef:
I'm implementing a server that needs to respect the `CLIENT_MULTI_STATEMENTS` capability and the `MYSQL_OPTION_MULTI_STATEMENTS_ON` and `_OFF` options that can be set through `COM_SET_OPTION`. According to the [docs](https://dev.mysql.com/doc/internals/en/com-set-option.html) it is supposed to return EOF on success so I added a backwards compatible way of handling these commands properly.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -125,6 +125,12 @@ func (c *Conn) dispatch(data []byte) interface{} {
 		} else {
 			return r
 		}
+	case COM_SET_OPTION:
+		if err := c.h.HandleOtherCommand(cmd, data); err != nil {
+			return err
+		}
+
+		return eofResponse{}
 	default:
 		return c.h.HandleOtherCommand(cmd, data)
 	}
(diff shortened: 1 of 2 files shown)


server/command.go
```diff
@@ -125,6 +125,12 @@ func (c *Conn) dispatch(data []byte) interface{} {
 		} else {
 			return r
 		}
+	case COM_SET_OPTION:
+		if err := c.h.HandleOtherCommand(cmd, data); err != nil {
```
lance6716, line 129: Currently this is always `EmptyHandler`, so always returns an error 🤔 . Is this your expectation?

skoef: Well in my server implementation I implemented a Handler that does handle these commands. And `EmptyHandler` would've returned an error for this anyway.

Or perhaps I don't understand what you mean?

lance6716 (approved):

PR #591, opened 2021-06-22
chain results when server indicates more results are coming

skoef:
This is my implementation of what https://github.com/go-mysql-org/go-mysql/pull/524 initially tries to accomplish, but in this case it actually chains the results (not resultsets) recursively when it reads them in the client code.


skoef:
When subjecting this code to some production load, I see that my server implementation had a memory leak, I will investigate this further.

skoef:
closing this for now

PR #594, opened 2021-06-25
replaced all %lu, %ld and %u to %d for correct formatting of errors

skoef:
Many error messages were probably taken verbatim from MySQL's source without converting the format specifiers from C to Go. In Go, `fmt` treats all integers as int64 internally and makes no difference in signed, unsigned or long integers.

Currently:
`ERROR 1094 (HY000): Unknown thread id: %!l(uint32=234423432)u`

With this PR:
`ERROR 1094 (HY000): Unknown thread id: 234423432`


skoef:
@atercattus this might be a no-brainer

lance6716 (approved):

PR #597, opened 2021-06-29
Support for number of warnings in OK and EOF

skoef:
Support for this used to exist but was disabled/removed for some reason. This PR reads number of warnings in the client-side code and allows you to set the number of warnings on the server-side.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -22,6 +22,7 @@ type Conn struct {
 	authPluginName string
 	connectionID   uint32
 	status         uint16
+	warnings       uint16
 	salt           []byte // should be 8 + 12 for auth-plugin-data-part-1 and auth-plugin-data-part-2
 
 	credentialProvider  CredentialProvider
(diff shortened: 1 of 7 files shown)


server/conn.go
```diff
@@ -190,3 +191,7 @@ func (c *Conn) UnsetStatus(status uint16) {
 func (c *Conn) HasStatus(status uint16) bool {
 	return c.status&status > 0
 }
+
+func (c *Conn) SetWarnings(warnings uint16) {
```
atercattus, line 195: Do you really need this method?

skoef: Yes, that is the only way to set the unexported warnings parameter on the Conn struct. If I want to tell clients connecting to my server there were warnings executing their queries, I have to be able to manipulate that parameter. Alternatively, the warnings field could be exported, but that is little consistent with for instance `status`, which also has a somewhat similar setter function `SetStatus`.

skoef: snippet from my HandleQuery code:
```golang
func (q MyHandler) HandleQuery(query string) (*mysql.Result, error) {
    // ...

    result, err := backendConn.Execute(query)
    clientConn.SetWarnings(result.Warnings)

    // ...
    return result, err
}
```

PR #598, opened 2021-07-03
refine decoding decimal type

lance6716:
in https://github.com/pingcap/dm/issues/1826, I suspect `decodeDecimal` is too slow compared with other types. There's a result of benchmark comparing DECIMAL(40, 16) and INT MySQL type on master branch.

```bash
$ go test -check.b "testDecodeSuite"
PASS: row_event_test.go:1337: testDecodeSuite.BenchmarkInt	20000000	       114 ns/op
PASS: row_event_test.go:1285: testDecodeSuite.BenchmarkUseDecimal	  100000     26638 ns/op
```

and after the PR, the result will be 

```bash
PASS: row_event_test.go:1294: testDecodeSuite.BenchmarkNotUseDecimal	  500000      3509 ns/op
PASS: row_event_test.go:1284: testDecodeSuite.BenchmarkUseDecimal	  200000     13473 ns/op
```

There's a notable changes: the return type when `useDecimal == false` is string rather than float. I think it's inproper to use float to represent a precise decimal. Returning the string representation is more reasonable. Please see the unit test.


lance6716:
PTAL @atercattus

GMHDBJD (approved):
LGTM

atercattus (approved):
I don't like the use of magic "9" (and others around) in code a little.
But I hope we'll fix it in the near future :)

PR #601, opened 2021-07-12
fix mysql8.0 will return error;

jiuker:
mysql8.0+
sql [ show index from Schema.Table] result have 15 fields ；


jiuker:
sorry，i dont konw it fixed！

PR #612, opened 2021-08-16
applied small refactor writeResultSet to minimize code duplication

skoef:
and added the tests to back it up

### DIFF
--- a/server/resp.go
+++ b/server/resp.go
@@ -125,15 +125,7 @@ func (c *Conn) writeResultset(r *Resultset) error {
 		return err
 	}
 
-	for _, v := range r.Fields {
-		data = data[0:4]
-		data = append(data, v.Dump()...)
-		if err := c.WritePacket(data); err != nil {
-			return err
-		}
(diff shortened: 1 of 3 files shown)


server/resp.go
```diff
@@ -125,15 +125,7 @@ func (c *Conn) writeResultset(r *Resultset) error {
...
-	for _, v := range r.Fields {
-		data = data[0:4]
-		data = append(data, v.Dump()...)
-		if err := c.WritePacket(data); err != nil {
-			return err
-		}
-	}
-
-	if err := c.writeEOF(); err != nil {
+	if err := c.writeFieldList(r.Fields); err != nil {
```
atercattus, line 128: This variant adds a new memory allocation because do not reuses `data`. 

How about a buffer transfer to `writeFieldList`?

skoef: good point

atercattus (approved):

skoef:
Thanks @atercattus !

PR #613, opened 2021-08-18
allow controll --hex-blob param on Mysqldump

jintaoxu:
we can control this parameter


jintaoxu:
The default true may be good enough

PR #624, opened 2021-09-15
Adds parsing of IntVarEvents

IvoGoman:
Added support for parsing the [IntVarEvents](https://mariadb.com/kb/en/intvar_event/) as I have the need to parse these events.

- Removes the comment about IntVarEvents not being parsed
- Adds IntVarEvent and enables parsing of the event
- Adds rudimentary test


atercattus (approved):
lgtm with https://dev.mysql.com/doc/internals/en/intvar-event.html

lance6716 (approved):

PR #628, opened 2021-09-17
virtal column stored generated fix

crspecter:
fix the column define stored generated, go-mysql can identify

close https://github.com/go-mysql-org/go-mysql/issues/625


schema/schema.go
```diff
@@ -145,7 +145,7 @@ func (ta *Table) AddColumn(name string, columnType string, collation string, ext
 
 	if extra == "auto_increment" {
 		ta.Columns[index].IsAuto = true
-	} else if extra == "VIRTUAL GENERATED" {
+	} else if extra == "VIRTUAL GENERATED" || extra == "STORED GENERATED" {
 		ta.Columns[index].IsVirtual = true
```
lance6716, line 149: the name is `IsVirtual`, so don't save true for "STORED GENERATED"

crspecter: They all belong to automatically generated columns. Currently, go-mysql does not save the extra infomation. Need to define another variable in column to save it.

crspecter (commented):
solve issue #625

go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/crspecter/go-mysql v1.3.2 h1:/aOj3ptsZu6+gDpazMkR6jlSJGNa+LSs8Q1Y4IIvfgk=
```
lance6716, line 3: you can run `go mod tidy` to remove these unnecessary dependencies.

lance6716 (commented):
rest lgtm

crspecter (commented):
remove unnecessary dependencies

crspecter:
remove unnecessary dependencies

lance6716:
> remove unnecessary dependencies

please use `go mod tidy` to remove all unnecessary dependencies

schema/schema.go
skoef: Wouldn't `IsStored` for `STORED GENERATED` be more consistent with `IsVirtual` for `VIRTUAL GENERATED`?

PR #640, opened 2021-11-01
Read auth_data_len from InitialHandshake packet instead of hard code

guoyuanchao1202:
# Now
When Fake Slave receive initialHandshake packet from Master, it'll decodes this packet. The current implementation uses hard-coded methods to obtain auth_data_part_2. 
# Change
But actually we can get auth_data_len from initHandshake packet.  We can use the same method to get auth_plugin_name(ref to [MySQL Doc](https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::Handshake))
# Test
I've tested this change with my local MySQL, it's ok

### DIFF
--- a/client/auth.go
+++ b/client/auth.go
@@ -77,21 +77,24 @@ func (c *Conn) readInitialHandshake() error {
 		c.capability = uint32(binary.LittleEndian.Uint16(data[pos:pos+2]))<<16 | c.capability
 		pos += 2
 
-		// skip auth data len or [00]
+		// auth_data is end with 0x00, so default data length is 20 + 1
+		authDataLen := 21
+		if c.capability&CLIENT_PLUGIN_AUTH != 0 {
+			authDataLen = int(data[pos])
+		}
+
 		// skip reserved (all [00])
 		pos += 10 + 1
 
-		// The documentation is ambiguous about the length.
-		// The official Python library uses the fixed length 12
-		// mysql-proxy also use 12
-		// which is not documented but seems to work.
-		c.salt = append(c.salt, data[pos:pos+12]...)
-		pos += 13
-		// auth plugin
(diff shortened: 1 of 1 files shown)


guoyuanchao1202:
Btw, when syncer receive an EOF Packet, it's means that user set 'BINLOG_DUMP_NON_BLOCK' flag in COM_BINLOG_DUMP command and there is no more binlog event now. master will send this Packet instead of blocking

client/auth.go
```diff
@@ -77,21 +77,25 @@ func (c *Conn) readInitialHandshake() error {
 		c.capability = uint32(binary.LittleEndian.Uint16(data[pos:pos+2]))<<16 | c.capability
 		pos += 2
 
-		// skip auth data len or [00]
+		// auth_data is end with 0x00, max data length is 13 + 8 = 21
+		// ref to https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::Handshake
+		maxAuthDataLen := 21
+		if c.capability&CLIENT_PLUGIN_AUTH != 0 && int(data[pos]) < maxAuthDataLen {
```
lance6716, line 83: if `int(data[pos])` is `9`, `resetOfAuthDataEndPos` will equals `pos`. But as stated `$len=MAX(13, length of auth-plugin-data - 8)` in https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::Handshake, `resetOfAuthDataEndPos` should be `pos + 13 - 1`?

guoyuanchao1202: sorry, it's my fault. it's `min_auth_data_len` is 21, not `max_data_len`.

lance6716 (approved):

client/auth.go
guoyuanchao1202: and this pr is a little bit old, need to be updated :-)

PR #643, opened 2021-11-10
Add ConnectWithDialer

bakins:
We need more control over how dialing is done - in particular, integrating directly with https://pkg.go.dev/github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


client/conn.go
```diff
@@ -55,10 +56,23 @@ func getNetProto(addr string) string {
 func Connect(addr string, user string, password string, dbName string, options ...func(*Conn)) (*Conn, error) {
 	proto := getNetProto(addr)
 
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
atercattus, line 59: Use `10*time.Second`, please :)

But, LGTM.

lance6716 (approved):
also lgtm 

@atercattus we can change it to your style after this PR merged :P

PR #644, opened 2021-11-11
Reuse BytesBuffer in BinlogParser

D3Hunter:
to avoid reallocation of buffer every time

although, in our bench test, reusing buf don't contribute much, about 1% CPU(by gc)


D3Hunter:
@lance6716 PTAL

replication/parser.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/go-mysql-org/go-mysql/utils"
```
atercattus, line 7: Use goimports for better imports sorting.

replication/parser.go
```diff
@@ -126,14 +130,15 @@ func (p *BinlogParser) parseSingleEvent(r io.Reader, onEvent OnEventFunc) (bool,
...
+	if n, err = io.CopyN(buf, r, int64(h.EventSize-EventHeaderSize)); err != nil {
 		return false, errors.Errorf("get event err %v, need %d but got %d", err, h.EventSize, n)
 	}
 	if buf.Len() != int(h.EventSize) {
 		return false, errors.Errorf("invalid raw data size in event %s, need %d but got %d", h.EventType, h.EventSize, buf.Len())
 	}
 
-	rawData := buf.Bytes()
+	var rawData []byte
+	rawData = append(rawData, buf.Bytes()...)
```
atercattus, line 142: But we still have an allocation of almost the same size here. Is it enough for you?

D3Hunter: syncer of pingcap/dm uses this rawdata, so cannot be removed

atercattus (approved):

replication/parser.go
D3Hunter: in our bench test, reusing buf don't contribute much, about 1% CPU(by gc)

lance6716 (approved):

PR #649, opened 2021-12-01
Add test for Conn.ExecuteSelectStreaming

atercattus:


client/conn_test.go
```diff
@@ -0,0 +1,126 @@
...
+		s.c.Close()
+	}
+}
+
+var (
+	testExecuteSelectStreamingRows      = [...]string{"foo", "helloworld", "bar", "", "spam"}
+	testExecuteSelectStreamingTablename = "execute_plain_table"
+)
+
+func (s *connTestSuite) testExecute_CreateTable(c *C) {
```
lance6716, line 50: not sure if we should follow go's naming style like `testExecuteCreateTable`. This repo has some variables using underscore 🤔

lance6716 (approved):
lgtm

client/conn_test.go
atercattus: I copypasted this style from neighbour clientTestSuite :)

PR #651, opened 2021-12-10
packet: memory optimization for reading network packet

zr-hebo:
读取网络包遇到大的 event之后，会导致 buffer冲的很大，之后一直没法释放。从而导致整个程序的内存占用居高不下。

这个PR对这块逻辑做了优化，只缓存比较小的缓存块，避免了这种情况。

--
The buffer will be increased and can't be released when the packet is very big, thus the memory consumption keeps high. 
This PR tries to avoid it.

### DIFF
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -99,21 +99,26 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 	}
 
 	readBytes := buf.Bytes()
-	readSize := len(readBytes) + len(dst)
-	result := make([]byte, 0, readSize)
+	readSize := len(readBytes)
+	var result []byte
 	if len(dst) > 0 {
-		result = append(result, dst...)
-		result = append(result, readBytes...)
+		result = append(dst, readBytes...)
+		// if read block is big, do not cache buf any more
+		if readSize > utils.TooBigBlockSize {
+			buf = nil
+		}
 
 	} else {
-		if readSize <= utils.TooBigBlockSize {
-			result = append(result, readBytes...)
-		} else {
+		if readSize > utils.TooBigBlockSize {
 			// if read block is big, use read block as result and do not cache buf any more
 			result = readBytes
 			buf = nil
+
+		} else {
+			result = append(dst, readBytes...)
 		}
 	}
+
 	return result, nil
 }
 
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -3,15 +3,14 @@ package packet
 import (
 	"bufio"
 	"bytes"
-	"io"
-	"net"
-	"sync"
-
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


packet/conn.go
```diff
@@ -97,10 +96,25 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 
 	if err := c.ReadPacketTo(buf); err != nil {
 		return nil, errors.Trace(err)
+	}
+
+	readBytes := buf.Bytes()
+	readSize := len(readBytes) + len(dst)
+	result := make([]byte, 0, readSize)
```
atercattus, line 103: No. It doesn't look like reusing of memory :)

What about leaving the logic in BytesBufferPut, but removing it here?

zr-hebo: I make some change again.
always cache big buffer cause too much memory usage.

packet/conn.go
```diff
@@ -97,10 +96,30 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 
 	if err := c.ReadPacketTo(buf); err != nil {
 		return nil, errors.Trace(err)
+	}
+
+	readBytes := buf.Bytes()
+	readSize := len(readBytes)
+	var result []byte
+	if len(dst) > 0 {
```
atercattus, line 104: Maybe it will be enough for us leave the old version
```go
result = append(dst, buf.Bytes()...)
```
but change
```go
defer utils.BytesBufferPut(buf)
```
to
```go
defer func(){
	if readSize <= utils.TooBigBlockSize { // Or move this logic into utils.BytesBufferPut itself?
		utils.BytesBufferPut(buf)
	}
}()
```

packet/conn.go
atercattus: Maybe it will be enough for us leave the old version
```go
result = append(dst, buf.Bytes()...)
```
but change
```go
defer utils.BytesBufferPut(buf)
```
to
```go
defer func(){
	if readSize <= utils.TooBigBlockSize { // Or move this logic into utils.BytesBufferPut itself?
		utils.BytesBufferPut(buf)
	}
}()
```

zr-hebo: bytes in buffer would be copied once
```
result := append(dst, buf.Bytes()...)
```
but if not put it to buffer pool again, do not need copy it any more.
```
		if readSize > utils.TooBigBlockSize {
			// if read block is big, use read block as result and do not cache buf any more
			result = readBytes
			buf = nil

		} else {
			result = append(dst, readBytes...)
		}
```

atercattus: Ok, I understood. And agree.

Could you fix golangci warnings?

zr-hebo: OK, done.

atercattus (approved):

PR #652, opened 2021-12-13
fix that forget to readOK after writing

### LINKED ISSUE #535: Canal error: "invalid sequence 6 != 1" on first authentication
*System:*
- MySQL 8.0 (authentication plugin: caching_sha2_password)

*How to reproduce:*
- Stop any running Canal instance
- Restart MySQL instance, or execute `FLUSH PRIVILEGES;`
- Restart Canal
  * The error will appear
  * After seeing first error, if we start the Canal again, Canal would work normally

*Investigation:*
After checking I found the method  `packet.WritePublicKeyAuthPacket` leaves dirty buffer after writing, causing _sequence_ mismatch in the next read op
https://github.com/siddontang/go-mysql/blob/0c5789dd0bd378b4b84f99b320a2d35a80d8858f/packet/conn.go#L250
I added an blank read op to clear the buffer and it works properly:
```
	// return errors.Wrap(c.WritePacket(data), "WritePacket failed")
	err = c.WritePacket(data)
	if err != nil {
		return errors.Wrap(err, "WritePacket failed")
	}
	_, err = c.reader.Read(data) // clear dirty buffer
	return err
```
I think the same would happen to `packet.WriteClearAuthPacket` too, but I have not tested it yet

I'm not sure this is the correct way to fix this as I'm not very experienced with Go, so I just open an issue here

lance6716:
close #535

trigger condition: https://github.com/pingcap/ticdc/issues/3847#issuecomment-992225272

tested on my local MySQL 8.0.17

### DIFF
--- a/client/resp.go
+++ b/client/resp.go
@@ -143,9 +143,9 @@ func (c *Conn) handleAuthResult() error {
 				if err = c.WritePublicKeyAuthPacket(c.password, c.salt); err != nil {
 					return err
 				}
-				if _, err = c.readOK(); err != nil {
-					return err
-				}
+			}
+			if _, err = c.readOK(); err != nil {
+				return err
 			}
 		} else {
 			return errors.Errorf("invalid packet %x", data[0])
--- a/client/resp.go
+++ b/client/resp.go
@@ -7,10 +7,11 @@ import (
 	"encoding/binary"
 	"encoding/pem"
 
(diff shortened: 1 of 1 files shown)


atercattus (approved):
Looks like forgotten 
```go
_, err = c.readOK()
return err
```
from neighbour branch AUTH_SHA256_PASSWORD

So, lgtm.

PR #658, opened 2021-12-22
store json as string in rows events

D3Hunter:
store JSON value as `string` instead of `[]byte` to make it easier to use in query statement, as [go mysql driver](https://github.com/go-sql-driver/mysql) prepends `[]byte` values [with `_binary`](https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L267), and fails with `ERROR 3144 (22032): Cannot create a JSON value from a string with CHARACTER SET 'binary'.` if the datatype is `[]byte`


lance6716:
trigger CI

lance6716 (approved):

PR #663, opened 2022-01-13
enabled govet

skoef:
Enabled govet in golangci config and fixed some low-hanging fruit code issues.

The reason I ended up enabling govet is that I eventually want to enable govet's `fieldalignment` linter to lower the memory footprint of this library. To make the changes more readable, I will do that in a follow up PR later.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,9 +15,9 @@ linters:
     - nakedret
     - unconvert
     - whitespace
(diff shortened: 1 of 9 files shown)


atercattus (approved):

skoef:
Thanks @atercattus! Will you merge it as well?

PR #664, opened 2022-01-17
enabled fieldalign in govet settings

skoef:
As mentioned in #663 I enabled the `fieldalignment` linter in govet and aligned all mis-aligned structs. This was done mostly by using the `fieldalignment` tool and running it:

```
% go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest
% $GOPATH/bin/fieldalignment -fix ./canal
```

In some cases I had to make sure the comments (that fieldalignment doesn't take into account) were restored.


atercattus:
Thank you for the efforts, fieldalign gives some interesting insights. But there is a fact that we need to respect maintainability and code readability too.

When rearranging fields, their logical grouping should be preserved.

We should not separate Host and Port fields [here](https://github.com/go-mysql-org/go-mysql/pull/664/files#diff-c43542c83be2c477175d2840851f96840a634b2cc7b86723bc22fb949f99c31fR77)
```go
	Host string
	Port uint16
```
VS
```go
	Host string
	// a lot of strings ...
	Port uint16
```

The same goes for Tables+TablesDB and Tables+Databases [here](https://github.com/go-mysql-org/go-mysql/pull/664/files#diff-1fc4e550207116ffcbc26b6c7da760548e484220357a9b6fb7859ce5ca1eabb9R31).
They should be near each other.
```go
	// Will override Databases
	Tables  []string
	TableDB string

	Databases []string
```
VS
```go
	TableDB string
	// ...
	Databases    []string
	ExtraOptions []string

	// Will override Databases
	Tables []string
```

And another example is a mutex field that is bundled with fields it protects [here](https://github.com/go-mysql-org/go-mysql/pull/664/files#diff-c0f7004d3d1085dc71c57ad253aab8ab6b5e68531becb356633c7dbceb945181R36)
```go
	synchro struct {
		sync.Mutex
		// other fields
	}
```
VS
```go
	synchro struct {
		// other fields
		sync.Mutex
		// other fields
	}
```

There are more examples, but I think 3 is enough to demonstrate the principle.

So, let's use fieldalign only when it doesn't break the code readability?

skoef:
While converting certain structs I already imagined comments like this. I completely get the readability aspect, I will see if I can ignore the linter on specific structs that benefit from certain order in parameters.

server/conn.go
```diff
@@ -16,26 +16,22 @@ import (
...
 	credentialProvider  CredentialProvider
-	user                string
+	h                   Handler
+	stmts               map[uint32]*Stmt
+	serverConf          *Server
+	authPluginName      string
 	password            string
+	user                string
+	salt                []byte // should be 8 + 12 for auth-plugin-data-part-1 and auth-plugin-data-part-2
+	stmtID              uint32
```
atercattus, line 27: Could you keep `stmts` and `stmtID` near each other?

canal/master.go
```diff
@@ -8,11 +8,11 @@ import (
 )
 
 type masterInfo struct {
-	sync.RWMutex
+	gset mysql.GTIDSet
 
 	pos mysql.Position
 
-	gset mysql.GTIDSet
+	sync.RWMutex
```
atercattus, line 15: Could you keep this too?

replication/row_event.go
```diff
@@ -20,42 +20,46 @@ import (
 var errMissingTableMapEvent = errors.New("invalid table id, no corresponding table map event")
 
 type TableMapEvent struct {
-	flavor      string
-	tableIDSize int
+	/*
+	   The some fields are available only after MySQL-8.0.1 or MariaDB-10.5.0
```
atercattus, line 24: I don't like that before we had a clear separation between the "old" and "new" fields. And now they're all mixed up. :(

But I have no idea how to fix this w/o nolint. Do you have some ideas maybe?

server/conn.go
skoef: While the config of binlog parsing shouldn't really put a hit on the memory usage when it isn't optimally aligned, especially `server.Conn` and `client.Conn` can make a real change in memory usage in a connection hungry application. I'd say this is the one struct you'd want to align as best as possible. I can add comments on what each field represents, perhaps that is the right middle ground?

replication/row_event.go
skoef: Manually aligning the fields is a real puzzle, and probably you wouldn't end up with the perfect readable order after all. I'll just add nolint to it

lance6716 (commented):
I'm neutral to this PR :smile:

server/conn.go
skoef: @atercattus what is your opinion about this?

skoef:
let's just leave this as is for now

PR #666, opened 2022-01-17
enabled ineffassign in golangci

skoef:
In line with #663, #664 and #665, I enabled the ineffassign linter and fixed some small issues that came from it.

### DIFF
--- a/client/resp.go
+++ b/client/resp.go
@@ -53,11 +53,11 @@ func (c *Conn) handleOKPacket(data []byte) (*Result, error) {
 
 		//todo:strict_mode, check warnings as error
 		r.Warnings = binary.LittleEndian.Uint16(data[pos:])
-		pos += 2
+		// pos += 2
 	} else if c.capability&CLIENT_TRANSACTIONS > 0 {
 		r.Status = binary.LittleEndian.Uint16(data[pos:])
 		c.status = r.Status
-		pos += 2
+		// pos += 2
 	}
 
 	//new ok package will check CLIENT_SESSION_TRACK too, but I don't support it now.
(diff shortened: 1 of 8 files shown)


mysql/mysql_test.go
```diff
@@ -40,6 +40,7 @@ func (t *mysqlTestSuite) TestMysqlGTIDInterval(c *check.C) {
 
 	i, err = parseInterval("1-2")
```
atercattus, line 41: I see no difference between this test and the first. Maybe we should delete this test?

mysql/rowdata.go
```diff
@@ -27,7 +27,7 @@ func (p RowData) ParseText(f []*Field, dst []FieldValue) ([]FieldValue, error) {
 	var v []byte
 	var isNull bool
 	var pos int = 0
-	var n int = 0
+	var n int
```
atercattus, line 30: Maybe remove zero value for `pos` too?

client/resp.go
```diff
@@ -53,14 +53,13 @@ func (c *Conn) handleOKPacket(data []byte) (*Result, error) {
...
 		r.Warnings = binary.LittleEndian.Uint16(data[pos:])
-		pos += 2
 	} else if c.capability&CLIENT_TRANSACTIONS > 0 {
 		r.Status = binary.LittleEndian.Uint16(data[pos:])
 		c.status = r.Status
-		pos += 2
 	}
 
 	//new ok package will check CLIENT_SESSION_TRACK too, but I don't support it now.
+	// pos += 2
```
atercattus, line 62: It seems to me that it is still worth leaving the change of position next to the reading, and not taking it outside. Any future refactoring or code change can either make the code harder to understand or break the code.

mysql/mysql_test.go
skoef: there is none, I didn't notice the earlier test. I deleted it.

mysql/rowdata.go
skoef: sure

client/resp.go
skoef: I've placed them back and commented them, I think that is the most readable way

atercattus (approved):

lance6716 (approved):

skoef:
Thanks for the approval @lance6716 and @atercattus ! But I can't merge in this repo so if wou want to do this, thanks!

PR #673, opened 2022-02-01
Allow TLS connections in the driver

atomicules:
Opening this for discussion and feedback. I'm not expecting this to be merged in as-is.

- This mostly exposes what is already elsewhere in this codebase to the driver
- Allows for an optional `ssl=` URL parameter (query string) of values `true` and `custom`. Database must be specified for this to work
- `true` just does `c.UseSSL(true)` on the connection
- `custom` uses a package level variable of `customTLSConfig` to do `c.SetTLSConfig(customTLSConfig)`
- That variable is set via a new function, `SetCustomTLSConfig`
- To use a custom config you'd do a full import of the driver (not for side-effects only) and then do something like:

        driver.SetCustomTLSConfig(CaPem, make([]byte, 0), make([]byte, 0), false, "my.domain.name")

For what I need the basic TLS support is enough (`ssl=true` and `c.UseSSL(true)`) so if need be I could reduce this PR down to the first commit; However, since we are basically exposing what is already supported elsewhere in this codebase to the driver I thought I should get the custom TLS config support is a usable state (this PR).

Things that I could do with feedback/opinions on:

- Since the DSN style for this driver uses a `?` as a separator/indicator for the database the `ssl` parameter only works and makes sense if a database is specified. E.g:

        `?mydatabase&ssl=true`

    vs

        `?ssl=true`

    Which would try to connect to a database called `ssl=true`.

    I am ok with this limitation/requirement.

- I have also tried to keep the current style for parsing the URL parameters as opposed to re-working everything to use `url.Parse`. I assume there is a reason why that's not used and why `?` is used instead of `/` for the database separator/indicator; At the very least I understand that switching to `/` would be a breaking change for many folk
- I've allowed the Cert and Key to be optional in `NewClientTLSConfig`. I think that is a good idea, but I don't know if I've gone about it the best way
- I'm just allowing for a single `custom` TLSConfig. This means that if a program uses the driver then only one custom config can be set - which might not be desirable if that driver is connecting to multiple endpoints. For me, right now, that's fine. But I'm willing to look at this some more
- `SetCustomTLSConfig` is a very simple pass-through function. I welcome suggestions on how to improve. But it means I can just import the driver and still take advantage of `NewClientTLSConfig`


lance6716:
Do you have time to take a look @skoef ? seems you are an active user of it 😄

skoef:
I consider myself an active (or perhaps heavy?) user of go-mysql indeed, but haven't touched the `driver` package or TLS at all. So as far as TLS change, couldn't tell if that's a good idea or bad.

For the DSN part, I think @atomicules raises a fair question: why is the database formatted as a query parameter where this is "always" the path parameter in the DSN. Not only would that make switching between drivers a lot easier but also parsing the DSN could be handled pretty much if not completely by `url.Parse`. So IMHO this was a poor choice back then. Perhaps we could refactor parsing the DSN a bit that it throws an error when it matches `\?([^&]+)` or something?

lance6716:
> I'm just allowing for a single custom TLSConfig

I mostly argue on this decision, I think it's not much work to use a map to support many certificates. But I'm also ok if you want to change it in following pr.

My desire API for certificates is
- a certID for DSN
- the certID is returned by a registering function. User can use empty string to let the function generate a valid certID. User can provide the preferred certID as argument, and a bool flag as another argument to choose "error on certID already exists" or "adjust the certID if already exists"

driver/driver.go
```diff
@@ -23,31 +28,74 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
 	lastIndex := strings.LastIndex(dsn, "@")
 	seps := []string{dsn[:lastIndex], dsn[lastIndex+1:]}
 	if len(seps) != 2 {
-		return nil, errors.Errorf("invalid dsn, must user:password@addr[?db]")
+		return nil, errors.Errorf("invalid dsn, must user:password@addr[[?db[&param=X]]")
```
lance6716, line 31:
```suggestion
		return nil, errors.Errorf("invalid dsn, must user:password@addr[?db[&param=X]]")
```

And the comment above the function can be updated

driver/driver.go
```diff
@@ -23,31 +28,74 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
...
+	var err error
+	var c *client.Conn
 
 	if ss := strings.Split(seps[0], ":"); len(ss) == 2 {
 		user, password = ss[0], ss[1]
 	} else if len(ss) == 1 {
 		user = ss[0]
 	} else {
-		return nil, errors.Errorf("invalid dsn, must user:password@addr[?db]")
+		return nil, errors.Errorf("invalid dsn, must user:password@addr[[?db[&param=X]]")
```
lance6716, line 46:
```suggestion
		return nil, errors.Errorf("invalid dsn, must user:password@addr[?db[&param=X]]")
```

and rest place

driver/driver.go
```diff
@@ -23,31 +28,74 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
...
+			db = ss[0]
+		} else {
+			// We have to assume the first is the db
+			// Then need to handle possible multiple parameters / query strings
+			for i, p := range ss {
+				if i == 0 {
+					db = p
+				} else {
+					// Build key value pairs
+					kv := strings.Split(p, "=")
```
lance6716, line 67: seems here we don't escape `?`, `&`, `=` in parameters, maybe left a TODO to warn developers

driver/driver.go
```diff
@@ -23,31 +28,74 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
...
+					params[kv[0]] = kv[1]
```
lance6716, line 68: and before here we could check kv has at least length 2

driver/driver.go
```diff
@@ -23,31 +28,74 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
...
+			// This actually does insecureSkipVerify
+			// But not even sure if it makes sense to handle false? According to
+			// client_test.go it doesn't - it'd result in an error
+			c, err = client.Connect(addr, user, password, db, func(c *client.Conn) { c.UseSSL(true) })
+		case "custom":
+			// I was too concerned about mimicking what go-sql-driver/mysql does which will
+			// allow any name for a custom tls profile and maps the query parameter value to
+			// that TLSConfig variable... there is no need to be that clever.
+			// Instead of doing that, let's store required custom TLSConfigs in a map that
+			// uses the DSN address as the key
```
lance6716, line 91: sgtm, because other users and I in fact don't request TLS feature 😄 

I'm not sure about your use case, will the map be concurrently accessed? There's a harmless data race then.

driver/driver.go
```diff
@@ -229,3 +277,20 @@ func (r *rows) Next(dest []sqldriver.Value) error {
 func init() {
 	sql.Register("mysql", driver{})
 }
+
```
lance6716, line 280: please describe the usage as the comment. And maybe you want to return an error?

lance6716 (commented):
@skoef do you want to give an advice?

skoef:
If we introduce changes to the way the URL can be parsed, what I would do before I start: abstract the code currently parsing the URL to a separate function and create a bunch of unit tests for it. This way you can know for sure you didn't mess up the original way of parsing whilst introducing new features.

atomicules:
Hey, sorry for the long delay in getting back to this.

I've just pushed two commits to address previous feedback (and I see you've already started leaving feedback).

1. Allows multiple custom TLS configs. We still use `?ssl=custom` to indicate instead of a specific named value, but we look up the TLS config based on the address (host and port) of the DSN. We'd register/store it up front like so:

        driver.SetCustomTLSConfig(dsn, caPem, make([]byte, 0), make([]byte, 0), insecureSkipVerify, serverName)

2. Re-jigs the DSN to use a `/` as the path separator instead of `?` and uses `url.Parse` to split out the DSN. It still allows the old style DSN to be used so won't break things for existing users. If people want to use the url parameters (currently only `ssl=X`) then you need to use the new style DSN.

atomicules:
I'll try to address the latest feedback a bit more quickly.

driver/driver.go
lance6716: Hi @atomicules , don't forget this comment. It also causes the CI failure

driver/driver.go
```diff
@@ -3,51 +3,85 @@
...
+// Support both legacy style DSNs: user:password@addr[?db]
+// And more standard DSNs: user:password@addr/db?param=value
+// Optional parameters are supported in the standard DSN
 func (d driver) Open(dsn string) (sqldriver.Conn, error) {
-	lastIndex := strings.LastIndex(dsn, "@")
-	seps := []string{dsn[:lastIndex], dsn[lastIndex+1:]}
-	if len(seps) != 2 {
-		return nil, errors.Errorf("invalid dsn, must user:password@addr[?db]")
+	// If a "/" occurs after "@" and then no more "@" or "/" occur after that
+	standardDSN, matchErr := regexp.MatchString("@[^@]+/[^@/]+", dsn)
```
lance6716, line 32: Can you add an unit test to explain this behaviour? That's more easier to understand. It's OK that you don't want to add it.

atomicules: Added a [test](https://github.com/atomicules/go-mysql/blob/a5b62b18e3fd0f43060a0a6a3ba9cf96de6f142a/driver/driver_test.go#L84-L107). It required me to split out the [parsing into a separate function](https://github.com/atomicules/go-mysql/blob/a5b62b18e3fd0f43060a0a6a3ba9cf96de6f142a/driver/driver.go#L45-L79). See commit: https://github.com/go-mysql-org/go-mysql/pull/673/commits/a5b62b18e3fd0f43060a0a6a3ba9cf96de6f142a

driver/driver.go
atomicules: Addressed (hopefully) in commit https://github.com/go-mysql-org/go-mysql/pull/673/commits/09ab3510215f59477cc56624c8d9e32785b547db

atomicules:
> If we introduce changes to the way the URL can be parsed, what I would do before I start: abstract the code currently parsing the URL to a separate function and create a bunch of unit tests for it. This way you can know for sure you didn't mess up the original way of parsing whilst introducing new features.

@skoef Good point, but since I'd already started I've split out the new parsing code to a separate function and added a unit test _after the fact_. If deemed necessary I could rebase things and alter the git history, etc so that I split out the existing/original code and add a unit test for that.

Alternatively we could just add DSN examples into the tests for the new parsing code.

atomicules:
Added some more commits:

- (Finally) fixed linting issues
- Return an error from the `SetCustomTLSConfig` function
- Split out the DSN parsing into a separate function and added a unit test

From my point of view just two open issues:

1. Whether we want to rebase, etc so that I split out the parsing and add a unit test for the original parsing code, before going on to alter that parsing code?
2. I want to spend a little time thinking about the race condition on the `customTLSConfigMap`. I haven't hit an issue yet during my testing, but I feel like I should fix that.

driver/driver.go
atomicules: Decided to add simple/basic use of `sync.Mutex`.

driver/driver.go
```diff
@@ -3,51 +3,119 @@
...
+	"regexp"
+	"sync"
 
 	"github.com/go-mysql-org/go-mysql/client"
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
 	"github.com/siddontang/go/hack"
 )
 
+var customTLSMutex sync.Mutex
```
lance6716, line 21:
```suggestion
var customTLSMutex sync.Mutex

```

lance6716 (approved):
rest lgtm.

yes I hope we can add an unit test for original parsing code to remind us about the compatibility.

@skoef Do you have other comments?

driver/driver.go
atomicules: Fixed this.

skoef:
As long as we're backwards compatible ánd moving towards primarily using `url.Parse`, I couldn't be happier 👍

lance6716:
> If deemed necessary I could rebase things and alter the git history, etc so that I split out the existing/original code and add a unit test for that.

Sorry I missed this comment and the word "rebase" in later comment. I think we do not need it. Will merge this PR if you have nothing else to do. @atomicules

atomicules:
Thanks for the approval. I'll wait for comments from @skoef before merging

> yes I hope we can add an unit test for original parsing code to remind us about the compatibility.

I've not pushed up a commit, but what I could do is add code to `driver_test.go` like this:

```golang
func TestOriginalParseDSN(t *testing.T) {
	// List of DSNs to test and expected results
	// Use different numbered domains to more readily see what has failed - since we
	// test in a loop we get the same line number on error
	testDSNs := map[string]connInfo{
		"user:password@localhost?db":                 connInfo{standardDSN: false, addr: "localhost", user: "user", password: "password", db: "db", params: url.Values(nil)},
		"user@1.domain.com?db":                       connInfo{standardDSN: false, addr: "1.domain.com", user: "user", password: "", db: "db", params: url.Values(nil)},
	}

	for supplied, expected := range testDSNs {
		actual, err := originalParseDSN(supplied)
		if err != nil {
			t.Errorf("TestOriginalParseDSN failed. Got error: %s", err)
		}
		// Compare that with expected
		if !reflect.DeepEqual(actual, expected) {
			t.Errorf("TestOriginalParseDSN failed.\nExpected:\n%#v\nGot:\n%#v", expected, actual)
		}
	}
}

// originalParseDSN is from: https://github.com/go-mysql-org/go-mysql/blob/7153c75fcadc4083dcff2e772c4d4c906eccdeb3/driver/driver.go
func originalParseDSN(dsn string) (connInfo, error) {
	ci := connInfo{}

	lastIndex := strings.LastIndex(dsn, "@")
	seps := []string{dsn[:lastIndex], dsn[lastIndex+1:]}
	if len(seps) != 2 {
		return ci, errors.Errorf("invalid dsn, must user:password@addr[?db]")
	}

	if ss := strings.Split(seps[0], ":"); len(ss) == 2 {
		ci.user, ci.password = ss[0], ss[1]
	} else if len(ss) == 1 {
		ci.user = ss[0]
	} else {
		return ci, errors.Errorf("invalid dsn, must user:password@addr[?db]")
	}

	if ss := strings.Split(seps[1], "?"); len(ss) == 2 {
		ci.addr, ci.db = ss[0], ss[1]
	} else if len(ss) == 1 {
		ci.addr = ss[0]
	} else {
		return ci, errors.Errorf("invalid dsn, must user:password@addr[?db]")
	}

	return ci, nil
}
```

If we wanted something in there for reference? 

That would be instead of re-writing git history to add in before the changes.

atomicules:
> Sorry I missed this comment and the word "rebase" in later comment. I think we do not need it. Will merge this PR if you have nothing else to do. @atomicules

😄 I just wrote the comment above before seeing your comment.

I am happy for it to be merged as is.

atomicules:
Thank you!

PR #676, opened 2022-02-10
fixed reading connection attributes on server side

skoef:
Connection attributes were not correctly parse during handshake. This was due to a bug in readPluginName (position wasn't shifted far enough) and a wrong assumption made in parsing attributes. This PR solves it.

I added some additional tests, so prevent regressions in the future. This involved adding a mock of Handler, which I think can be useful to replace `testHandler` as well.


lance6716 (approved):
Thanks!

Issue #677, opened 2022-02-11
time "-00:00:00.01" parse value is not right


rudy-gao:
Hello,

When I use go-mysql version 1.4.0, then I insert "-00:00:00.01" to time(2) type ，but parse binlog value is "00:00:00"

```
mysql>  select version();
+---------------+
| version()     |
+---------------+
| 5.7.25-28-log |
+---------------+
1 row in set (0.00 sec)

CREATE TABLE `t1` (
  `id` int(11) NOT NULL,
  `a_time` time(2) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;

mysql> insert into t1(id,a_time) values(1,'-00:00:00.01');
Query OK, 1 row affected (0.20 sec)

mysql> select * from t1;
+----+--------------+
| id | a_time       |
+----+--------------+
|  1 | -00:00:00.01 |
+----+--------------+

=== WriteRowsEventV2 ===
Date: 2022-02-11 19:29:08
Log position: 1533
Event size: 44
TableID: 765
Flags: 1
Column count: 2
Values:
0:1
1:"00:00:00"

```

lance6716:
thanks for reporting, I can reproduce it.

I'll take a look at my spare time, can you help to check what's wrong with the source code? maybe https://github.com/go-mysql-org/go-mysql/blob/d1666538b005e996414063695ca223994e9dc19d/replication/row_event.go#L1098-L1107 is a good entry

rudy-gao:
I test it will go to code

```
case MYSQL_TYPE_TIME2:
	v, n, err = decodeTime2(data, meta)
```
when intPart=-1, then it will intPart++  to intPart=0
```
func decodeTime2(data []byte, dec uint16) (string, int, error) {
		if intPart < 0 && frac != 0 {
			intPart++     /* Shift to the next integer value */
			frac -= 0x100 /* -(0x100 - frac) */
		}
```

lance6716:
@rudy-gao Thanks. I now know we forget to check the `frac` part in https://github.com/go-mysql-org/go-mysql/blob/d1666538b005e996414063695ca223994e9dc19d/replication/row_event.go#L1509-L1511 and https://github.com/go-mysql-org/go-mysql/blob/d1666538b005e996414063695ca223994e9dc19d/replication/row_event.go#L1408-L1410

[...]

PR #678, opened 2022-02-18
fix TIME value near zero will become zero

lance6716:
close https://github.com/go-mysql-org/go-mysql/issues/677

Thanks @rudy-gao for finishing all the code work, I just help him file a PR


lance6716:
@atercattus PTAL

lance6716:
In fact rudy-gao is the author and I'm the reviewer :P just merge it.

PR #680, opened 2022-03-03
添加rowEvent当前的position字段和fakeRotate也会调用onRotate

qq906907952:
主要是两个修改
rowEvent添加CurPos 表示当前event的position 
fakeRotate也会调用onRotate 而不是直接continue 目的是同步一开始是有时候需要记录当前的log文件名

### DIFF
--- a/canal/rows.go
+++ b/canal/rows.go
@@ -3,6 +3,8 @@ package canal
 import (
 	"fmt"
 
+	"github.com/go-mysql-org/go-mysql/mysql"
+
 	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/go-mysql-org/go-mysql/schema"
 )
@@ -16,6 +18,8 @@ const (
 
 // RowsEvent is the event for row replication.
 type RowsEvent struct {
+	// the position in this event
+	CurPos mysql.Position
 	Table  *schema.Table
 	Action string
 	// changed row list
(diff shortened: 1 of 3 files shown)


canal/rows.go
```diff
@@ -3,6 +3,8 @@ package canal
 import (
 	"fmt"
 
+	"github.com/go-mysql-org/go-mysql/mysql"
+
```
lance6716, line 7: this empty line is not needed

canal/rows.go
```diff
@@ -16,6 +18,8 @@ const (
 
 // RowsEvent is the event for row replication.
 type RowsEvent struct {
+	// the position in this event
+	CurPos mysql.Position
```
lance6716, line 22: Can your application read the last position from `replication.EventHeader` or use `LogPos-EventSize`? So we don't need change this library.

qq906907952: 就我的理解 header的LogPos 是next position. 由于log rotate , 那么要获取当前event对应的position和logName 需要记录rotate前的logName 我觉得这个既然在 runSyncBinlog的主循环能拿到 为什么不直接记录到event呢?

lance6716: There're many fake event so that last event's end position is not next event's begin position. I'm not sure if they are totally checked. For example, heartbeat event in MySQL and MariaDB GTID list event

qq906907952: 所以意思是我并不能准确拿到当前event对应的position是吗?

lance6716: I think you can use `LogPos-EventSize`

considering

RowEvent1 EventSize=10, LogPos=110,
SomeFakeEvent EventSize=undetermined, LogPos=undetermined,
RowEvent2 EventSize=10, LogPos=120,

Now RowEvent2 has a start position of 110. But its previous event is SomeFakeEvent, the LogPos may not be 110 or 0.

To detect SomeFakeEvent we have found some rules, like LogPos=0, LOG_EVENT_ARTIFICIAL_F, but I don't have confidence that we list them totally.

qq906907952: 好的 我大概明白了 LogPos-EventSize 应该也能满足需求 pr我close掉了

PR #686, opened 2022-03-17
Utilise byteSlice as a part of struct to avoid allocations

moredure:


lance6716:
Hi, can you explain what's the benefit of this change? I'm not sure what you want to improve, maybe `*[]byte` is better than `*struct{ []byte }`

moredure:
from code quality perspective struct is more simple, *[]byte needed to be dereferenced first, but original link to the slice pointer should be returned back to the pool.

lance6716:
> from code quality perspective struct is more simple, *[]byte needed to be dereferenced first, but original link to the slice pointer should be returned back to the pool

Yes, I just can't understand which part contributes to the title "Utilise byteSlice as a part of struct to avoid allocations". In this PR I guess you want to avoid allocating the slice header for `[]byte`? And remove the 10-length channel just keep same style as https://github.com/go-mysql-org/go-mysql/pull/682 ?

moredure:
Exactly, slice header + channel of size 10.

atercattus:
lgtm too. Thank you for waiting for me :)

PR #689, opened 2022-04-02
handle columnIdx out range data size

bingoohuang:
1. handle columnIdx out range data size 
2. refactor code to avoid type cast

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -100,7 +100,7 @@ func (c *Canal) runSyncBinlog() error {
 			}
 		case *replication.RowsEvent:
 			// we only focus row based event
-			err = c.handleRowsEvent(ev)
+			err = c.handleRowsEvent(ev, e)
 			if err != nil {
 				e := errors.Cause(err)
 				// if error is not ErrExcludedTable or ErrTableNotExist or ErrMissingTableMeta, stop canal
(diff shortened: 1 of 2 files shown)


canal/sync.go
```diff
@@ -245,9 +245,7 @@ func (c *Canal) updateReplicationDelay(ev *replication.BinlogEvent) {
 	atomic.StoreUint32(c.delay, newDelay)
 }
 
-func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
-	ev := e.Event.(*replication.RowsEvent)
-
+func (c *Canal) handleRowsEvent(e *replication.BinlogEvent, ev *replication.RowsEvent) error {
```
lance6716, line 248: we can change the parameter to (*EventHeader, *RowsEvent), that's not duplicated

bingoohuang: Yeah, you are right!

bingoohuang: As your suggestion, the parameters have been updated.

skoef:
Is there perhaps a unit test for this piece of code that can be adjusted to test the specific issue @bingoohuang ran into? To prevent future regression?

dveeden:
@bingoohuang @lance6716 what's the next step for this PR?

lance6716:
replaced by https://github.com/go-mysql-org/go-mysql/pull/856 , because that PR has unit tests 😄

Issue #695, opened 2022-05-30
Get the user's command. After I use PHP to link the prox request, the loop does not exit. Do you have any good ideas, boss?


Skobo:
Get the user's command. After I use PHP to link the prox request, the loop does not exit. Do you have any good ideas, boss?

for {
	conn.HandleCommand()
}

skoef:
If you're referring to the example from the README and in fact are using `EmptyHandler{}`, you should change your code to something like this:

```golang
for {
  if err := conn.HandleCommand(); err != nil {
    fmt.Printf("error: %s\n", err)
    break
  }
}
```

This would give away that any command handled by the `EmptyHandler` will fail and you should implement your own handler (implementing the `Handler` interface) to do something with the commands your PHP script sends you. If you already implement your own handler, you should still handle the errors like in my example above to see what is going on.

Skobo:
boss,thank you very much

PR #696, opened 2022-05-31
improved the server example in README

skoef:
Following #695 I improved the server example a bit in the documentation


lance6716 (approved):
thanks!

Issue #698, opened 2022-06-09
Add options  or expose method to set logger output media


naughtyGitCat:
now default output to console, in some circumstance

```bash
nohup main &> /tmp/canal.log & 
```
may cause the partition no free space, and cause the whole system to hang

So, It will be better to take these solutions into think:
1. Add options  when new canal struct, set logger, or pass a user-defined logger
2. Add options  when new canal struct,  set mute or to some file, better rotated

Issue #632, opened 2021-10-01
Disable the logger


uudashr:
Is there a way to disable the logging. It logs thing that I didn't expect to log.

uudashr:
In my opinion, logging should not reside inside library. If the user expect to be notified, then we can use callback.

dveeden:
I think with `log/slog` this might be fixed?

cc @lance6716

PR #699, opened 2022-06-09
Allow logger override

cameron-p-m:
## Problem

Ideally libraries don't have a logger at all. But this is a smaller change by allowing a user to set the Logger. This will allow the previous behaviour if needed.

Closes: https://github.com/go-mysql-org/go-mysql/issues/698 and maybe others

## Feedback?

Should errors ever log or just return?

### DIFF
--- a/canal/config.go
+++ b/canal/config.go
@@ -4,11 +4,13 @@ import (
 	"crypto/tls"
 	"io/ioutil"
 	"math/rand"
+	"os"
 	"time"
 
 	"github.com/BurntSushi/toml"
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
+	"github.com/siddontang/go-log/log"
 )
 
 type DumpConfig struct {
@@ -86,6 +88,9 @@ type Config struct {
 
 	// Set TLS config
 	TLSConfig *tls.Config
+
+	//Set Logger
+	Logger *log.Logger
 }
 
 func NewConfigWithFile(name string) (*Config, error) {
@@ -123,6 +128,9 @@ func NewDefaultConfig() *Config {
 	c.Dump.ExecutionPath = "mysqldump"
 	c.Dump.DiscardErr = true
 	c.Dump.SkipMasterData = false
+
+	streamHandler, _ := log.NewStreamHandler(os.Stdout)
+	c.Logger = log.NewDefault(streamHandler)
 
 	return c
 }
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -11,7 +11,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/ast"
-	"github.com/siddontang/go-log/log"
 )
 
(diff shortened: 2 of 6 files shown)


canal/sync.go
```diff
@@ -107,7 +105,6 @@ func (c *Canal) runSyncBinlog() error {
 				if e != ErrExcludedTable &&
 					e != schema.ErrTableNotExist &&
 					e != schema.ErrMissingTableMeta {
-					log.Errorf("handle rows event at (%s, %d) error %v", pos.Name, curPos, err)
```
skoef, line 110: why not replace this with a call to `c.cfg.Logger.Errorf`?

cameron-p-m: I can if you want!

cameron-p-m: I don't think we should log errors in the library, just return them

skoef: that seems like the consistent thing to do

cameron-p-m: sure I will change

lance6716:
thanks for your pr! I'm quite busy today, I will take a look tomorrow.

canal/config.go
```diff
@@ -124,5 +128,7 @@ func NewDefaultConfig() *Config {
 	c.Dump.DiscardErr = true
 	c.Dump.SkipMasterData = false
 
+	c.Logger = log.NewDefault(&log.NullHandler{})
```
lance6716, line 131: Can you use the handler from `NewStreamHandler(os.Stdout)` (or also file a PR to siddontang/golog to make newStdHandler public)? so we won't break compatibility and users like you can still change the logger.

lance6716 (commented):
I'll remove "close #632" in PR description, because other place such as `binlogsyncer.go` still access the default logger.

cameron-p-m:
@lance6716 I made the same change in binlog syncer too now

lance6716 (commented):
found that there's still some reference to global logger, like go-mysqldump, https://github.com/cameron-p-m/go-mysql/blob/fix/allow-setting-logger/mysql/mariadb_gtid.go#L96 . But it's OK to left them for a future PR.

unit tests failed, rest lgtm

@skoef do you have any comment?

skoef:
I'm fine with this!

Perhaps sometimes we can refactor using `siddontang/go-log` and have it replaced with something more generic/mainstream as it seems that `siddontang/go-log` is no longer maintained. But that's beyond the scope of this PR!

cameron-p-m:
@lance6716 updated and fixed tests

lance6716 (approved):

naughtyGitCat:
@lance6716 when we can use this feature, v1.6? is there any way to use latest master code

lance6716:
@atercattus Do you have a release plan?

@naughtyGitCat Currently you can specify the master branch or commit hash in go.mod https://go.dev/ref/mod#vcs-branch

atercattus:
@lance6716, I'll do it soon.

atercattus:
Done https://github.com/go-mysql-org/go-mysql/releases/tag/v1.6.0

PR #706, opened 2022-06-27
Fix initialization of TLS connections

### LINKED ISSUE #705: Is this code supposed to work?
https://github.com/go-mysql-org/go-mysql/blob/master/client/conn.go#L87

`c.tlsConfig` can't be anything but nil at that point since the options array hasn't been applied to the connection yet

viciious:
The check for tlsConfig being non-nil was in the wrong place. Fixes #705

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -84,16 +84,11 @@ func ConnectWithDialer(ctx context.Context, network string, addr string, user st
 		return nil, errors.Trace(err)
 	}
 
-	if c.tlsConfig != nil {
-		c.Conn = packet.NewTLSConn(conn)
-	} else {
-		c.Conn = packet.NewConn(conn)
-	}
-
 	c.user = user
 	c.password = password
 	c.db = dbName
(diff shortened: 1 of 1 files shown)


client/conn.go
```diff
@@ -103,6 +98,12 @@ func ConnectWithDialer(ctx context.Context, network string, addr string, user st
 		options[i](c)
 	}
 
+	if c.tlsConfig != nil {
+		seq := c.Conn.Sequence
+		c.Conn = packet.NewTLSConn(conn)
+		c.Conn.Sequence = seq
```
lance6716, line 104: `packet.NewConn(conn)` just created some buffer, it didn't write MySQL data so I think Sequence is untouched.

lance6716 (commented):
rest lgtm

client/conn.go
viciious: Yet it could be modified via an option call since it's a public field

PR #714, opened 2022-07-13
use net.JoinHostPort to join host and port

D3Hunter:
as title, so we can use ipv6 address more naturally(not workaround with explicit square brackets)


lance6716 (approved):

D3Hunter:
close, will create another one later

PR #719, opened 2022-07-18
Added SetTableCache method to set new schema for a table without load…

sergan99:
## Problem
Canal lacks ability to set table schema cache bypassing the schema reload from the replicated source database. This makes it impossible to replay the binlog with several table schema changes in the sequence even when you have history of table schemas for every change at your hand.
## Solution
This PR adds SetTableCache method for canal users to be able to directly set table schema whenever a user sees fit, for example, when detecting schema change (through a table rename statement e.x. RENAME _temp_migration_123 to my_table) while replaying binlog history.

### DIFF
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -380,6 +380,18 @@ func (c *Canal) ClearTableCache(db []byte, table []byte) {
 	c.tableLock.Unlock()
 }
 
+// SetTableCache sets table cache value for the given key
+func (c *Canal) SetTableCache(db []byte, table []byte, schema *schema.Table) {
+	key := fmt.Sprintf("%s.%s", db, table)
+	c.tableLock.Lock()
+	c.tables[key] = schema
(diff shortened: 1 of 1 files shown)


sergan99:
@lance6716 this PR is minimal and straight forward I think. Also it is a bit of a blocker for us, of course unless we use a fork. Thanks for looking at it!

canal/canal.go
```diff
@@ -380,6 +380,18 @@ func (c *Canal) ClearTableCache(db []byte, table []byte) {
 	c.tableLock.Unlock()
 }
 
+// SetTableCache sets table cache value for the given key
```
lance6716, line 383: "for the given key" maybe not as straightforward as "for the given table"

lance6716 (approved):

canal/canal.go
sergan99: Changed it to be "for the given table". Thanks!

PR #721, opened 2022-07-22
fix: return null values properly in proxy mode

btoonk:
When using go-mysql in proxy mode, the null values are not handled correctly.
This PR is an attempt to solve this. 

The same logic is already in use in the BuildSimpleTextResultset function.


skoef:
Thanks for your contribution! Could you extend any existing test case to prove your fix solves the issue?

lance6716:
Hi @btoonk , I found a bit different place to fix this issue.

`LengthEncodedInt` and `PutLengthEncodedInt` in util.go seems are inverse functions of each other, can you check if we add the `n=0` case in `LengthEncodedInt` the issue will also be fixed?

If it's OK, can you add an unit test to verify that the two functions can handle the output of each other and return the original input?

btoonk:
@skoef I added a test that proves this fix solved the issue.

btoonk:
@lance6716 I don't think we can do it in PutLengthEncodedInt (or in PutLengthEncodedString), I think not all data that is nil (or n=0) is equal to NULL (0xfb), I think that is only the case for row values. That is why we can do it in the writeFieldValues function.

lance6716:
Hi, I'll try to fix the CI problem and then come to your PR 😄

lance6716 (approved):

PR #724, opened 2022-08-17
removed use of deprecated ioutil

skoef:
The package `io/ioutil` has been deprecated since Go 1.16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ 1.18, 1.17, 1.16, 1.15 ]
+        go: [ 1.18, 1.17, 1.16 ]
     name: Tests Go ${{ matrix.go }}
     runs-on: ubuntu-18.04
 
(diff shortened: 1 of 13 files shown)


skoef:
It seems CI fails because we're still supporting Go 1.15, is that really still necessary?

lance6716:
GitHub ["Used by"](https://github.com/go-mysql-org/go-mysql/network/dependents?package_id=UGFja2FnZS0yMjg4MzUzNTc0) lists other projects that using this library. I checked about 20 active projects, only https://github.com/alibaba/ilogtail uses golang version < 1.16. No big deal. I agree on this PR.

What do you think? @atercattus 

and seems golangci is still complaining about some problems, can you also fix them? @skoef

skoef:
While not being super-descriptive, the `File is not goimport-ed` actually meant "I don't like the whitespace in your comments!". It is fixed now.

lance6716:
I have changed the "required CI" for master branch, you can push again to trigger the new rules. Removing v1.15 from the test matrix is a good commit 😄 @skoef 

https://github.com/go-mysql-org/go-mysql/blob/4c42f69ded2468f35b1c24fa7b25fbb3c8f6a355/.github/workflows/ci.yml#L8

PR #726, opened 2022-08-26
fix: fix client field list

black-06:
### What bug does this MR fix?
Client FieldList is missing the first field because check ERR Packet.
### How is the bug fixed?
Check ERR Packet in loop, in this way we can parse the first packet into FieldData

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -328,20 +328,17 @@ func (c *Conn) FieldList(table string, wildcard string) ([]*Field, error) {
 		return nil, errors.Trace(err)
 	}
 
-	data, err := c.ReadPacket()
-	if err != nil {
-		return nil, errors.Trace(err)
-	}
-
 	fs := make([]*Field, 0, 4)
 	var f *Field
-	if data[0] == ERR_HEADER {
-		return nil, c.handleErrorPacket(data)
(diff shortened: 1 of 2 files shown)


lance6716:
Hi, can you add an unit test to explain what's FieldList? Do you mean https://dev.mysql.com/doc/internals/en/com-field-list.html ?

black-06:
Reference example: 
`show columns from test;`
| Field      | Type         | Null | Key | Default | Extra |
|------------|--------------|------|-----|---------|-------|
| first_name | varchar(127) | YES  |     | NULL    |       |
| last_name  | varchar(127) | YES  |     | NULL    |       |

However, the result in the client FieldList is only `last_name`:
| Field      | Type         | Null | Key | Default | Extra |
|------------|--------------|------|-----|---------|-------|
| last_name  | varchar(127) | YES  |     | NULL    |       |

it is missing the first field

black-06:
Yes, it is ComFieldList. 
I will add a test

black-06:
I added a simple test that asserts fileds length.

black-06:
I should use connTestSuite :(

lance6716 (approved):

PR #729, opened 2022-09-07
packet: fix misusing of defer

### LINKED ISSUE #728: Misusing of defer
https://github.com/go-mysql-org/go-mysql/blob/1c55c36dc919cb9cf7971359e2e185940dd17d2c/packet/conn.go#L90-L112

`defer utils.BytesBufferPut(buf)` and
```go
 		// if read block is big, do not cache buf any more 
 		if readSize > utils.TooBigBlockSize { 
 			buf = nil 
 		} 
```

`defer` capture `buf` early, so modify `buf` after `defer` not working.

shizhx:
defer capture arguments at once, so modify arguments after defer line not works
Closes #728

### DIFF
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -90,7 +90,9 @@ func (c *Conn) ReadPacket() ([]byte, error) {
 func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 	// Here we use `sync.Pool` to avoid allocate/destroy buffers frequently.
 	buf := utils.BytesBufferGet()
-	defer utils.BytesBufferPut(buf)
+	defer func() {
+		utils.BytesBufferPut(buf)
+	}()
 
 	if err := c.ReadPacketTo(buf); err != nil {
 		return nil, errors.Trace(err)


lance6716 (approved):
Thanks for fixing it!

Just for discuss, do you think we should also remove two `buf = nil` below? Because we know `BytesBufferPut` can skip putting it back to cache pool when it's too big. 
But line 113 "exposing the `buf.Bytes()` to caller" must be used along with "not putting `buf` back to cache pool". Explicitly assigning `buf` to `nil` is better than implicitly doing it in `BytesBufferPut`.

For now I slightly prefer we remove the `buf = nil` in line 108 and its if-block. What do you think?

shizhx:
If we remove the `buf = nil` line, we need to add a new API to achieve "copy buffer" feature, maybe `utils.BytesBufferCopyAndPut` or better name, the new API implementation can decide between "actually copy buffer" and "return buffer".

But the new API only suitable for successful `return`, failed `return` we also has to call "BytesBufferPut".

So, I think removing the `buf = nil` line is difficult.

lance6716:
I can't understand why we need to add a new API to remove `buf = nil`. Can you explain why current code plus deleting `buf = nil` is not proper?

shizhx:
> I can't understand why we need to add a new API to remove `buf = nil`. Can you explain why current code plus deleting `buf = nil` is not proper?

as your post above:
> But line 113 "exposing the buf.Bytes() to caller" must be used along with "not putting buf back to cache pool". Explicitly assigning buf to nil is better than implicitly doing it in BytesBufferPut.

If we simply remove the `buf = nil` line, when the `put or not` condition changed in `BytesBufferPut`, current code maybe buggy.

So we need an explicit API to do it, maybe a `BytesBufferCopyAndPut` or `IsBytesBufferActuallyPut`.

atercattus:
Hello.
What is wrong with defer here? :) `buf` is a `*bytes.Buffer`, and we do not modify it (expect buf = nil cases).

shizhx:
@atercattus The following code `buf = nil` want `utils.BytesBufferPut(nil)` to prevent actually put `buf`. but `defer utils.BytesBufferPut(buf)` cannot achieve that, always non-nil argument passed to `utils.BytesBufferPut`.

atercattus:
Oh, sorry, I'm blind :)
It's my fault, I missed it in #651.

And I agree what a better way is wrap defer to func here.

atercattus (approved):

PR #732, opened 2022-09-23
Make logger require an interface instead of a concrete implementation

mcuelenaere:
It is much easier to integrate 3rd-party log implementations by requiring an interface instead of requiring a Logger struct (which implements this interface).

### DIFF
--- a/canal/config.go
+++ b/canal/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
 	"github.com/siddontang/go-log/log"
+	"github.com/siddontang/go-log/loggers"
 )
 
 type DumpConfig struct {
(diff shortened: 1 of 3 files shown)


lance6716 (commented):
please fix lint errors, rest lgtm

And is the `Contextual` from https://github.com/siddontang/go-log/blob/master/loggers/loggers.go better suit your use case? Seems it's more flexible.

mcuelenaere:
> And is the `Contextual` from https://github.com/siddontang/go-log/blob/master/loggers/loggers.go better suit your use case? Seems it's more flexible.

It is more flexible, but AFAIK it is not implemented by the logger, so it would break things: https://github.com/siddontang/go-log/blob/master/log/logger.go#L63

mcuelenaere:
> please fix lint errors, rest lgtm

Should be fixed now.

lance6716 (approved):

PR #736, opened 2022-11-15
Code refractoring

tech-sumit:
Hi there, 
I added some tweaks before using the library in the code. Please Addin changes.
These are primarily refractory changes.


lance6716 (approved):
Thanks!

PR #737, opened 2022-11-16
parser: allow user-defined rows_event decode func

GMHDBJD:
add RowsEventDecodeFunc in BinlogSyncerConfig that allow user-defined rows_event decode func


replication/parser.go
```diff
@@ -40,6 +40,8 @@ type BinlogParser struct {
 	useDecimal          bool
 	ignoreJSONDecodeErr bool
 	verifyChecksum      bool
+
+	rowsEventDecodeFunc func(*RowsEvent, []byte) error
```
lance6716, line 44: maybe we can set the default value to `(*RowsEvent).Decode`, so in `parseEvent` we don't need to check it many times.

lance6716 (commented):
Also we have planed some use cases of this feature, such as skipping decoding rows events of uninterested table. But in order to do that, we need to know the table name of this RowsEvent. In the original Decode function, the table name is written to RowsEvent.Table by accessing private members

```
	e.TableID = FixedLengthInt(data[0:e.tableIDSize])
...
	e.Table, ok = e.tables[e.TableID]
```

and our customized RowsEventDecodeFunc can not do that.

Maybe we should adjust more code? Or we inject the decoding function at `e.decodeRows`?

GMHDBJD:
> and our customized RowsEventDecodeFunc can not do that.

```
rowsEventDecodeFunc = func(re *replication.RowsEvent, data []byte) error {
			_, err := re.DecodeHeader(data)
			tb := &filter.Table{
				Schema: string(re.Table.Schema),
				Name:   string(re.Table.Table),
			}
}
```

lance6716 (approved):

PR #738, opened 2022-11-17
speed up gtid process for mysql

GMHDBJD:
- add InsertInterval for mysql gtidset

### DIFF
--- a/mysql/gtid.go
+++ b/mysql/gtid.go
@@ -1,6 +1,9 @@
 package mysql
 
-import "github.com/pingcap/errors"
+import (
+	"github.com/google/uuid"
+	"github.com/pingcap/errors"
+)
 
 type GTIDSet interface {
 	String() string
@@ -15,6 +18,8 @@ type GTIDSet interface {
 	Update(GTIDStr string) error
 
 	Clone() GTIDSet
+
+	AddGTID(uuid.UUID, int64, uint32, uint32, uint64) error
 }
 
 func ParseGTIDSet(flavor string, s string) (GTIDSet, error) {
--- a/mysql/gtid.go
+++ b/mysql/gtid.go
@@ -1,7 +1,6 @@
 package mysql
 
 import (
-	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


mysql/gtid.go
```diff
@@ -15,6 +18,8 @@ type GTIDSet interface {
 	Update(GTIDStr string) error
 
 	Clone() GTIDSet
+
+	AddGTID(uuid.UUID, int64, uint32, uint32, uint64) error
```
lance6716, line 22: this function signature is very unfriendly. I prefer we don't add it.

We can use type assert to get MariadbGTIDSet and MysqlGTIDSet and call its AddGTID

mysql/mysql_gtid.go
```diff
@@ -111,6 +111,52 @@ func (s IntervalSlice) Normalize() IntervalSlice {
...
+			break
+		} else {
+			(*s)[i-1].Start = min((*s)[i-1].Start, (*s)[i].Start)
+			(*s)[i-1].Stop = max((*s)[i-1].Stop, (*s)[i].Stop)
+			count++
+		}
+	}
+	i++
+	if count > 0 {
+		for i != total {
```
lance6716, line 149: seems we can use a `copy` to move multiple elements with `count` offset, and truncate the last `count` elements

```
	if count > 0 {
		i++
		copy((*s)[i:], (*s)[i+count:])
		*s = (*s)[:total-count]
	}
```

mysql/mysql_gtid.go
```diff
@@ -439,6 +484,17 @@ func (s *MysqlGTIDSet) Update(GTIDStr string) error {
 	return nil
 }
 
+func (s *MysqlGTIDSet) AddGTID(uuid uuid.UUID, gno int64) error {
```
lance6716, line 487: remove returned error?

lance6716 (commented):
rest lgtm

lance6716 (approved):

PR #739, opened 2022-11-25
位点信息增加canal时间戳支持，可向用户端展示消费延迟时间

BLAZZ:
handler可以获取位点时间戳，用于展示延迟信息：
```go
func (h *handler) OnXID(nextPos mysql.Position) error {
	h.queue <- model.Position{
		Name:      nextPos.Name,
		Pos:       nextPos.Pos,
		Timestamp: nextPos.Timestamp,
		Force:     false,
	}
	return nil
}
```
展示：
binlog文件：mysql-bin.000102
binlog位点：1142900
binlog时间戳：2022-11-25 15:54:07 (1669362847)
距离上次更新： 2秒

### DIFF
--- a/mysql/position.go
+++ b/mysql/position.go
@@ -10,6 +10,8 @@ import (
 type Position struct {
 	Name string
 	Pos  uint32
+
+	Timestamp uint32 // for canal handler to show delay time
 }
 
 func (p Position) Compare(o Position) int {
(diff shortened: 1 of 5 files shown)


mysql/position.go
```diff
@@ -10,6 +10,8 @@ import (
 type Position struct {
```
lance6716, line 10: this structure means binlog position which can start a position-based replication, we should not add other fields. Maybe you can add a new interface `EventHandlerV2` with `OnXID(nextPos mysql.Position, xidEvent *replication.XIDEvent) error`.

And please translate the PR description to English so I can refer non-Chinese reviewers

PR #740, opened 2022-11-25
EventHandlerV2 support handle event with replication.EventHeader

BLAZZ:
add EventHandlerV2 to get timestamp from event, so that canal can show the how much time delay when process like

```go
func (s *handler) OnXID(header *replication.EventHeader, nextPos mysql.Position) error {
	s.queue <- model.Position{
		Name:      nextPos.Name,
		Pos:       nextPos.Pos,
		Timestamp: header.Timestamp,
		Force:     false,
	}
	return nil
}
```

it can show on UI like
binlog file: mysql-bin.000102
binlog position: 1142900
**binlog timestamp: 2022-11-25 15:54:07 (1669362847)**
**tiem delay:  2 sec**

### DIFF
--- a/canal/handler.go
+++ b/canal/handler.go
@@ -6,43 +6,6 @@ import (
 )
 
 type EventHandler interface {
-	OnRotate(rotateEvent *replication.RotateEvent) error
-	// OnTableChanged is called when the table is created, altered, renamed or dropped.
-	// You need to clear the associated data like cache with the table.
-	// It will be called before OnDDL.
-	OnTableChanged(schema string, table string) error
-	OnDDL(nextPos mysql.Position, queryEvent *replication.QueryEvent) error
-	OnRow(e *RowsEvent) error
-	OnXID(nextPos mysql.Position) error
-	OnGTID(gtid mysql.GTIDSet) error
-	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
-	OnPosSynced(pos mysql.Position, set mysql.GTIDSet, force bool) error
-	String() string
-}
-
-type DummyEventHandler struct {
-}
-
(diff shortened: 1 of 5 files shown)


atercattus:
Maybe we should change an `EventHandler` and increment semver as a major update?

BLAZZ:
> Maybe we should change an `EventHandler` and increment semver as a major update?

change EventHandlerV2 name to another like EventHandlerWithHeader or change the method in EventHandler ?

lance6716:
> Maybe we should change an `EventHandler` and increment semver as a major update?

Oh i forget that for major version updates we can break compatibility. I agree with this.



> > Maybe we should change an `EventHandler` and increment semver as a major update?
> 
> change EventHandlerV2 name to another like EventHandlerWithHeader or change the method in EventHandler ?

It means we can change the API interface while keeping the old name EventHandler. 

But because this chance is rare, we might ask other developers what they want to change and do them at once

canal/handler.go
```diff
@@ -40,3 +40,24 @@ func (h *DummyEventHandler) String() string { return "DummyEventHandler" }
 func (c *Canal) SetEventHandler(h EventHandler) {
 	c.eventHandler = h
 }
+
+// EventHandlerV2 can process event with replication.EventHeader
+type EventHandlerV2 interface {
```
lance6716, line 45: sorry for my previous comments, now I think we can change the old `EventHandler` to this interface (will review more carefully later)

BLAZZ: I have changed EventHandlerV2 to EventHandler

lance6716:
please fix CI, some test files need to be updated

lance6716 (approved):

PR #741, opened 2022-11-26
support download flushed binlog and parse event for cloud comput…

BLAZZ:
in cloud computing paltform, binlog may flush to cloud storage, that will cause 'not find first log', this change support to consume the position from a flushed binlog

for aliyun example
```go
type RDSBinlogDownloader struct {
	Endpoint   string
	AccessKey  string
	KeySecret  string
	InstanceId string
	UsePubNet  bool
}

func (r *RDSBinlogDownloader) DownloaderFunc() canal.BinlogFileDownload {
	return func(pos mysql.Position) (path string, err error) {
		return r.downBinlogFile(pos.Name)
	}
}

func (r *RDSBinlogDownloader) downBinlogFile(fileName string) (string, error) {
	//	call rds api to find binlog file by name and download it to local filesytem
	...
}
```

use RDSBinlogDownloader
```go
var current *model.Position // get from myapp's position storage
var canalErr error
if current == nil {
	if t.rdsDownload != nil {
		t.canal.WithLocalBinlogDownload(t.rdsDownload)
	}
	canalErr = t.canal.Run()
} else {
	myPos := mysql.Position{
		Name: current.Name,
		Pos:  current.Pos,
	}
	if t.rdsDownload != nil {
		t.canal.WithLocalBinlogDownload(t.rdsDownload)
	}
	canalErr = t.canal.RunFrom(myPos)
}
```

### DIFF
--- a/canal/local.go
+++ b/canal/local.go
@@ -1,0 +1,105 @@
+package canal
+
+import (
+	"context"
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/replication"
+	"github.com/pingcap/errors"
+)
+
+// BinlogFileDownload download the binlog file from cloud computing platform (etc. aliyun)
+type BinlogFileDownload func(mysql.Position) (localBinFilePath string, err error)
+
+// WithLocalBinlogDownload registers the local bin file download,
+// that allows download the flushed binlog file to local (etc. aliyun)
+func (c *Canal) WithLocalBinlogDownload(d BinlogFileDownload) {
+	c.binFileDownload = d
+}
+
+func (c *Canal) adaptLocalBinFileStreamer(syncMasterStreamer *replication.BinlogStreamer, err error) (*LocalBinFileAdapterStreamer, error) {
+	return &LocalBinFileAdapterStreamer{
+		BinlogStreamer:     syncMasterStreamer,
+		syncMasterStreamer: syncMasterStreamer,
+		canal:              c,
+		binFileDownload:    c.binFileDownload,
+	}, err
+}
+
+// LocalBinFileAdapterStreamer will support to download flushed binlog file for continuous sync in cloud computing platform
+type LocalBinFileAdapterStreamer struct {
+	*replication.BinlogStreamer                             // the running streamer, it will be localStreamer or sync master streamer
+	syncMasterStreamer          *replication.BinlogStreamer // syncMasterStreamer is the streamer from startSyncer
+	canal                       *Canal
+	binFileDownload             BinlogFileDownload
+}
+
+// GetEvent will auto switch  the running streamer and return replication.BinlogEvent
+func (s *LocalBinFileAdapterStreamer) GetEvent(ctx context.Context) (*replication.BinlogEvent, error) {
+	if s.binFileDownload == nil { // not support to use local bin file
+		return s.BinlogStreamer.GetEvent(ctx)
+	}
+
+	ev, err := s.BinlogStreamer.GetEvent(ctx)
+
+	if err == nil {
+		switch ev.Event.(type) {
+		case *replication.RotateEvent: // RotateEvent means need to change steamer back to sync master to retry sync
+			s.BinlogStreamer = s.syncMasterStreamer
+		}
+		return ev, err
+	}
+
+	if err == replication.ErrNeedSyncAgain { // restart master if last sync master syncer has error
+		s.canal.syncer.Close()
+		_ = s.canal.prepareSyncer()
+
+		newStreamer, startErr := s.canal.startSyncer()
+		if startErr == nil {
+			ev, err = newStreamer.GetEvent(ctx)
+		}
+		// set all streamer to the new sync master streamer
+		s.BinlogStreamer = newStreamer
+		s.syncMasterStreamer = newStreamer
+	}
+
+	if mysqlErr, ok := err.(*mysql.MyError); ok {
+		// change to local binlog file streamer to adapter the steamer
+		if mysqlErr.Code == mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG &&
+			mysqlErr.Message == "Could not find first log file name in binary log index file" {
+			gset := s.canal.master.GTIDSet()
+			if gset == nil || gset.String() == "" { // currently only support xid mode
+				s.canal.cfg.Logger.Info("Could not find first log, try to download the local binlog for retry")
+				pos := s.canal.master.Position()
+				newStreamer := newLocalBinFileStreamer(s.binFileDownload, pos)
+
+				s.syncMasterStreamer = s.BinlogStreamer
+				s.BinlogStreamer = newStreamer
+
+				return newStreamer.GetEvent(ctx)
+			}
+		}
+	}
+
+	return ev, err
+}
+
+func newLocalBinFileStreamer(download BinlogFileDownload, position mysql.Position) *replication.BinlogStreamer {
+	streamer := replication.NewBinlogStreamer()
+	binFilePath, err := download(position)
+	if err != nil {
+		streamer.CloseWithError(errors.New("local binlog file not exist"))
+	}
+
+	beginFromHere := false
+	go replication.NewBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
+		if be.Header.LogPos == position.Pos || position.Pos == 4 { // go ahead to check if begin
+			beginFromHere = true
+		}
+		if beginFromHere {
+			streamer.PutEvent(be)
+		}
+		return nil
+	})
+
+	return streamer
+}
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -52,6 +52,8 @@ type Canal struct {
 
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	binFileDownload BinlogFileDownload
 }
 
 // canal will retry fetching unknown table's meta after UnknownTableRetryPeriod
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -35,7 +35,7 @@ func (c *Canal) startSyncer() (*replication.BinlogStreamer, error) {
 }
 
 func (c *Canal) runSyncBinlog() error {
-	s, err := c.startSyncer()
+	s, err := c.adaptLocalBinFileStreamer(c.startSyncer())
 	if err != nil {
 		return err
 	}
--- a/replication/binlogstreamer.go
+++ b/replication/binlogstreamer.go
@@ -92,3 +92,16 @@ func newBinlogStreamer() *BinlogStreamer {
 
 	return s
 }
+
+// PutEvent puts event to BinlogStreamer
+func (s *BinlogStreamer) PutEvent(ev *BinlogEvent) {
+	s.ch <- ev
+}
+
+func (s *BinlogStreamer) CloseWithError(err error) {
+	s.closeWithError(err)
+}
+
+func NewBinlogStreamer() *BinlogStreamer {
+	return newBinlogStreamer()
+}


canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/replication"
+	"github.com/pingcap/errors"
+)
+
+// BinlogFileDownload download the binlog file from cloud computing platform (etc. aliyun)
+type BinlogFileDownload func(mysql.Position) (localBinFilePath string, err error)
```
lance6716, line 12:
```suggestion
// BinlogFileDownloader downloads the binlog file and return the path to it. It's often used to download binlog backup from RDS service.
type BinlogFileDownloader func(mysql.Position) (localBinFilePath string, err error)
```

and please change other comments

canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+
+// WithLocalBinlogDownload registers the local bin file download,
+// that allows download the flushed binlog file to local (etc. aliyun)
+func (c *Canal) WithLocalBinlogDownload(d BinlogFileDownload) {
+	c.binFileDownload = d
+}
+
+func (c *Canal) adaptLocalBinFileStreamer(syncMasterStreamer *replication.BinlogStreamer, err error) (*LocalBinFileAdapterStreamer, error) {
```
lance6716, line 20: how about use `remoteBinlogStreamer` rather than `syncMasterStreamer`? local vs remote

canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+	return &LocalBinFileAdapterStreamer{
+		BinlogStreamer:     syncMasterStreamer,
+		syncMasterStreamer: syncMasterStreamer,
+		canal:              c,
+		binFileDownload:    c.binFileDownload,
+	}, err
+}
+
+// LocalBinFileAdapterStreamer will support to download flushed binlog file for continuous sync in cloud computing platform
+type LocalBinFileAdapterStreamer struct {
```
lance6716, line 30: no need to be public?

canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+	*replication.BinlogStreamer                             // the running streamer, it will be localStreamer or sync master streamer
+	syncMasterStreamer          *replication.BinlogStreamer // syncMasterStreamer is the streamer from startSyncer
+	canal                       *Canal
+	binFileDownload             BinlogFileDownload
+}
+
+// GetEvent will auto switch  the running streamer and return replication.BinlogEvent
```
lance6716, line 37:
```suggestion
// GetEvent will auto switch the local and remote streamer to get binlog event if possible.
```

canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+	if err == replication.ErrNeedSyncAgain { // restart master if last sync master syncer has error
+		s.canal.syncer.Close()
+		_ = s.canal.prepareSyncer()
+
+		newStreamer, startErr := s.canal.startSyncer()
+		if startErr == nil {
+			ev, err = newStreamer.GetEvent(ctx)
+		}
+		// set all streamer to the new sync master streamer
+		s.BinlogStreamer = newStreamer
```
lance6716, line 62: if `startErr` is not nil, `newStreamer` will be nil and future s.BinlogStreamer.GetEvent will panic

canal/local.go
```diff
@@ -0,0 +1,108 @@
...
+		s.syncMasterStreamer = newStreamer
+	}
+
+	if mysqlErr, ok := err.(*mysql.MyError); ok {
+		// change to local binlog file streamer to adapter the steamer
+		if mysqlErr.Code == mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG &&
+			mysqlErr.Message == "Could not find first log file name in binary log index file" {
+			gset := s.canal.master.GTIDSet()
+			if gset == nil || gset.String() == "" { // currently only support xid mode
```
lance6716, line 71: what is xid mode? do you mean "position based replication"?

canal/local.go
BLAZZ: I have changed the function name, please review again

canal/local.go
BLAZZ: I only change the function args name, is that ok?

canal/local.go
BLAZZ: comments have changed, please review again

canal/local.go
BLAZZ: yes, position based replication. comments have changed, please review again

canal/local.go
```diff
@@ -0,0 +1,109 @@
...
+func newLocalBinFileStreamer(download BinlogFileDownloader, position mysql.Position) *replication.BinlogStreamer {
+	streamer := replication.NewBinlogStreamer()
+	binFilePath, err := download(position)
+	if err != nil {
+		streamer.CloseWithError(errors.New("local binlog file not exist"))
+	}
+
+	go func(binFilePath string, streamer *replication.BinlogStreamer) {
+		beginFromHere := false
+		_ = replication.NewBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
```
lance6716, line 97: what if we meet parsing error, and the new `streamer` always stuck because no event is put?
And if the we meet parsing error in the middle of binlog file, the `RotateEvent` is not sent to `(*localBinFileAdapterStreamer).GetEvent`, we can't recover to normal routine?

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+		if startErr != nil {
+			return nil, startErr
+		}
+		ev, err = newStreamer.GetEvent(ctx)
+		// set all streamer to the new sync master streamer
+		s.BinlogStreamer = newStreamer
+		s.syncMasterStreamer = newStreamer
+	}
+
+	if mysqlErr, ok := err.(*mysql.MyError); ok {
```
lance6716, line 67: there're too many ident levels, I prefer we check the fail path and return for them

```
..., ok := ...
if !ok {
  return ...
}
```

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err == replication.ErrNeedSyncAgain { // restart master if last sync master syncer has error
+		s.canal.syncer.Close()
+		_ = s.canal.prepareSyncer()
+
+		newStreamer, startErr := s.canal.startSyncer()
+		if startErr != nil {
+			return nil, startErr
+		}
+		ev, err = newStreamer.GetEvent(ctx)
```
lance6716, line 61: I prefer we move it under 

```
s.syncMasterStreamer = newStreamer
```

as in line 78~80

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+		// change to local binlog file streamer to adapter the steamer
+		if mysqlErr.Code == mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG &&
+			mysqlErr.Message == "Could not find first log file name in binary log index file" {
+			gset := s.canal.master.GTIDSet()
+			if gset == nil || gset.String() == "" { // currently only support position based replication
+				s.canal.cfg.Logger.Info("Could not find first log, try to download the local binlog for retry")
+				pos := s.canal.master.Position()
+				newStreamer := newLocalBinFileStreamer(s.binFileDownloader, pos)
+
+				s.syncMasterStreamer = s.BinlogStreamer
```
lance6716, line 77: no need to change it?

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return ev, err
+}
+
+func newLocalBinFileStreamer(download BinlogFileDownloader, position mysql.Position) *replication.BinlogStreamer {
+	streamer := replication.NewBinlogStreamer()
+	binFilePath, err := download(position)
+	if err != nil {
+		streamer.CloseWithError(errors.New("local binlog file not exist"))
```
lance6716, line 92: forget return? otherwise empty `binFilePath` will be used in line 97

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	go func(binFilePath string, streamer *replication.BinlogStreamer) {
+		beginFromHere := false
+		err := replication.NewBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
+			if be.Header.LogPos == position.Pos || position.Pos == 4 { // go ahead to check if begin
```
lance6716, line 98: why do we need the special check of `position.Pos == 4`?

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+func newLocalBinFileStreamer(download BinlogFileDownloader, position mysql.Position) *replication.BinlogStreamer {
+	streamer := replication.NewBinlogStreamer()
+	binFilePath, err := download(position)
+	if err != nil {
+		streamer.CloseWithError(errors.New("local binlog file not exist"))
+	}
+
+	go func(binFilePath string, streamer *replication.BinlogStreamer) {
+		beginFromHere := false
+		err := replication.NewBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
```
lance6716, line 97: the real MySQL syncer has many configurations that can be set by user, this BinlogParser should also set them

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		ev, err = newStreamer.GetEvent(ctx)
+		// set all streamer to the new sync master streamer
+		s.BinlogStreamer = newStreamer
+		s.syncMasterStreamer = newStreamer
+	}
+
+	if mysqlErr, ok := err.(*mysql.MyError); ok {
+		// change to local binlog file streamer to adapter the steamer
+		if mysqlErr.Code == mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG &&
```
lance6716, line 69: If the binlog file is purged just at the time we are switching to it, should we also generate a "fake rotate evnet" as the real binlog streamer?

https://github.com/mysql/mysql-server/blob/4f1d7cf5fcb11a3f84cff27e37100d7295e7d5ca/sql/rpl_binlog_sender.cc#L248

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	go func(binFilePath string, streamer *replication.BinlogStreamer) {
+		beginFromHere := false
+		err := replication.NewBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
+			if be.Header.LogPos == position.Pos || position.Pos == 4 { // go ahead to check if begin
+				beginFromHere = true
+			}
+			if beginFromHere {
+				streamer.PutEvent(be)
```
lance6716, line 102: there's more logic than simply putting the parsed event into channel, please refer to (*BinlogSyncer).parseEvent

lance6716 (changes requested):

canal/local.go
BLAZZ: mysql binlog first position number is 4, use position.Pos == 4 means need to process whole binlog file

canal/local.go
BLAZZ: Is (*BinlogSyncer).parseEvent modified the fields of BinlogSyncer itself only? I think the only change on BinlogEvent called by parseEvent is 'event.GSet = getCurrentGtidSet()', but currently only support position based mode, so not process gtid maybe OK?

canal/local.go
BLAZZ: if syncer switch to a purged  binlog file, 'not find first log file' happen and syncer closed. it try to download binlog to local.  **the local file contains rotate event**, and will put the rotate event to streamer. then canal's masterInfo will change to the new position and it restart syncer to try if new position file is on master,  if not the new started syncer will also closed by  'not find first log file' and download next  binlog to local. so  generate a "fake rotate evnet" maybe not necessary?

canal/local.go
lance6716: Oh, I forget there's no event that the `Header.LogPos` is 4.

And maybe we should also check the position <= 4?

canal/local.go
lance6716: I mean the "fake rorate event" which comes before the beginning of any events in binlog, not the real rotate event at the end of each binlog file. You can find it when reading binlog

```
=== RotateEvent ===
Date: 2022-12-08 11:13:59
Log position: 1969
Event size: 47
Position: 4
Next log name: mysql-bin.000002

=== RotateEvent ===
Date: 1970-01-01 08:00:00
Log position: 0
Event size: 47
Position: 4
Next log name: mysql-bin.000002
```

for above two events, the first one is the real rotate event which can be found at the end of mysql-bin.000001, and the second one is a fake rotate event

canal/local.go
```diff
@@ -0,0 +1,112 @@
...
+		s.BinlogStreamer = newStreamer
+		s.syncMasterStreamer = newStreamer
+	}
+
+	if mysqlErr, ok := err.(*mysql.MyError); ok {
+		// change to local binlog file streamer to adapter the steamer
+		if mysqlErr.Code == mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG &&
+			mysqlErr.Message == "Could not find first log file name in binary log index file" {
+			gset := s.canal.master.GTIDSet()
+			if gset == nil || gset.String() == "" { // currently only support position based replication
```
lance6716, line 72: `gset.String() == ""` may means empty GTID set, which is a legal GTID replication status

canal/local.go
BLAZZ: if position <= 4 not put any event to channel is OK?

canal/local.go
```diff
@@ -0,0 +1,117 @@
...
+		err := s.canal.syncer.GetBinlogParser().ParseFile(binFilePath, 0, func(be *replication.BinlogEvent) error {
+			if position.Pos < 4 { // binlog first pos is 4, if pos < 4 means canal gives error position info
+				return nil
+			}
+			if be.Header.LogPos == position.Pos || position.Pos == 4 { // go ahead to check if begin
+				beginFromHere = true
+			}
+			if beginFromHere {
+				if err := s.canal.syncer.StorePosAndGTID(be); err != nil {
+					streamer.CloseWithError(err)
```
lance6716, line 104: can return `err` to let caller `streamer.CloseWithError(err)` in line 112

canal/local.go
```diff
@@ -0,0 +1,117 @@
...
+	// only 'Could not find first log' can create local streamer, ignore other errors
+	if !ok || mysqlErr.Code != mysql.ER_MASTER_FATAL_ERROR_READING_BINLOG ||
+		mysqlErr.Message != "Could not find first log file name in binary log index file" {
+		return ev, err
+	}
+
+	s.canal.cfg.Logger.Info("Could not find first log, try to download the local binlog for retry")
+
+	// local binlog need next position to find binlog file and begin event
+	pos := s.canal.master.Position()
```
lance6716, line 78: so we don't check it's a GTID-based replication? For GTID-based replication, filename and position is not dependable to reset replication because upstream master may failover to a new MySQL node and they have same binlog filename but different position.

canal/local.go
lance6716: "fake rorate event" does not exist in binlog file but it is sent to downstream and can be find when reading binlog 😅

BLAZZ: when using local binlog, the real syncer is closed, canal will not read event from server's binlog, all event read from AdapterStreamer is parsed from binlog file, that will not exist "fake rorate event". "fake rorate event" will send to streamer after finish parse binlog file and restart the read syncer, and runSyncBinlog will process that event

BLAZZ:
> any progress for this support?

you may try [gitee.com/Blazz/go-mysql-plus.git](https://gitee.com/Blazz/go-mysql-plus.git).
our company has run for this feature in production for almost one year, it works OK.

PS: cloud computing platform will auto switch master to salver and raise `local binlog file not exist` error.
you should record the GTID and using 'SHOW BINLOG EVENTS IN xxx' to find the next position after `canal.RunFrom` stop and restart canal if found the next position

dveeden:
@BLAZZ I would like to get this merged. Could you check the open conversations and fix the conflicts?

dveeden:
@BLAZZ Any plans to continue with this? Need any help?

PR #744, opened 2022-11-26
support to check if a column is primary key

fungaren:


schema/schema.go
```diff
@@ -183,10 +183,21 @@ func (ta *Table) FindColumn(name string) int {
 	return -1
 }
 
+// Get TableColumn by Index index
 func (ta *Table) GetPKColumn(index int) *TableColumn {
 	return &ta.Columns[ta.PKColumns[index]]
 }
 
+// Get Index by column index. Return nil if the column is not a index.
```
lance6716, line 191: > Get Index by column index

What if the column has multiple indexes?

fungaren: Yes, I think this function is to get primary index only. `PKColumns` must be `PRIMARY`. I Will update.

schema/schema.go
```diff
@@ -183,10 +183,21 @@ func (ta *Table) FindColumn(name string) int {
 	return -1
 }
 
+// Get TableColumn by Index index
 func (ta *Table) GetPKColumn(index int) *TableColumn {
 	return &ta.Columns[ta.PKColumns[index]]
 }
 
+// Get Primary Index by column index. Return nil if the column is not a primary index.
+func (ta *Table) GetPKIndex(colIndex int) *Index {
```
lance6716, line 192: I'm not sure what's your purpose. If you want to get PK, you can add a new field just like `Table .PKColumns`

Also please change schema_test.go to add a test covering your change

fungaren: This function is very useful when you want to check if a column is a primary key. For example, when you get a `canal.RowsEvent`, and you may want to find out the primary key in the row. However, what we have currently is just `Table.PKColumns`, which is a slice of column idx for the table `Index`.

https://github.com/go-mysql-org/go-mysql/blob/09995299e84479a5bc7c65e75e11d8eb60f1335a/schema/schema.go#L423

lance6716: The usage of this function "Get Primary Index by column index." is very weird, primary key is a property of table, column is a property of table or index, it's strange to do reverse thing. I'm considering to return a bool to indicate if a column is in a primary key, but that's too trivial to add a public function, one line `slices.Contains(Table.PKColumns, colIndex)` should be enough.

fungaren: Yes I agree with that it is enough to return a boolean. And the existing `GetPKColumn()` is also weird, haha.

And `slices.Contains` requires an extra import which is not necessary if you just need this simple function.

I also added unit test for the existing `GetPKColumn()`.

schema/schema.go
```diff
@@ -183,10 +183,23 @@ func (ta *Table) FindColumn(name string) int {
 	return -1
 }
 
+// Get TableColumn by primary index index
```
lance6716, line 186:
```suggestion
// Get TableColumn by column index of primary key.
```

schema/schema.go
```diff
@@ -183,10 +183,23 @@ func (ta *Table) FindColumn(name string) int {
...
 func (ta *Table) GetPKColumn(index int) *TableColumn {
+	if index >= len(ta.PKColumns) {
+		return nil
+	}
 	return &ta.Columns[ta.PKColumns[index]]
 }
 
+func (ta *Table) IsPrimaryKey(colIndex int) bool {
+	for _, _colIndex := range ta.PKColumns {
```
lance6716, line 195: maybe you can just use `i` instead of `_colIndex`, we don't have `_xxx` stype in other codes.

lance6716 (commented):
rest lgtm

schema/schema.go
fungaren: done

lance6716 (approved):

PR #746, opened 2022-11-29
binlog_syncer: reduce the times of Clone of MySQLGTIDSet to speed up

GMHDBJD:
- add DiscardGTIDSet to allow user not to maintain Gset for QueryEvent/XIDEvent
- add prevMySQLGTIDEvent to calculate prevGSet instead of clone currGSet

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -831,13 +831,12 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
 		}
 		prev := b.currGset.Clone()
 		err = b.currGset.(*MariadbGTIDSet).AddSet(&event.GTID)
-		if err == nil {
-			// right after reconnect we will see same gtid as we saw before, thus currGset will not get changed
-			if !b.currGset.Equal(prev) {
-				b.prevGset = prev
-			}
-		} else {
+		if err != nil {
 			return errors.Trace(err)
+		}
+		// right after reconnect we will see same gtid as we saw before, thus currGset will not get changed
+		if !b.currGset.Equal(prev) {
+			b.prevGset = prev
 		}
 	case *XIDEvent:
 		if !b.cfg.DiscardGTIDSet {
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -120,6 +120,8 @@ type BinlogSyncerConfig struct {
 	Dialer client.Dialer
 
 	RowsEventDecodeFunc func(*RowsEvent, []byte) error
+
+	DiscardGTIDSet bool
 }
 
 // BinlogSyncer syncs binlog event from server.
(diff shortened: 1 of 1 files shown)


replication/binlogsyncer.go
```diff
@@ -830,24 +815,37 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
...
+			b.prevGset.(*MysqlGTIDSet).AddGTID(u, b.prevMySQLGTIDEvent.GNO)
 		}
+		b.prevMySQLGTIDEvent = event
 	case *MariadbGTIDEvent:
 		if b.prevGset == nil {
 			break
 		}
-		GTID := event.GTID
-		err := advanceCurrentGtidSet(uuid.Nil, 0, GTID.DomainID, GTID.ServerID, GTID.SequenceNumber)
+		prev := b.currGset.Clone()
```
lance6716, line 832: why we don't do it for MariaDB?

replication/binlogsyncer.go
```diff
@@ -120,6 +120,8 @@ type BinlogSyncerConfig struct {
 	Dialer client.Dialer
 
 	RowsEventDecodeFunc func(*RowsEvent, []byte) error
+
+	DiscardGTIDSet bool
```
lance6716, line 124: I think the DiscardGTIDSet should also skip maintaining `prevGset` and `currGset`, only skipping output is no benefit to performance

replication/binlogsyncer.go
```diff
@@ -830,24 +815,37 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
...
+			b.currGset = b.prevGset.Clone()
+		}
 		u, _ := uuid.FromBytes(event.SID)
-		err := advanceCurrentGtidSet(u, event.GNO, 0, 0, 0)
-		if err != nil {
-			return errors.Trace(err)
+		b.currGset.(*MysqlGTIDSet).AddGTID(u, event.GNO)
+		if b.prevMySQLGTIDEvent != nil {
+			u, _ = uuid.FromBytes(b.prevMySQLGTIDEvent.SID)
+			b.prevGset.(*MysqlGTIDSet).AddGTID(u, b.prevMySQLGTIDEvent.GNO)
```
lance6716, line 825: should we also consider the case "right after reconnect we will see same gtid as we saw before, thus currGset will not get changed"? (but I don't know when it will happen)

replication/binlogsyncer.go
GMHDBJD: prevGset and currGset need to maintain for reconnect. Skip output reduce the clone operation

replication/binlogsyncer.go
GMHDBJD: That won't happen, because we will leave the currSet empty when we retry, and the mariadb comment will just stay as it is.
https://github.com/go-mysql-org/go-mysql/blob/7685eee9a345f53977d3c42e1aa220e0534f5cbf/replication/binlogsyncer.go#L664

replication/binlogsyncer.go
lance6716: OK seems MariaDB has a simpler representation of GTID, we can leave this improvement to future when we found a bottleneck

lance6716 (approved):

PR #748, opened 2022-12-05
modify the type of Position

liumengya94:
The `Pos` should be `uint64` instead of `uint32`, to conform to MySQL's behavior:
https://github.com/mysql/mysql-server/blob/a246bad76b9271cb4333634e954040a970222e0a/sql/rpl_source.cc#L1225-L1226

### DIFF
--- a/canal/dump.go
+++ b/canal/dump.go
@@ -173,7 +173,7 @@ func (c *Canal) dump() error {
 		return errors.Trace(err)
 	}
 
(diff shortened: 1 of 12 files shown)


lance6716:
it's 4 bytes indeed https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_binlog_dump.html

Issue #754, opened 2022-12-13
canal can't parse value of ddl


BLAZZ:
canal.valueExpr can't work
```sql
ALTER TABLE `sometable` MODIFY COLUMN `some_col` varchar(1024) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL DEFAULT '' COMMENT 'comments' AFTER `other_col`
```
become
```sql
ALTER TABLE `sometable` MODIFY COLUMN `some_col` VARCHAR(1024) CHARACTER SET UTF8 COLLATE utf8_bin NOT NULL DEFAULT  COMMENT  AFTER `other_col`
```
after (*ast.AlterTableStmt).Restore(ctx *format.RestoreCtx), default value and comment value lost

change "github.com/pingcap/parser" test_driver to public or copy the implements to canal?

lance6716:
I didn't find where go-mysql uses `(*ast.AlterTableStmt).Restore(ctx *format.RestoreCtx)`, can you explain it?

BLAZZ:
AlterTableStmt is parsed from replication.QueryEvent, and send to eventHandler.OnDDL, use (*ast.AlterTableStmt).Restore(ctx *format.RestoreCtx) is easy to sync ddl to  the target database when using canal

lance6716:
So where go-mysql uses `(*ast.AlterTableStmt).Restore(ctx *format.RestoreCtx)`? If it's used in your application, you can change your code rather than go-mysql

PR #755, opened 2022-12-13
change valueExpr to pingcap test_driver implements

BLAZZ:
fix valueExpr not work call Restore(ctx *format.RestoreCtx) #754

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/BurntSushi/toml v0.3.1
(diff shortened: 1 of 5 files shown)


lance6716 (commented):
This PR is very complicated 😂  can you refer to where did you copy the code in comment, and add a test for it?

BLAZZ:
those code is from [https://github.com/pingcap/parser/tree/master/test_driver](https://github.com/pingcap/parser/tree/master/test_driver)

PR #759, opened 2022-12-22
Adding replication protocol support to mysql server implementation

Fizic:


mysql/errcode.go
```diff
@@ -867,4 +867,5 @@ const (
 	ER_MUST_CHANGE_PASSWORD_LOGIN                                       = 1862
 	ER_ROW_IN_WRONG_PARTITION                                           = 1863
 	ER_ERROR_LAST                                                       = 1863
+	ER_HANDLER_DOES_NOT_SUPPORT                                         = 1864
```
lance6716, line 870: Hi, these error codes are from MySQL server, should not add customised error here

https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html

replication/binlogstreamer.go
```diff
@@ -84,11 +84,24 @@ func (s *BinlogStreamer) closeWithError(err error) {
...
+	select {
+	case s.ch <- ev:
+		return nil
+	case err := <-s.ech:
+		return err
+	}
+}
+
+func (s *BinlogStreamer) AddErrorToStreamer(err error) {
+	s.ech <- err
```
lance6716, line 106: it's a bit dangerous that s.ech is full so it will be blocked. we can use `select` to avoid blocking. And if the developer cares about the success of appending we can add a bool return value

server/command.go
```diff
@@ -30,6 +31,13 @@ type Handler interface {
 	HandleOtherCommand(cmd byte, data []byte) error
 }
 
+type ReplicationHandler interface {
+	//handle Replication command
```
lance6716, line 35:
```suggestion
	// handle Replication command
```

server/command.go
```diff
@@ -131,6 +139,39 @@ func (c *Conn) dispatch(data []byte) interface{} {
 		}
 
 		return eofResponse{}
+	case COM_REGISTER_SLAVE:
+		if h, ok := c.h.(ReplicationHandler); ok {
+			if err := h.HandleRegisterSlave(data); err != nil {
```
lance6716, line 144:
```suggestion
		return h.HandleRegisterSlave(data)
```

server/replication.go
```diff
@@ -0,0 +1,21 @@
...
+
+func parseBinlogDump(data []byte) (*mysql.Position, error) {
+	var p mysql.Position
+	p.Pos = binary.LittleEndian.Uint32(data[0:4])
+	p.Name = string(data[10:])
+
+	return &p, nil
+}
+
+func parseBinlogDumpGTID(data []byte) (*mysql.MysqlGTIDSet, error) {
```
lance6716, line 17: do you plan to support MariaDB GTID set? you can see `ParseMariadbGTID`

server/resp.go
```diff
@@ -197,6 +200,25 @@ func (c *Conn) writeFieldValues(fv []FieldValue) error {
 	return c.WritePacket(data)
 }
 
+func (c *Conn) writeBinlogEvents(s *replication.BinlogStreamer) error {
+	for {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		ev, err := s.GetEvent(ctx)
+		cancel()
+
+		if err == context.DeadlineExceeded {
```
lance6716, line 209: Do we really need a context timeout? it just keeps checking `GetEvent`

server/resp.go
```diff
@@ -197,6 +200,25 @@ func (c *Conn) writeFieldValues(fv []FieldValue) error {
...
+			continue
+		}
+		data := make([]byte, 4, 32)
```
lance6716, line 212: we can calculate the buffer capacity by looking at length of `ev.RawData`

server/resp.go
```diff
@@ -197,6 +200,25 @@ func (c *Conn) writeFieldValues(fv []FieldValue) error {
...
+		data = append(data, OK_HEADER)
+
+		data = append(data, ev.RawData...)
```
lance6716, line 215: and please add some reference to binlog protocal, such as https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_replication.html

server/command.go
```diff
@@ -139,6 +180,10 @@ func (c *Conn) dispatch(data []byte) interface{} {
 type EmptyHandler struct {
 }
 
+type EmptyReplicationHandler struct {
```
lance6716, line 183: you can add

```
var _ ReplicationHandler = EmptyReplicationHandler{}
```

to command_test.go

replication/binlogstreamer.go
```diff
@@ -84,11 +84,24 @@ func (s *BinlogStreamer) closeWithError(err error) {
...
+func NewBinlogStreamer() *BinlogStreamer {
 	s := new(BinlogStreamer)
 
 	s.ch = make(chan *BinlogEvent, 10240)
 	s.ech = make(chan error, 4)
 
 	return s
 }
+
+func (s *BinlogStreamer) AddEventToStreamer(ev *BinlogEvent) error {
```
lance6716, line 96: this is exposed to Handler? please add some comment on how to use it

server/command.go
```diff
@@ -131,6 +139,39 @@ func (c *Conn) dispatch(data []byte) interface{} {
...
+		} else {
+			return fmt.Errorf("the handler does not support replication protocol, use ReplicationHandler instead")
+		}
+	case COM_BINLOG_DUMP:
+		if h, ok := c.h.(ReplicationHandler); ok {
+			pos, _ := parseBinlogDump(data)
+			s := replication.NewBinlogStreamer()
+			go h.HandleBinlogDump(pos, s)
+
+			return s
```
lance6716, line 157: another choice is that this Conn never returns, we directly call writeBinlogEvents in a loop. I'm not sure what should we do if the Conn is returned and MySQL client send another command to this connection

server/command.go
```diff
@@ -30,6 +31,13 @@ type Handler interface {
 	HandleOtherCommand(cmd byte, data []byte) error
 }
 
+type ReplicationHandler interface {
+	//handle Replication command
+	HandleRegisterSlave(data []byte) error
+	HandleBinlogDump(pos *Position, s *replication.BinlogStreamer)
+	HandleBinlogDumpGTID(gtidSet *MysqlGTIDSet, s *replication.BinlogStreamer)
```
ostinru, line 38: I think that `s *replication.BinlogStreamer` shouldn't be here. It is up to users who will implement `ReplicationHandler` what to do with these events.

server/replication.go
ostinru: MariaDB doesn't implement COM_BINLOG_DUMP_GTID - so, we can assume that this message sent by MySQL.

https://github.com/MariaDB/server/blob/10.9/include/mysql_com.h#L118

server/command.go
ostinru: Interesting question. mysql-clients can be disconnected when they are not responsive (see [net_write_timeout](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_net_write_timeout) and [net_read_timeout](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_net_read_timeout)). We should check how replication protocol works.

lance6716: Thanks for reminding us, there're two questions

1. I wrongly thought the connection will be re-used for next command at first, and want to remind author to check what's the behaviour if the connection is both serving a binlog stream and client's request. Now I see `writeBinlogEvents` won't do that.
2. mysql client may [set heartbeat interval](https://github.com/go-mysql-org/go-mysql/blob/81966e15b9c5b1c66847a52c6f194771885ef64e/replication/binlogsyncer.go#L324-L325) and binlog streamer should append heartbeat event when the stream is idle for the interval. I think it's OK to implement it later because some use case of this feature will not let stream be idle for a long time.

server/replication.go
lance6716: Yes, I mean if there's not too much work to do we can support MariaDB by the way. But as you said MariaDB does not use COM_BINLOG_DUMP_GTID so it may not be a trivial work

server/command.go
lance6716: Sorry I don't understand, you mean we can expose `Conn` to user and user directly write binlog event as []byte to `Conn`?

server/command.go
```diff
@@ -160,6 +205,16 @@ func (h EmptyHandler) HandleStmtClose(context interface{}) error {
...
 
+func (h EmptyReplicationHandler) HandleRegisterSlave(data []byte) error {
+	return fmt.Errorf("not supported now")
+}
+
+func (h EmptyReplicationHandler) HandleBinlogDump(pos *Position, r *replication.BinlogStreamer) {
+}
+
+func (h EmptyReplicationHandler) HandleBinlogDumpGTID(gtidSet *MysqlGTIDSet, r *replication.BinlogStreamer) {
+}
```
ostinru, line 216: I suggest adding `r.close()` here, just to gracefully release resources.

server/command.go
```diff
@@ -131,6 +139,39 @@ func (c *Conn) dispatch(data []byte) interface{} {
...
 
 		return eofResponse{}
+	case COM_REGISTER_SLAVE:
+		if h, ok := c.h.(ReplicationHandler); ok {
+			if err := h.HandleRegisterSlave(data); err != nil {
+				return err
+			}
+			return nil
+		} else {
+			return fmt.Errorf("the handler does not support replication protocol, use ReplicationHandler instead")
```
ostinru, line 149: Maybe it is better to fall back to
```
return c.h.HandleOtherCommand(cmd, data)
```

server/command.go
ostinru: Oh... I worded it really badly.

I will agree with you that it is not a good idea to expose `Conn`.

ostinru: For consistency we can
* move `s *replication.BinlogStreamer` to return values
* add `error` to return values
* remove `go` in `Conn.dispatch()`... or document that this method will be called form its own coroutine.

server/replication.go
```diff
@@ -0,0 +1,21 @@
+package server
+
+import (
+	"encoding/binary"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+)
+
+func parseBinlogDump(data []byte) (*mysql.Position, error) {
```
lance6716, line 9:
```suggestion
func parseBinlogDump(data []byte) (mysql.Position, error) {
```

in other place of this repo we directly use the struct, maybe it's small so putting it on stack.

And seems you can check the length of data to return an error

server/replication.go
```diff
@@ -0,0 +1,21 @@
...
+	var p mysql.Position
+	p.Pos = binary.LittleEndian.Uint32(data[0:4])
+	p.Name = string(data[10:])
+
+	return &p, nil
+}
+
+func parseBinlogDumpGTID(data []byte) (*mysql.MysqlGTIDSet, error) {
+	lenPosName := binary.LittleEndian.Uint32(data[11:15])
```
lance6716, line 18: maybe also check the error of slice out of bounds

lance6716 (commented):
rest lgtm

and I remember `mysqlbinlog` can actively stop a binlog stream https://dev.mysql.com/doc/refman/8.0/en/mysqlbinlog.html#option_mysqlbinlog_stop-never I'm not sure the termination is done at client-side or server-side so need we implement more behaviour. We can do it later.

lance6716 (approved):
do you have more comments @ostinru @atercattus

ostinru:
👍 Looks good to me

Issue #650, opened 2021-12-08
binlog event PARTIAL_UPDATE_ROWS_EVENT not support


zhujingfa:
mysql version 8.0.26 Update_rows_partial binlog event not support.

ref:
https://github.com/mysql/mysql-server/blob/8.0/libbinlogevents/include/binlog_event.h#L347
https://github.com/alibaba/canal/blob/master/dbsync/src/main/java/com/taobao/tddl/dbsync/binlog/event/UpdateRowsLogEvent.java

PR #762, opened 2023-01-11
Allow to run tests in local environment + fix tests for it

atercattus:
We can run our tests on CI stage (github actions). But it's difficult for local usage 😅

Now I'm working on [#650](https://github.com/go-mysql-org/go-mysql/issues/650) and I really need tests locally.
I'm trying to use [act](https://github.com/nektos/act), but it's not working for me.
So I wrote a docker env for it. Also, I did tiny refactoring and fixes in tests for it.

All we need to run tests is simple `$ make test-local`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,16 @@ build:
 test:
 	go test --race -timeout 2m ./...
 
+test-local:
+	docker run --rm -d --network=host --name go-mysql-server \
+		-e MYSQL_ALLOW_EMPTY_PASSWORD=true \
+		-e MYSQL_DATABASE=test \
+		-v $${PWD}/docker/resources/replication.cnf:/etc/mysql/conf.d/replication.cnf \
+		mysql:5.7
+	docker/resources/waitfor.sh 127.0.0.1 3306 \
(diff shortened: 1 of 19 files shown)


docker/autotests.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apt-get update && \
+    apt-get install --yes git curl build-essential && \
+    update-ca-certificates
+RUN curl -L -o go.tar.gz https://go.dev/dl/go${GOVERSION}.linux-amd64.tar.gz && tar -C /usr/local -xzf go.tar.gz
+COPY docker/resources/waitfor.sh /
+COPY go.* ./
+RUN go mod download
+COPY ./ .
+
+CMD go test -race -v -timeout 2m ./... -gocheck.v
```
lance6716, line 17: I'm not sure about your environment. The first thought comes to my mind is using docker to start a MySQL at 3306 and running `go test` outside the docker. The good point is if I change some code, I don't need to build the docker image again and no need to prune unused docker local volumes periodically.

lance6716 (approved):

PR #767, opened 2023-02-01
Feature dump with create info

sayume:
support dump with create info


dump/parser.go
```diff
@@ -59,6 +64,25 @@ func Parse(r io.Reader, h ParseHandler, parseBinlogPos bool) error {
...
 		})
 
+		if m := tableStartExp.FindAllStringSubmatch(line, -1); len(m) == 1 {
+			query = line
+			inTableParsing = true
+			continue
+		}
+
+		if inTableParsing {
+			if m := endExp.FindAllStringSubmatch(line, -1); len(m) == 1 {
```
atercattus, line 74: Why are you using a regex instead of simple `strings.IndexByte(line, ';')`?

Also, can you add some test cases (multi-line and one-line create table, ';' inside defaults as an example) for this new code? This change doesn't look intuitive and simple.

canal/dump.go
```diff
@@ -113,6 +114,19 @@ func (h *dumpParseHandler) Data(db string, table string, values []string) error
...
+	if err := h.c.ctx.Err(); err != nil {
+		return err
+	}
+
+	event := &replication.QueryEvent{
+		Schema: []byte(db),
+		Query:  []byte(query),
+	}
+
+	return h.c.eventHandler.OnDDL(nil, mysql.Position{}, event)
```
atercattus, line 127: I don't really like that we don't pass `header` and `nextPos` to `OnDDL`. If a zero `nextPos` looks safe, then a nil `header` may break external code.

sayume: There is no header can be init here, nil is reasonable, handler should check pointer before use like OnPosSynced in dump().

dump/parser.go
sayume: resolved in [7344273](https://github.com/go-mysql-org/go-mysql/pull/767/commits/73442733bb257815fcf57c57a96c556e1054ea7e)

canal/dump.go
sayume: Add nextPos in [7344273](https://github.com/go-mysql-org/go-mysql/pull/767/commits/73442733bb257815fcf57c57a96c556e1054ea7e), no header info provided here, it's OK to set nil pointer like OnPosSynced in dump(), user should check header before use in handler func.

canal/dump.go
```diff
@@ -113,6 +114,20 @@ func (h *dumpParseHandler) Data(db string, table string, values []string) error
...
+		return err
+	}
+
+	event := &replication.QueryEvent{
+		Schema: []byte(db),
+		Query:  []byte(query),
+	}
+	pos := mysql.Position{Name: h.name, Pos: uint32(h.pos)}
+
+	return h.c.eventHandler.OnDDL(nil, pos, event)
```
lance6716, line 128: I also suppose we should provide a valid header, but if we can't set proper values, we should at least add detailed comments to the interface

dump/parser.go
```diff
@@ -85,6 +88,25 @@ func Parse(r io.Reader, h ParseHandler, parseBinlogPos bool) error {
...
 		}
 
+		if !inTableParsing {
+			if m := tableStartExp.FindAllStringSubmatch(line, -1); len(m) == 1 {
+				inTableParsing = true
+			}
+		}
+
+		if inTableParsing {
+			if i := strings.IndexByte(line, byte(';')); i > 0 {
```
lance6716, line 98: maybe you want `strings.Contains`

And what's the purpose of this logic? what if the column comments contain `;`?

sayume: resolved in [e91fc1f](https://github.com/go-mysql-org/go-mysql/pull/767/commits/e91fc1f23e4f15d5bbe30c3be88bf402beb78621)

sayume: IndexByte may be faster than strings.Contains

sayume:
It's annoying to exclude all situation in comment, you can event write insert sql in comments, I don't think parser can work in this situation.

PR #774, opened 2023-03-03
Support for PARTIAL_UPDATE_ROWS_EVENT binlog event and PARTIAL_JSON mode

atercattus:
Implementation for `PARTIAL_UPDATE_ROWS_EVENT` binlog event.

`PARTIAL_UPDATE_ROWS_EVENT` is an extension of `UPDATE_ROWS_EVENT`, allowing partial values according to [binlog_row_value_options](https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_binlog_row_value_options) in mysql8. Value `binlog_row_value_options=PARTIAL_JSON` allows for mysql [store only minimal diff from the previous column value](https://dev.mysql.com/doc/refman/8.0/en/json.html#json-partial-updates).

Tested for mysql 5.7 and 8, `binlog_row_image` `FULL` and `MINIMAL`, with `binlog_row_value_options` and without.

Known limitations:
- current implementation doesn't apply a diff to a column value. Instead `BinlogParser` returns `RowsEvent` where a field `Rows` contains instances of a new type:
```go
JsonDiff struct {
	Op    JsonDiffOperation
	Path  string
	Value string
}
```
And a user can apply a patch in any way. Why? In `binlog_row_image=MINIMAL` mode a mysql doesn't write the previous value (BEFORE_IMAGE) to binlog. So the code cannot apply the path to an unknown value.

For example:
```sql
SET SESSION binlog_row_image = 'MINIMAL';

SET SESSION binlog_row_value_options = 'PARTIAL_JSON';

INSERT INTO test_json_v2 VALUES (101, '{"a":"aaaaaaaaaaaaa", "c":"ccccccccccccccc", "ab":["abababababababa", "babababababab"]}');

UPDATE test_json_v2 SET c = JSON_SET(c, '$.ab', '["ab_updatedccc"]') WHERE id = 101;
```
After this `PARTIAL_UPDATE_ROWS_EVENT` event will contain only `enum_json_diff_operation::REPLACE`, `$.ab`, and `["ab_updatedccc"]'`. And nothing all.


lance6716:
Hi, is this still a WIP?

atercattus:
I checked it in our prod and now it looks working and correct :)

replication/row_event.go
```diff
@@ -807,27 +811,44 @@ func (e *TableMapEvent) IsEnumOrSetColumn(i int) bool {
 	return rtyp == MYSQL_TYPE_ENUM || rtyp == MYSQL_TYPE_SET
 }
 
+// JsonColumnCount returns the number of JSON columns in this table
+func (e *TableMapEvent) JsonColumnCount() uint64 {
+	e.jsonColumnCountOnce.Do(func() {
```
lance6716, line 816: Just to confirm, this `sync.Once` wants to avoid duplicate work for binlog events pattern "TableMapEvent, RowsEvent, RowsEvent, RowsEvent..." (multiple RowsEvents share same TableMapEvent)? Because I see we always create a new TableMapEvent each time we see it in binlog

https://github.com/go-mysql-org/go-mysql/blob/e3201442601efb6995888212097f03d84bf993ca/replication/parser.go#L254-L257

replication/row_event.go
```diff
@@ -856,6 +877,39 @@ type RowsEvent struct {
 	ignoreJSONDecodeErr     bool
 }
 
+// enum class enum_row_image_type { WRITE_AI, UPDATE_BI, UPDATE_AI, DELETE_BI };
```
lance6716, line 880: Does AI and BI represent "after image" and "before image"? A bit unfamiliar

replication/row_event.go
```diff
@@ -946,43 +1014,68 @@ func isBitSet(bitmap []byte, i int) bool {
...
 		if !isBitSet(bitmap, i) {
 			skips = append(skips, i)
 			continue
 		}
 
-		isNull := (uint32(nullBitmap[nullbitIndex/8]) >> uint32(nullbitIndex%8)) & 0x01
-		nullbitIndex++
-
-		if isNull > 0 {
+		if isNull := isBitSetIncr(nullBitmap, &nullbitIndex); isNull {
```
lance6716, line 1071:
```suggestion
		if isBitSetIncr(nullBitmap, &nullbitIndex) {
```

replication/row_event.go
```diff
@@ -946,43 +1014,68 @@ func isBitSet(bitmap []byte, i int) bool {
...
 	nullBitmap := data[pos : pos+count]
 	pos += count
 
+	isPartialIndex := 0
 	nullbitIndex := 0
 
-	var n int
-	var err error
 	for i := 0; i < int(e.ColumnCount); i++ {
+		isPartial := isPartialJsonUpdate &&
```
lance6716, line 1061: seems `isPartialIndex` is same as `i`? and we can move creating `isPartial` closer to `e.decodeValue(..., isPartial)`

replication/row_event.go
```diff
@@ -1161,10 +1254,23 @@ func (e *RowsEvent) decodeValue(data []byte, tp byte, meta uint16) (v interface{
...
-		d, err = e.decodeJsonBinary(data[meta:n])
-		if err == nil {
-			v = hack.String(d)
+
+		if isPartial {
+			var diff *JsonDiff
+			diff, err = e.decodeJsonPartialBinary(data[meta:n])
+			if err == nil {
+				if diff == nil {
+					v = []byte{}
```
lance6716, line 1263: Can we add a "JsonDiffOperationNoOp" JsonDiff, or tell the caller nil is no-op for JsonDiff? It's a bit ambiguous that we return a `[]byte{}`

lance6716 (commented):
rest lgtm 👍

replication/row_event.go
atercattus: >seems isPartialIndex is same as i?

No, it's an index inside a partial bitmap and we increment it only for partial fields.

>can move creating isPartial closer to e.decodeValue(..., isPartial)

[Partial bitmap checking must prepend to null bitmap checking](https://github.com/mysql/mysql-server/blob/1bfe02bdad6604d54913c62614bde57a055c8332/sql/log_event.cc#L2197-L2203).

I'll update this code for more clearance.

replication/row_event.go
atercattus: Yes, I preserve naming from [mysql sources](https://github.com/mysql/mysql-server/blob/1bfe02bdad6604d54913c62614bde57a055c8332/sql/rpl_record.h#L39). I can rename it for more clarity if you think so.

replication/row_event.go
atercattus: Ooops. I agree with this. For now, I'll just simplify the code by removing Once.

atercattus: updated

replication/row_event.go
atercattus: updated

replication/row_event.go
atercattus: MySQL doesn't allow any other variants. So I just removed a nil JsonDiff :)

replication/row_event.go
lance6716: I can see their meanings in the linked MySQL code now 😄

lance6716 (approved):

atercattus:
OK. I'll make last checks and merge it.

lance6716:
ping @atercattus looking forward to this commit

atercattus:
@lance6716 I'm done with this PR

atercattus:
I accidentally messed up the branch, but fixed it through force push :)

lance6716 (approved):

PR #777, opened 2023-04-03
canal: use localhost option

AleksandrShibanov:
according to https://github.com/go-mysql-org/go-mysql/pull/36

I use canal and need to provide localhost option to syncer stream.


lance6716:
Hi, don't worry about the CI, I'll fix it in https://github.com/go-mysql-org/go-mysql/pull/779

lance6716 (approved):

PR #780, opened 2023-04-03
fix: should use column index to set null bitmap when building binary resultset

luyui:
simple fix


lance6716:
Hi, thanks for correcting this! If you find the problem with some real payload, can you add a unit test to check the output?

lance6716 (approved):

PR #781, opened 2023-04-05
mysql8 auth compatibility

atercattus:


client/auth.go
```diff
@@ -69,35 +73,51 @@ func (c *Conn) readInitialHandshake() error {
...
+		// https://github.com/vapor/mysql-nio/blob/main/Sources/MySQLNIO/Protocol/MySQLProtocol%2BHandshakeV10.swift
+		if c.capability&CLIENT_LONG_PASSWORD != 0 {
+			// skip reserved (all [00])
+			pos += 4
+		} else {
+			// unknown
+			pos += 4
+		}
+
+		if rest := int(authPluginDataLen) - 8; rest > 0 {
```
lance6716, line 106: should we increase the rest to at least 13?

> Rest of the plugin provided data (scramble), $len=MAX(13, length of auth-plugin-data - 8)
https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_v10.html

Will take a look at free time later

atercattus: I do not have a reproduction of such a case, but yes, let it be

lance6716: Seems even if `rest := int(authPluginDataLen) - 8` and `rest < 0`, we should also increase `rest` to 13

client/conn.go
```diff
@@ -408,3 +408,129 @@ func (c *Conn) exec(query string) (*Result, error) {
 
 	return c.readResult(false)
 }
+
+func (c *Conn) CapabilityString() string {
+	var caps []string
+	capability := c.capability
+	for i := 0; capability != 0; i++ {
+		field := uint32(1 << i)
+		if capability&field == 0 {
```
lance6716, line 417: not sure if you want [low-bit trick](https://www.quora.com/What-does-x-x-lowbit-do-in-C++)

lance6716 (commented):
rest lgtm

client/conn.go
atercattus: Oh, I'm not a big fan of unnecessary tricks :) I prefer more clear code.
This code executes only one time per connection, so I don't consider it a bottleneck.

client/auth.go
atercattus: I looked into other implementations and changed the code a little. And I check compatibility on 5.7 and 8.0.

client/auth.go
```diff
@@ -69,35 +77,53 @@ func (c *Conn) readInitialHandshake() error {
...
-		resetOfAuthDataEndPos := pos + maxAuthDataLen - 8 - 1
-		c.salt = append(c.salt, data[pos:resetOfAuthDataEndPos]...)
+		if c.capability&CLIENT_SECURE_CONNECTION != 0 {
+			// Rest of the plugin provided data (scramble)
 
-		// skip reset of end pos
-		pos = resetOfAuthDataEndPos + 1
+			// $len=MAX(13, length of auth-plugin-data - 8)
+			rest := int(authPluginDataLen) - 8
+			if max := 13; rest > max {
```
lance6716, line 106:
```suggestion
			if max := 13; rest < max {
```

we should increase the length to 13 😄

client/auth.go
```diff
@@ -69,35 +77,53 @@ func (c *Conn) readInitialHandshake() error {
...
+				rest = max
+			}
+			if data[pos+rest-1] != 0 {
```
lance6716, line 109:
```suggestion
			if data[pos+rest] != 0 {
```

BTW, im not familiar with this part so double check that why there's NULL after scramble? In the MySQL protocol doc I think it's a fixed length string, rather than a NULL-terminated string.

atercattus: I found [the origin](https://github.com/mysql/mysql-server/blob/c47fab04/sql/auth/sql_authentication.cc#L584-L585) of it:
>n - rest of the plugin provided data (at least 12 bytes)
>1 - \0 byte, terminating the second part of a scramble

And [sha256_password auth implementation](https://github.com/mysql/mysql-server/blob/a153bf585e827818641fb534296be3e9a44db466/sql/auth/sql_authentication.cc#L4767-L4771) follows this rule: 
>Native authentication sent 20 bytes + '\0' character = 21 bytes.
>This plugin must do the same to stay consistent with historical behavior

Also, alibaba channel [contains the same description](https://github.com/alibaba/canal/blob/0ec46991499a22870dde4ae736b2586cbcbfea94/driver/src/main/java/com/alibaba/otter/canal/parse/driver/mysql/packets/server/HandshakeInitializationPacket.java#L94-L95):
>Packet规定最后13个byte是剩下的scrumble, 但实际上最后一个字节是0, 不应该包含在scrumble中.

*I used a translator for this.*


I can remove this check by \0, but it looks like a historic standard convention.

atercattus: What's about 20 bytes len I found [this](https://github.com/mysql/mysql-server/blob/1bfe02bdad6604d54913c62614bde57a055c8332/sql/auth/sql_authentication.cc#L1641-L1642):
>the first packet *must* have at least 20 bytes of a scramble.
>if a plugin provided less, we pad it to 20 with zeros

Right now I don't see places where it would be more than 20 bytes, but let it be for the future.

atercattus: I checked this logic for mysql 8.0 and 5.7. Unfortunately, compatibility with 5.6- was broken since we added `json` fields in our tests in 2021.

client/auth.go
```diff
@@ -69,35 +77,55 @@ func (c *Conn) readInitialHandshake() error {
...
 
-		// skip reserved (all [00])
-		pos += 10 + 1
+		// skip reserved (all [00] ?)
+		pos += 10
 
-		// auth_data is end with 0x00, so we need to trim 0x00
-		resetOfAuthDataEndPos := pos + maxAuthDataLen - 8 - 1
-		c.salt = append(c.salt, data[pos:resetOfAuthDataEndPos]...)
+		if c.capability&CLIENT_SECURE_CONNECTION != 0 {
```
lance6716, line 101: Just remind that this is a deprecated flag https://dev.mysql.com/doc/dev/mysql-server/latest/group__group__cs__capabilities__flags.html#ga8be684cc38eeca913698414efec06933 , we can skip this if-check or leave it because no test is broken.

lance6716 (approved):

client/auth.go
atercattus: Yes, it seems always set.

PR #782, opened 2023-04-08
fix: server/handshake_resp_test.go 's mysql doc link can't be opened

zhiiw:
the mysql doc link in server/handshake_resp_test.go is out of date .


lance6716 (approved):

zhiiw:
By the way ,the server/handshake_resp.go 's  test function TestReadPluginName says it use the example data in mysql doc ,but the newest doc don't have the three example data.... So how to deal with this situation ?

lance6716:
> By the way ,the server/handshake_resp.go 's test function TestReadPluginName says it use the example data in mysql doc ,but the newest doc don't have the three example data.... So how to deal with this situation ?

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_response.html#sect_protocol_connection_phase_packets_protocol_handshake_response41

Issue #647, opened 2021-11-16
does go-mysql/client support mysql compress protocol?


zhanghaiyang9999:
does it support mysql compression protocol to connect and transfer data?

dveeden:
This looks like a duplicate of #645

PR #787, opened 2023-05-05
Support MySQL Compressed Protocol

### LINKED ISSUE #645: does replication support mysql compress protocol?
I want to know whether the replication support the mysql compression protocol?
If yes, how to enable the compress protocol in BinlogSyncerConfig? thanks!

dveeden:
Closes #645, #647 

Protocol docs: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_compression.html

This supports both zlib (`CLIENT_COMPRESS`) and zstandard (`CLIENT_ZSTD_COMPRESSION_ALGORITHM`).

This can help with cross-AZ data transfer cost in the cloud and with slower network links.

This probably needs some unittests.

To test:
```go
package main

import (
	"fmt"
	"strings"

)

func run(url, user, password string, flag uint32) {
	println("======================================")
	conn, _ := client.Connect(url, user, password, "test", func(c *client.Conn) {
		c.SetCapability(flag)
	})

	// Trigger a small, uncompresed return packet
	_, err := conn.Execute(`DO 1`)
	if err != nil {
		panic(err)
	}

	for i := 0; i < 2; i++ {
		// With and without a 30 char comment to trigger compressed
		// packets with and without compressed payload
		comment := "/* " + strings.Repeat("x", 30*i) + " */"
		r, err := conn.Execute(`SELECT VERSION(), NOW(6)` + comment)
		if err != nil {
			panic(err)
		}

		for _, row := range r.Values {
			fmt.Printf("%s, %s\n", row[0].AsString(), row[1].AsString())
		}
	}

	// conn.Quit()

	println("======================================")
}

func main() {
	run("127.0.0.1:4000", "root", "", mysql.CLIENT_COMPRESS)
	run("127.0.0.1:3306", "root", "root", mysql.CLIENT_COMPRESS)

	run("127.0.0.1:4000", "root", "", mysql.CLIENT_ZSTD_COMPRESSION_ALGORITHM)
	run("127.0.0.1:3306", "root", "root", mysql.CLIENT_ZSTD_COMPRESSION_ALGORITHM)

	run("127.0.0.1:4000", "root", "", mysql.CLIENT_ZSTD_COMPRESSION_ALGORITHM|mysql.CLIENT_COMPRESS)
	run("127.0.0.1:3306", "root", "root", mysql.CLIENT_ZSTD_COMPRESSION_ALGORITHM|mysql.CLIENT_COMPRESS)
}
```

When inspecting the traffic with Wireshark I would recommend a recent build from the master branch of wireshark.


dveeden:
cc @lance6716

lance6716 (commented):
I'll review later, quite busy these days. Do you have spare time to review @atercattus

packet/conn.go
```diff
@@ -94,8 +106,43 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
...
+			if compressedSequence != c.CompressedSequence {
+				return nil, errors.Errorf("invalid compressed sequence %d != %d",
+					compressedSequence, c.CompressedSequence)
+			}
+
+			if uncompressedLength > 0 {
+				var err error
+				switch c.Compression {
+				case MYSQL_COMPRESS_ZLIB:
+					c.compressedReader, err = zlib.NewReader(c.reader)
```
lance6716, line 126: Can we unify the `c.compressedReader` with `c.reader` by interface?

packet/conn.go
```diff
@@ -209,14 +256,95 @@ func (c *Conn) WritePacket(data []byte) error {
...
+	switch c.Compression {
+	case MYSQL_COMPRESS_NONE:
+		if n, err := c.Write(data); err != nil {
+			return errors.Wrapf(ErrBadConn, "Write failed. err %v", err)
+		} else if n != len(data) {
+			return errors.Wrapf(ErrBadConn, "Write failed. only %v bytes written, while %v expected", n, len(data))
+		}
+	case MYSQL_COMPRESS_ZLIB:
+		fallthrough
+	case MYSQL_COMPRESS_ZSTD:
```
lance6716, line 268: maybe case MYSQL_COMPRESS_ZLIB, MYSQL_COMPRESS_ZSTD

packet/conn.go
```diff
@@ -209,14 +256,95 @@ func (c *Conn) WritePacket(data []byte) error {
...
+		if n, err := c.writeCompressed(data); err != nil {
+			return errors.Wrapf(ErrBadConn, "Write failed. err %v", err)
+		} else if n != len(data) {
+			return errors.Wrapf(ErrBadConn, "Write failed. only %v bytes written, while %v expected", n, len(data))
+		}
+		c.compressedReader = nil
```
lance6716, line 274: Why do we need to set it to nil?

lance6716 (commented):
rest lgtm

packet/conn.go
dveeden: 419796ecc7f9d17cb9beb3002a4f3f23c7c8a342

packet/conn.go
dveeden: On line 138:
```go
        if c.compressedReader != nil {
                if err := c.ReadPacketTo(buf, c.compressedReader); err != nil {
                        return nil, errors.Trace(err)
                }
        } else {
```

I could move it to here (line 122) if that would make things more obvious:
```
                        if uncompressedLength > 0 {
                                var err error
                                switch c.Compression {
                                case MYSQL_COMPRESS_ZLIB:
                                        c.compressedReader, err = zlib.NewReader(c.reader)
                                case MYSQL_COMPRESS_ZSTD:
                                        c.compressedReader = zstd.NewReader(c.reader)
                                }
                                if err != nil {
                                        return nil, err
                                }
                        }
```

The payload of a compressed packet is not always actually compressed. Payloads of less than 50 bytes are often uncompressed as the compression overhead isn't worth it.  If the `uncompressedLength` is 0 then the payload isn't actually compressed.

lance6716 (approved):

PR #789, opened 2023-05-07
tests: select testdb and use MySQL 8.0 by default

dveeden:
- Without the `USE...` in `TestExecuteMultiple` the test panics as the result in `conn_test.go:108` is nil. This doesn't seem to happen consistently. Seems to happen with both 5.7 and 8.0. 
- Use MySQL 8.0 instead of MySQL 5.7
- Add Go 1.20
- Upgrade `actions/setup-go`
- Include OS name in the name of the test

```
$ go test

----------------------------------------------------------------------
PANIC: conn_test.go:92: connTestSuite.TestExecuteMultiple

... Panic: runtime error: invalid memory address or nil pointer dereference (PC=0x438592)

/usr/lib/golang/src/runtime/panic.go:884
  in gopanic
/usr/lib/golang/src/runtime/panic.go:260
  in panicmem
/usr/lib/golang/src/runtime/signal_unix.go:837
  in sigpanic
conn_test.go:108
  in connTestSuite.TestExecuteMultiple.func1
conn.go:273
  in Conn.ExecuteMultiple
conn_test.go:101
  in connTestSuite.TestExecuteMultiple
/usr/lib/golang/src/reflect/value.go:370
  in Value.Call
/home/dvaneeden/go/pkg/mod/github.com/pingcap/check@v0.0.0-20190102082844-67f458068fc8/check.go:836
  in suiteRunner.forkTest.func1
/home/dvaneeden/go/pkg/mod/github.com/pingcap/check@v0.0.0-20190102082844-67f458068fc8/check.go:730
  in suiteRunner.forkCall.func1
/usr/lib/golang/src/runtime/asm_amd64.s:1598
  in goexit

----------------------------------------------------------------------
FAIL: conn_test.go:132: connTestSuite.TestExecuteSelectStreaming

conn_test.go:170:
    c.Assert(err, IsNil)
... value *mysql.MyError = &mysql.MyError{Code:0x416, Message:"No database selected", State:"3D000"} ("ERROR 1046 (3D000): No database selected")

----------------------------------------------------------------------
FAIL: conn_test.go:86: connTestSuite.TestFieldList

conn_test.go:88:
    c.Assert(err, IsNil)
... value *mysql.MyError = &mysql.MyError{Code:0x416, Message:"No database selected", State:"3D000"} ("ERROR 1046 (3D000): No database selected")

----------------------------------------------------------------------
FAIL: conn_test.go:38: connTestSuite.TearDownSuite

conn_test.go:43:
    s.testExecute_DropTable(c)
conn_test.go:83:
    c.Assert(err, IsNil)
... value *mysql.MyError = &mysql.MyError{Code:0x416, Message:"No database selected", State:"3D000"} ("ERROR 1046 (3D000): No database selected")

OOPS: 18 passed, 3 FAILED, 1 PANICKED
--- FAIL: Test (0.46s)
FAIL
exit status 1
FAIL	github.com/go-mysql-org/go-mysql/client	0.465s
```


.github/workflows/ci.yml
```diff
@@ -5,9 +5,9 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ 1.19, 1.18, 1.17, 1.16 ]
+        go: [ "1.20", "1.19", "1.18", "1.17", "1.16" ]
         os: [ ubuntu-22.04, ubuntu-20.04 ]
-    name: Tests Go ${{ matrix.go }} # This name is used in main branch protection rules
+    name: Tests Go ${{ matrix.go }} on ${{ matrix.os }} # This name is used in main branch protection rules
```
atercattus, line 10: We have "Protect matching branches" for the main branch which is based on full name like "Tests Go 1.18".
If we will use a 5 `matrix.go` with a 2 `matrix.os` we need to have 10 protection rules.

I suggest supporting only the last 2 or 3 versions of Go.

lance6716:
@dveeden Hi, would you mind I resolve git conflicts directly in your PR?

dveeden:
> @dveeden Hi, would you mind I resolve git conflicts directly in your PR?

Done

.github/workflows/ci.yml
dveeden: When I created this PR there were 4 Go versions. Now (since 1bd36e617e4313a6a231f0a34a8fc007cbb8aff2) there are only 3 Go versions. So instead of 4x2=8 there are 3x2=6 matrix combinations.

https://endoflife.date/go shows that only 1.20 and 1.19 are supported upstream (the release policy is on https://go.dev/doc/devel/release)

lance6716 (approved):

lance6716:
Thanks! we now keep golang, OS of testing image and MySQL all up-to-date

PR #791, opened 2023-05-08
added query event in dummy event handler

axfor:
added more query event

- [x] onQueryEvent  query event include(create user,drop user,create index event,etd.)

### DIFF
--- a/canal/handler.go
+++ b/canal/handler.go
@@ -19,19 +19,21 @@ type EventHandler interface {
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
 	// OnQueryEvent is query event include(create user,drop user,create index event,etd.)
+	// Note: the OnQueryEvent has lower priority than OnDDL event
 	OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error
 	String() string
 }
 
-type DummyEventHandler struct {
-}
+type DummyEventHandler struct{}
 
 func (h *DummyEventHandler) OnRotate(*replication.EventHeader, *replication.RotateEvent) error {
 	return nil
 }
+
 func (h *DummyEventHandler) OnTableChanged(*replication.EventHeader, string, string) error {
 	return nil
 }
+
 func (h *DummyEventHandler) OnDDL(*replication.EventHeader, mysql.Position, *replication.QueryEvent) error {
 	return nil
 }
@@ -41,6 +43,7 @@ func (h *DummyEventHandler) OnGTID(*replication.EventHeader, mysql.GTIDSet) erro
 func (h *DummyEventHandler) OnPosSynced(*replication.EventHeader, mysql.Position, mysql.GTIDSet, bool) error {
 	return nil
 }
+
 func (h *DummyEventHandler) OnQueryEvent(*replication.BinlogEvent, *replication.QueryEvent, ast.StmtNode, *Position) error {
 	return nil
 }
--- a/canal/handler.go
+++ b/canal/handler.go
@@ -19,8 +19,8 @@ type EventHandler interface {
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
 	// OnQueryEvent is query event include (create user, drop user, create index, etc.)
-	// Note: the OnQueryEvent has lower priority than OnDDL event
-	OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error
+	// Note: the OnQueryEvent has lower priority than OnDDL even
+	OnQueryEvent(header *replication.EventHeader, stmt ast.StmtNode, pos mysql.Position, e *replication.QueryEvent) (bool, bool, error)
 	String() string
 }
 
@@ -43,9 +43,9 @@ func (h *DummyEventHandler) OnGTID(*replication.EventHeader, mysql.GTIDSet) erro
 func (h *DummyEventHandler) OnPosSynced(*replication.EventHeader, mysql.Position, mysql.GTIDSet, bool) error {
 	return nil
 }
-
-func (h *DummyEventHandler) OnQueryEvent(*replication.BinlogEvent, *replication.QueryEvent, ast.StmtNode, *Position) error {
-	return nil
+func (h *DummyEventHandler) OnQueryEvent(*replication.EventHeader, ast.StmtNode, mysql.Position, *replication.QueryEvent) (bool, bool, error) {
+	savePos, force := false, false
+	return savePos, force, nil
 }
 func (h *DummyEventHandler) String() string { return "DummyEventHandler" }
 
--- a/canal/handler.go
+++ b/canal/handler.go
@@ -18,7 +18,7 @@ type EventHandler interface {
 	OnGTID(header *replication.EventHeader, gtid mysql.GTIDSet) error
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
-	// OnQueryEvent is query event include(create user,drop user,create index event,etd.)
+	// OnQueryEvent is query event include (create user, drop user, create index, etc.)
 	// Note: the OnQueryEvent has lower priority than OnDDL event
 	OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error
 	String() string
--- a/canal/handler.go
+++ b/canal/handler.go
@@ -3,6 +3,7 @@ package canal
 import (
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/replication"
+	"github.com/pingcap/tidb/parser/ast"
 )
 
 type EventHandler interface {
@@ -17,6 +18,8 @@ type EventHandler interface {
 	OnGTID(header *replication.EventHeader, gtid mysql.GTIDSet) error
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
+	// OnQueryEvent is query event include(create user,drop user,create index event,etd.)
+	OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error
 	String() string
 }
 
@@ -38,7 +41,9 @@ func (h *DummyEventHandler) OnGTID(*replication.EventHeader, mysql.GTIDSet) erro
 func (h *DummyEventHandler) OnPosSynced(*replication.EventHeader, mysql.Position, mysql.GTIDSet, bool) error {
 	return nil
 }
-
+func (h *DummyEventHandler) OnQueryEvent(*replication.BinlogEvent, *replication.QueryEvent, ast.StmtNode, *Position) error {
+	return nil
+}
 func (h *DummyEventHandler) String() string { return "DummyEventHandler" }
 
 // `SetEventHandler` registers the sync handler, you must register your
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -141,29 +141,22 @@ func (c *Canal) runSyncBinlog() error {
 				c.cfg.Logger.Errorf("parse query(%s) err %v, will skip this event", e.Query, err)
 				continue
 			}
-			posInfo := &Position{
-				Position: pos,
-				SavePos:  savePos,
-				Force:    force,
-			}
 			for _, stmt := range stmts {
 				nodes := parseDDLStmt(stmt)
 				if len(nodes) > 0 {
-					posInfo.SavePos = true
-					posInfo.Force = true
-					err := c.handleDDLEvent(ev, e, nodes, posInfo)
+					savePos = true
+					force = true
+					err := c.handleDDLEvent(ev, e, nodes, pos)
 					if err != nil {
 						c.cfg.Logger.Errorf("handle ddl event err %v", err)
 					}
 				} else {
-					err := c.handleQueryEvent(ev, e, stmt, posInfo)
+					savePos, force, err = c.handleQueryEvent(ev.Header, stmt, pos, e)
 					if err != nil {
 						c.cfg.Logger.Errorf("handle query event err %v", err)
 					}
 				}
 			}
-			savePos = posInfo.SavePos
-			force = posInfo.Force
 			if savePos && e.GSet != nil {
 				c.master.UpdateGTIDSet(e.GSet)
 			}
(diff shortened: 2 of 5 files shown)


canal/handler.go
```diff
@@ -17,6 +17,14 @@ type EventHandler interface {
 	OnGTID(header *replication.EventHeader, gtid mysql.GTIDSet) error
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
+	// OnCreateUser is the create user event in QueryEvent
```
lance6716, line 20: My concern is, the rest methods of this interface are related to binlog event type, not DDL type. So what do you think we add another interface `QueryEventHandler` to handle it? BTW, please add comment to describe the callback order of `OnDDL` and these newly added methods

axfor: Make the canal better meet the synchronization of two heterogeneous or remote data centers, usually the two data centers contain the synchronization of data and user rights

---

I removed `QueryEventHandler ` define, and changed to a more understanding case mode

```go
// handleQueryEvent is handle some common query events (e.g., DDL,CREATE or DROP USER,GRANT),
// others use UnknownQueryEvent unified callbacks to expose to users
func (c *Canal) handleQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent,
	stmt ast.StmtNode, pos mysql.Position, savePos, force *bool) error {
	switch t := stmt.(type) {
	case *ast.RenameTableStmt, *ast.AlterTableStmt, *ast.DropTableStmt, *ast.CreateTableStmt, *ast.TruncateTableStmt:
		return c.handleDDLEvent(ev, e, t, pos, savePos, force)
	case *ast.CreateUserStmt:
		return c.handleCreateUserEvent(ev, e, t, pos, savePos, force)
	case *ast.DropUserStmt:
		return c.handleDropUserEvent(ev, e, t, pos, savePos, force)
	case *ast.GrantStmt:
		return c.handleGrantEvent(ev, e, t, pos, savePos, force)
	default:
		return c.handleUnknownQueryEvent(ev, e, t, pos, savePos, force)
	}
}
```

lance6716: > the rest methods of this interface are related to binlog event type, not DDL type

Please explain how your solution solves this confusing point.

> add comment to describe the callback order of OnDDL and these newly added methods

And this one

axfor:
> the rest methods of this interface are related to binlog event type, not DDL type

sorry, The type `QUERY_EVENT` in the `binlog` log in mysq contains not only `DDL` operations, but also changes to the database,     such as：`Create User`

In our `github.com/go-mysql-org/go-mysql`, only DDL Type is parsed in the Canal package，and now some event parsing has been added

> add comment to describe the callback order of OnDDL and these newly added methods

Sorry, I had a problem with my previous implementation, they have no dependencies, they all belong to the subtypes of `QUERY_EVENT`  and changed to a more understanding case mode

---
```go
func (c *Canal) runSyncBinlog() error {
   //.....
   for {
   //.....
         switch e := ev.Event.(type) {
		//.....
                case *replication.QueryEvent:
			stmts, _, err := c.parser.Parse(string(e.Query), "", "")
			if err != nil {
				c.cfg.Logger.Errorf("parse query(%s) err %v, will skip this event", e.Query, err)
				continue
			}
			for _, stmt := range stmts {
				err := c.handleQueryEvent(ev, e, stmt, pos, &savePos, &force)
				if err != nil {
					c.cfg.Logger.Errorf("handle query event(%s) err %v", e.Query, err)
				}
			}
               //.......
      }
   //.....

// handleQueryEvent is handle some common query events (e.g., DDL,CREATE or DROP USER,GRANT),
// others use UnknownQueryEvent unified callbacks to expose to users
func (c *Canal) handleQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent,
	stmt ast.StmtNode, pos mysql.Position, savePos, force *bool) error {
	switch t := stmt.(type) {
	case *ast.RenameTableStmt, *ast.AlterTableStmt, *ast.DropTableStmt, *ast.CreateTableStmt, *ast.TruncateTableStmt:
		return c.handleDDLEvent(ev, e, t, pos, savePos, force)
	case *ast.CreateUserStmt:
		return c.handleCreateUserEvent(ev, e, t, pos, savePos, force)
	case *ast.DropUserStmt:
		return c.handleDropUserEvent(ev, e, t, pos, savePos, force)
	case *ast.GrantStmt:
		return c.handleGrantEvent(ev, e, t, pos, savePos, force)
	default:
		return c.handleUnknownQueryEvent(ev, e, t, pos, savePos, force)
	}
}
```

canal/handler.go
axfor: > the rest methods of this interface are related to binlog event type, not DDL type

sorry, The type `QUERY_EVENT` in the `binlog` log in mysq contains not only `DDL` operations, but also changes to the database,     such as：`Create User`

In our `github.com/go-mysql-org/go-mysql`, only DDL Type is parsed in the Canal package，and now some event parsing has been added

> add comment to describe the callback order of OnDDL and these newly added methods

Sorry, I had a problem with my previous implementation, they have no dependencies, they all belong to the subtypes of `QUERY_EVENT`  and changed to a more understanding case mode

---
```go
func (c *Canal) runSyncBinlog() error {
   //.....
   for {
   //.....
         switch e := ev.Event.(type) {
		//.....
                case *replication.QueryEvent:
			stmts, _, err := c.parser.Parse(string(e.Query), "", "")
			if err != nil {
				c.cfg.Logger.Errorf("parse query(%s) err %v, will skip this event", e.Query, err)
				continue
			}
			for _, stmt := range stmts {
				err := c.handleQueryEvent(ev, e, stmt, pos, &savePos, &force)
				if err != nil {
					c.cfg.Logger.Errorf("handle query event(%s) err %v", e.Query, err)
				}
			}
               //.......
      }
   //.....

// handleQueryEvent is handle some common query events (e.g., DDL,CREATE or DROP USER,GRANT),
// others use UnknownQueryEvent unified callbacks to expose to users
func (c *Canal) handleQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent,
	stmt ast.StmtNode, pos mysql.Position, savePos, force *bool) error {
	switch t := stmt.(type) {
	case *ast.RenameTableStmt, *ast.AlterTableStmt, *ast.DropTableStmt, *ast.CreateTableStmt, *ast.TruncateTableStmt:
		return c.handleDDLEvent(ev, e, t, pos, savePos, force)
	case *ast.CreateUserStmt:
		return c.handleCreateUserEvent(ev, e, t, pos, savePos, force)
	case *ast.DropUserStmt:
		return c.handleDropUserEvent(ev, e, t, pos, savePos, force)
	case *ast.GrantStmt:
		return c.handleGrantEvent(ev, e, t, pos, savePos, force)
	default:
		return c.handleUnknownQueryEvent(ev, e, t, pos, savePos, force)
	}
}
```

lance6716: Oh I understand, thanks for explanation

canal/handler.go
```diff
@@ -17,6 +17,14 @@ type EventHandler interface {
...
+	OnCreateUser(queryEvent *replication.QueryEvent, user *CreateUser) error
+	// OnDropUser is the drop user event in QueryEvent
+	OnDropUser(queryEvent *replication.QueryEvent, user *DropUser) error
+	// OnGrant is the grant event in QueryEvent
+	OnGrant(queryEvent *replication.QueryEvent, grant *Grant) error
+	// OnQueryEvent is query event (Note: Exclude OnCreateUser and OnDropUser and OnGrant)
```
lance6716, line 26: Maybe we should not add this method? Because this `EventHandler` does not try to cover all types of events. It will be more compatible that we only add well-defined callbacks. 

For example, `OnQueryEvent` may handle CreateIndexStmt, and if in future we decide to add a separate method for CreateIndexStmt, the usage of `OnQueryEvent` will be changed and user may meet problem without compile-time warning.

axfor: I have re-adjusted the code and only added basic `queryEvent`. The detailed events inside (such as `CREATE USER`, `CREATE INDEX`, etc.) will be handled by the users themselves. our library provides basic event notification functionality."

canal/sync.go
```diff
@@ -336,3 +330,48 @@ func (c *Canal) CatchMasterPos(timeout time.Duration) error {
...
+	mysql.Position
+	SavePos bool
+	Force   bool
+}
+
+// handleQueryEvent is handle some common query events (e.g., DDL,CREATE or DROP USER,GRANT),
+// others use UnknownQueryEvent unified callbacks to expose to users
+func (c *Canal) handleQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error {
+	switch t := stmt.(type) {
+	case *ast.RenameTableStmt, *ast.AlterTableStmt, *ast.DropTableStmt, *ast.CreateTableStmt, *ast.TruncateTableStmt:
```
lance6716, line 344: I see that these types are copied from `parseStmt`, but the duplication may cause problems in future. For example, someday we want to add another type, but we forget to change both of them.

I slightly prefer we keep the old logic, 

```
	nodes := parseStmt(stmt)
	for ... range nodes {}
	if len(nodes) == 0 {
		eventHandler.OnQueryEvent()
	} else {
		eventHandler.OnDDL()
	}
```

axfor: # The logic of the `parseSmt` function was also switch case mode
```go
func parseStmt(stmt ast.StmtNode) (ns []*node) {
	switch t := stmt.(type) {
	case *ast.RenameTableStmt:
		for _, tableInfo := range t.TableToTables {
			n := &node{
				db:    tableInfo.OldTable.Schema.String(),
				table: tableInfo.OldTable.Name.String(),
			}
			ns = append(ns, n)
		}
	case *ast.AlterTableStmt:
		n := &node{
			db:    t.Table.Schema.String(),
			table: t.Table.Name.String(),
		}
		ns = []*node{n}
	case *ast.DropTableStmt:
		for _, table := range t.Tables {
			n := &node{
				db:    table.Schema.String(),
				table: table.Name.String(),
			}
			ns = append(ns, n)
		}
	case *ast.CreateTableStmt:
		n := &node{
			db:    t.Table.Schema.String(),
			table: t.Table.Name.String(),
		}
		ns = []*node{n}
	case *ast.TruncateTableStmt:
		n := &node{
			db:    t.Table.Schema.String(),
			table: t.Table.Name.String(),
		}
		ns = []*node{n}
	}
	return ns
}
```

lance6716: I say duplication

axfor: Sorry,  The code has been adjusted.

canal/handler.go
```diff
@@ -17,6 +18,8 @@ type EventHandler interface {
 	OnGTID(header *replication.EventHeader, gtid mysql.GTIDSet) error
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
+	// OnQueryEvent is query event include(create user,drop user,create index event,etd.)
```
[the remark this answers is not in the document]
axfor: comment has been added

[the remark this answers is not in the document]
lance6716:
```suggestion
	// OnQueryEvent is query event include (create user, drop user, create index, etc.)
```

canal/handler.go
```diff
@@ -17,18 +18,22 @@ type EventHandler interface {
...
+	// Note: the OnQueryEvent has lower priority than OnDDL event
+	OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos *Position) error
```
lance6716, line 23: I prefer the parameters are (*EventHeader, StmtNode, *Position, *QueryEvent)

BinlogEvent contains QueryEvent, if we provide two ways to do a thing user may be confused.

canal/sync.go
```diff
@@ -336,3 +341,37 @@ func (c *Canal) CatchMasterPos(timeout time.Duration) error {
 
 	return c.WaitUntilPos(pos, timeout)
 }
+
+type Position struct {
```
lance6716, line 345: why do we need it?

axfor: Based on `mysql.Position` adds `SavePos` and `Force` fields

I think queryEvent contains multiple sub-events, pos save logic, and it is up to the end-user event to decide whether to save or not

---
image


``` go

func (h *DummyEventHandler) OnQueryEvent(header *replication.EventHeader, stmt ast.StmtNode, pos *Position, e *replication.QueryEvent) error { 
         //....
        pos.SavePos =true
	return nil
}

```

---
Of course, it can also be handled uniformly in the main loop of canal/sync.go events.


``` go
func (c *Canal) startSyncer() (*replication.BinlogStreamer, error) { 
        //....
       for _, stmt := range stmts {
		nodes := parseDDLStmt(stmt)
		if len(nodes) > 0 {
		      // OnDDL
		} else {
			err := c.handleQueryEvent(ev.Header, stmt, posInfo, e)
			if err != nil {
				c.cfg.Logger.Errorf("handle query event err %v", err)
			} else {
				savePos = true   // for save pos
				force = true   // for save pos
			}
               }
              //....
       } 
      //....
}
```

canal/handler.go
axfor: Well, okay

canal/sync.go
lance6716: > it is up to the end-user event to decide whether to save or not

I think it's more obvious to use return value to do it.

```
OnQueryEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, stmt ast.StmtNode, pos mysql.Position) (savePos bool, force bool, err error)
```

axfor: Ok, I'll modify it

canal/handler.go
```diff
@@ -17,18 +18,22 @@ type EventHandler interface {
 	OnGTID(header *replication.EventHeader, gtid mysql.GTIDSet) error
 	// OnPosSynced Use your own way to sync position. When force is true, sync position immediately.
 	OnPosSynced(header *replication.EventHeader, pos mysql.Position, set mysql.GTIDSet, force bool) error
+	// OnQueryEvent is query event include (create user, drop user, create index, etc.)
+	// Note: the OnQueryEvent has lower priority than OnDDL even
+	OnQueryEvent(header *replication.EventHeader, stmt ast.StmtNode, pos mysql.Position, e *replication.QueryEvent) (bool, bool, error)
```
lance6716, line 23: Please use named return in this interface, or explain what's the meaning of two `bool`s in comments. So users of this library will understand when they implement this interface.

canal/sync.go
```diff
@@ -336,3 +338,39 @@ func (c *Canal) CatchMasterPos(timeout time.Duration) error {
...
+func (c *Canal) handleDDLEvent(ev *replication.BinlogEvent, e *replication.QueryEvent, nodes []*node, pos *Position) error {
+	for _, node := range nodes {
+		if node.db == "" {
+			node.db = string(e.Schema)
+		}
+		if err := c.updateTable(ev.Header, node.db, node.table); err != nil {
+			return errors.Trace(err)
+		}
+	}
+	if len(nodes) > 0 {
```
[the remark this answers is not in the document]
lance6716: This is always true so no need to check it?

canal/sync.go
```diff
@@ -336,3 +334,32 @@ func (c *Canal) CatchMasterPos(timeout time.Duration) error {
...
+}
+
+// handleQueryEvent is handle some common query events (e.g., DDL,CREATE or DROP USER,GRANT)
+// DDL event use handleDDLEvent, others use the handleQueryEvent
+func (c *Canal) handleQueryEvent(header *replication.EventHeader, stmt ast.StmtNode, pos mysql.Position, e *replication.QueryEvent) (bool, bool, error) {
+	savePos, force, err := c.eventHandler.OnQueryEvent(header, stmt, pos, e)
+	if err != nil {
+		return savePos, force, errors.Trace(err)
+	}
+	return savePos, force, nil
```
lance6716, line 364:
```suggestion
	return savePos, force, errors.Trace(err)
```

canal/handler.go
lwxyzove: i think it also contains the begin of a transcation with the word BEGIN, will a EventHandler.OnTransBegin here?

dveeden:

Issue #713, opened 2022-07-12
当添加无符号列后， 提示数组越界


892294101:
panic: runtime error: index out of range [2] with length 2

goroutine 1 [running]:
myGithubLib/dds/ddscanal.(*RowsEvent).handleUnsigned(0xc0002a2280)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/rows.go:56 +0x5c5
myGithubLib/dds/ddscanal.newRowsEvent(0xc0002c8300, 0x5e4bc9, 0x6, 0xc0002385a0, 0x1, 0x2, 0xc000241080, 0x0)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/rows.go:39 +0xa8
myGithubLib/dds/ddscanal.(*Canal).handleRowsEvent(0xc0000c4000, 0xc000238600, 0x10, 0xc000025df1)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/sync.go:339 +0x15e
myGithubLib/dds/ddscanal.(*Canal).runSyncBinlog(0xc0000c4000, 0x0, 0x0)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/sync.go:116 +0x16a7
myGithubLib/dds/ddscanal.(*Canal).run(0xc0000c4000, 0x0, 0x0)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/canal.go:229 +0xe5
myGithubLib/dds/ddscanal.(*Canal).Run(...)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/canal.go:184
myGithubLib/dds/ddscanal.(*Canal).RunFrom(0xc0000c4000, 0xc0000197a0, 0x10, 0x25d3e, 0x1, 0xc0000197a0)
	D:/workspace/gowork/src/myGithubLib/dds/ddscanal/canal.go:191 +0x65
main.main()
	D:/workspace/gowork/src/myGithubLib/dds/test/mysqltest.go:233 +0x3c5

892294101:
root@localhost [admin]> insert into testunint values(11,11,11);
Query OK, 1 row affected (0.03 sec)

root@localhost [admin]> insert into testunint(name,id) values(22,22);
Query OK, 1 row affected (0.03 sec)

root@localhost [admin]> ALTER TABLE `admin`.`testunint`  ADD COLUMN `b` int(10) unsigned NULL AFTER `a`;
[...]

PR #792, opened 2023-05-18
fix: UnsignedColumns idx out range

guihouchang:
fix:#713

### DIFF
--- a/canal/rows.go
+++ b/canal/rows.go
@@ -53,6 +53,13 @@ func (r *RowsEvent) handleUnsigned() {
 
 	for i := 0; i < len(r.Rows); i++ {
 		for _, columnIdx := range r.Table.UnsignedColumns {
+			if columnIdx >= len(r.Rows[i]) {
+				if columnIdx-1 < len(r.Rows[i]) {
+					columnIdx = columnIdx - 1
+				} else {
+					continue
+				}
+			}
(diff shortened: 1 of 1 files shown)


atercattus:
Hello.
Can you add a test case ([based on this](https://github.com/go-mysql-org/go-mysql/issues/713#issuecomment-1182044043)) that will panic?

atercattus:
Same fix is [here](https://github.com/go-mysql-org/go-mysql/pull/689)

guihouchang:
> Hello. Can you add a test case ([based on this](https://github.com/go-mysql-org/go-mysql/issues/713#issuecomment-1182044043)) that will panic?

I don't know how to add this unit test.It looks very complicated. I need help

lance6716:
> > Hello. Can you add a test case ([based on this](https://github.com/go-mysql-org/go-mysql/issues/713#issuecomment-1182044043)) that will panic?
> 
> I don't know how to add this unit test. It looks very complicated. I need help

Hi, do you find problems when try to use the test framework, or have no idea how to test it?

lance6716:
> have no idea how to test it, plz show me a example

In fact I don't use canal, do you have a real use case that will meet this problem? We can mimic that API invocation and binlog content.

lance6716:
replaced by https://github.com/go-mysql-org/go-mysql/pull/856

Issue #794, opened 2023-06-01
How to close client pool ?


RainJoe:
```go
import (
    "github.com/go-mysql-org/go-mysql/client"
)

pool := client.NewPool(log.Debugf, 100, 400, 5, "127.0.0.1:3306", `root`, ``, `test`)
// ...
conn, _ := pool.GetConn(ctx)
defer pool.PutConn(conn)

conn.Execute() / conn.Begin() / etc...
```
In some usecases, we want to close the pool and also close the underlying mysql connections to avoid resource leaks, but I didn't find a method like pool.Close to implement it.

lance6716:
yes currently we don't have this feature. You can open a PR to implement it

PR #797, opened 2023-06-15
    client: add close method for client pool

RainJoe:
PR for this issue #794


client/pool.go
```diff
@@ -228,6 +233,9 @@ func (pool *Pool) newConnectionProducer() {
 	var err error
 
 	for {
+		if atomic.LoadUint32(&pool.closed) == 1 {
```
lance6716, line 236: in addition to this, we should wait `newConnectionProducer` goroutine exit in `Close` otherwise it may open new connection after we finished closing idle connection and release the lock.

lance6716 (commented):
BTW, can you help us fix the CI 😄

client/pool.go
```diff
@@ -143,6 +145,9 @@ func (pool *Pool) GetStats(stats *ConnectionStats) {
 // GetConn returns connection from the pool or create new
 func (pool *Pool) GetConn(ctx context.Context) (*Conn, error) {
 	for {
+		if atomic.LoadUint32(&pool.closed) == 1 {
```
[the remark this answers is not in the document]
RainJoe: In normal condition, we won't call Close concurrent with GetConn. I add this line is in order to tell caller that pool was closed when someone calls GetConn after Close.

client/pool.go
RainJoe: Yes, I wiil fix this soon.

client/pool.go
```diff
@@ -224,10 +231,14 @@ func (pool *Pool) putConnectionUnsafe(connection Connection) {
 }
 
 func (pool *Pool) newConnectionProducer() {
```
lance6716, line 233: I found another problem, please help me check it.
in line 271, `pool.readyConnection <- connection` may block the for loop. So if `Close` tries to drain `pool.readyConnection` first and then `newConnectionProducer` sends at channel, it will become deadlock.

Generally, I think a `<-pool.ctx.Done()` branch with `pool.readyConnection <- connection` will help to avoid blocking.  And `pool.ctx.Err() != nil` can replace `pool.closed`.

RainJoe: Replacing pool.closed with ctx.Err is a good idea, I wiil update with it. Thanks.

client/pool.go
```diff
@@ -298,6 +312,9 @@ func (pool *Pool) closeOldIdleConnections() {
 	ticker := time.NewTicker(5 * time.Second)
 
 	for range ticker.C {
+		if pool.ctx.Err() != nil {
```
lance6716, line 315: not a big problem. If you change it to

select {
case <- ctx: ...
case <- ticker: ...
}

it will not require 5 seconds to exit at worst.

And if this goroutine is also tracked by pool.wg, we won't have temporary goroutine leak after pool is closed. Some unit test tools may complain about it, like https://github.com/uber-go/goleak

lance6716 (approved):
rest lgtm

I'll merge it after you reply to the comment

client/pool.go
RainJoe: @lance6716, PTAL, it's updated according to your suggestion.

PR #798, opened 2023-06-17
Support LOCAL INFILE request

masahide:
Make `LOAD DATA LOCAL INFILE` available.

### DIFF
--- a/client/auth.go
+++ b/client/auth.go
@@ -202,7 +202,8 @@ func (c *Conn) writeAuthHandshake() error {
 	capability |= c.ccaps&CLIENT_FOUND_ROWS | c.ccaps&CLIENT_IGNORE_SPACE |
 		c.ccaps&CLIENT_MULTI_STATEMENTS | c.ccaps&CLIENT_MULTI_RESULTS |
 		c.ccaps&CLIENT_PS_MULTI_RESULTS | c.ccaps&CLIENT_CONNECT_ATTRS |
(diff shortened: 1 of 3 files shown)


lance6716 (commented):
Even if we turn on the capability flag, in fact we don't support LOAD DATA INFILE?

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response_local_infile_request.html

It wil return a 0xFB packet and searching for `LocalInFile_HEADER` you can see we simply return `ErrMalformPacket`

masahide:
I apologize for the confusion, you're absolutely right. I used go-mysql/client to create a MySQL proxy. 
When using the LOCAL INFILE in this proxy, it is necessary to configure the CLIENT_LOCAL_FILES capability when connecting from go-mysql/client to the MySQL server, and this modification was required for that. This proxy operates by using go-mysql/client only for MySQL handshake, and after connection, the client's net.Conn is directly transferred to the server-side net.Conn, hence making the LOCAL INFILE operational with just the LocalInFile_HEADER.
https://github.com/masahide/mysql8-audit-proxy/blob/main/pkg/mysqlproxy/clntsess.go#L33-L38

Project top page: https://github.com/masahide/mysql8-audit-proxy

lance6716 (approved):
lgtm. Can you help fix the linter? If you don't have enough time, I'll make a commit tomorrow

lance6716:
btw, what's your golang version? in https://github.com/go-mysql-org/go-mysql/issues/799 we are considering to drop support for older golang versions

masahide:
I am using go version 1.20.

Issue #799, opened 2023-06-17
Is it possible to use github.com/klauspost/compress instead of github.com/DataDog/zstd


hjweddie:
github.com/DataDog/zstd is not compatible with arm64 platform and it relies on cgo and linux heavily.
We use arm64 graviton3 instance on aws and arm instances on other cloud provider heavily to reduce cost. We find that it could not compile when we upgrade github.com/go-mysql-org/go-mysql to lastest commit on master branch.

lance6716:
it's because github.com/klauspost/compress does not support older golang versions. may I ask what's your golang version in production?

and maybe we can provide two implementations which is controlled by golang version. for old versions we can use github.com/DataDog/zstd which requires cgo.

hjweddie:
we use go 1.20 and go1.19 in production.
Regards to golang release policy, we prefer to upgrade to new version. So that we suggest that go-mysql may support new golang version.

refers:
[Release Policy](https://go.dev/doc/devel/release#Release%20Policy)
Each major Go release is supported until there are two newer major releases

lance6716:
oh that's reasonable! what do you think @atercattus @dveeden

hjweddie:
It may be of great help for us, if it can be controlled by golang version build tags

dveeden:
I think we can remove support for some older Go versions that don't have upstream support anymore and have a single implementation that works on all platforms.

PR #800, opened 2023-06-18
use github.com/klauspost/compress/zstd instead of github.com/DataDog/…

hjweddie:
related to issue https://github.com/go-mysql-org/go-mysql/issues/799


dveeden:
This is basically the same as #801 but I hadn't noticed this PR before I started.

My PR removes older Go versions and add some tests.

hjweddie:
thanks

PR #803, opened 2023-06-22
change unit test framework to https://github.com/stretchr/testify

lance6716:


lance6716:
Maybe the failed test is because in old test framework all Suite are run sequentially, but in testify they are run concurrently. And we have a RESET MASTER in `(*testSyncerSuite).TestMysqlBinlogCodec()`

I'll think how to bypass it. Do you have some advice? @atercattus

lance6716:
> Maybe the failed test is because in old test framework all Suite are run sequentially, but in testify they are run concurrently. And we have a RESET MASTER in `(*testSyncerSuite).TestMysqlBinlogCodec()`
> 
> I'll think how to bypass it. Do you have some advice? @atercattus

Seems it's fixed https://github.com/go-mysql-org/go-mysql/actions/runs/5368879516
image

ptal @atercattus

lance6716:
The PR is very big and hard to review, and now the tests can stably pass. I'll merge it.

PR #804, opened 2023-06-23
Update sync.go

wuwenbao:
#713 
Skip if the number of fields is different

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -251,6 +251,10 @@ func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


lance6716:
Can you describe how does this problem occur, and what's your expected behaviour in issue #713? I'm not sure what's the correct fix for it, and I think we'd better discuss about it first. For example, if we silently ignore the problem as your PR, it will cause replication of the table stopped without a warning.

dveeden:
@wuwenbao could you check the question from @lance6716 ?

dveeden:
@wuwenbao Could you check the question?

dveeden:

PR #812, opened 2023-08-08
add throttle to prevent cpu load

alkaaf:
i hope this is usable. i need this when resume from certain bin log position

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -122,6 +122,11 @@ type BinlogSyncerConfig struct {
 	RowsEventDecodeFunc func(*RowsEvent, []byte) error
 
 	DiscardGTIDSet bool
+
+	// do throttle when the event channel is more than this value. 0 means no throttle
+	ThrottleCap int
+	// delay read for this duration when throttle is triggered
+	ThrottleDuration time.Duration
 }
 
 // BinlogSyncer syncs binlog event from server.
@@ -852,6 +857,11 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
 	}
 
 	needStop := false
+	// throttle the event, prevent cpu flood
+	if b.cfg.ThrottleCap > 0 && len(s.ch) >= b.cfg.ThrottleCap {
+		b.cfg.Logger.Infof(`throttling binlog read`)
+		time.Sleep(b.cfg.ThrottleDuration)
+	}
 	select {
 	case s.ch <- e:
 	case <-b.ctx.Done():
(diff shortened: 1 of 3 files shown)


replication/binlogsyncer.go
```diff
@@ -852,6 +857,11 @@ func (b *BinlogSyncer) parseEvent(s *BinlogStreamer, data []byte) error {
 	}
 
 	needStop := false
+	// throttle the event, prevent cpu flood
+	if b.cfg.ThrottleCap > 0 && len(s.ch) >= b.cfg.ThrottleCap {
+		b.cfg.Logger.Infof(`throttling binlog read`)
+		time.Sleep(b.cfg.ThrottleDuration)
```
lance6716, line 863: should use 
```golang
select {
case <-time.After...
case <-b.ctx.Done...
}
```
to find context cancel timely

lance6716 (commented):
And can you explain your problem first? I guess you are using this library and other processes in the same machine, and this library comsumes too much CPU. But a more common way is to use cgroup or something to limit the resource of the process using this library.

alkaaf:
I have experience using maxwell for etl from mysql to mongodb where reading result from maxwell i store in redis queue. there are some workers consume queue to store it in mongodb

This approach has a lot of overhead from the network side and also expensive (redis, worker, maxwell that using java). therefore I need a cheaper, concise, reliable and consistent approach where delay is not an issue.

the ability to continue from the last log position was something I was looking for from this library. the problem is that when this app fails/stops some time and when restarting the app, the channel gets flooded with events starting from a certain position. it causes event processing to slow down to even stall (yeah you are right that i do event processing too on the same machine even application). this is why limiting resources is not an option.

therefore I need something to delay the binlog read by a few milliseconds when the channel reaches a certain size. (in my case, when the channel size is 512 ,the read will be delayed by 1 second)

lance6716:
I'm still not very clear about your usage, but I'd ask do you think it's a good idea to add an option to control the channel size? So when channel is full, sending to channel will be blocked and no further CPU / network resources will be consumed.

https://github.com/go-mysql-org/go-mysql/blob/dc97dfaa93aed1671e2733e335aa0e4e5bc423a0/replication/binlogstreamer.go#L87-L90

Maybe you can set the channel size to 512, and the delay will not as bad as 1 second.

lance6716:
Hi, can you check https://github.com/go-mysql-org/go-mysql/pull/830 is helpful to your problem?

dveeden:
@alkaaf is this still a problem for you? if so, could you fix the conflict?

dveeden:

PR #813, opened 2023-08-15
Add VisibilityBitmap to TableMapEvent in replication

dongwook-chan:
I was implementing optional_metadata field in table_map_event for [python-mysql-replication](https://github.com/julien-duponchelle/python-mysql-replication).
And came across this repo and found that metadata type COLUMN_VISIBILITY isn't implemented.

### Changes
- Added `TableMapEvent.VisibilityBitmap`
  - `VisibilityBitmap` is a bitmap where each bit represents the visibility of a corresponding column in a table.
  - If a bit is set, it indicates that the corresponding column is NOT an invinsible column.
  - [Invisible column](https://dev.mysql.com/doc/refman/8.0/en/invisible-columns.html) was introduced in MySQL version 8.0.23.
- `TableMapEvent.VisibilityMap`
  - `VisibilityMap` lists boolean values of which is true if column of the same index is NOT an invisible column.

### Examples
You can check the visibility of a column by its table by executing `DESCRIBE` command.
```sql
mysql> SELECT VERSION(); -- must be MySQL 8.0.23 or higher. Not supported by MariaDB yet.
+-----------+
| VERSION() |
+-----------+
| 8.0.33-25 |
+-----------+
1 row in set (0.00 sec)

mysql> CREATE TABLE test_visibility (c1 INT, c2 INT INVISIBLE, c3 INT);
Query OK, 0 rows affected (0.03 sec)

mysql> DESC test_visibility;
+-------+------+------+-----+---------+-----------+
| Field | Type | Null | Key | Default | Extra     |
+-------+------+------+-----+---------+-----------+
| c1    | int  | YES  |     | NULL    |           |
| c2    | int  | YES  |     | NULL    | INVISIBLE |
| c3    | int  | YES  |     | NULL    |           |
+-------+------+------+-----+---------+-----------+
3 rows in set (0.00 sec)
```


dongwook-chan:
@lance6716
I appreciate your positive reaction to my PR, and I'm grateful for your contributions to this repository. `go-mysql`  is indispensable to me. Could you please advise whether I should include tests for this new addition?

dongwook-chan:
I added a co-author.

replication/row_event.go
```diff
@@ -730,6 +740,22 @@ func (e *TableMapEvent) GeometryTypeMap() map[int]uint64 {
...
+// Invisible column was introduced in MySQL 8.0.23
+// nil is returned if not available.
+func (e *TableMapEvent) VisibilityMap() map[int]bool {
+	if len(e.VisibilityBitmap) == 0 {
+		return nil
+	}
+	p := 0
+	ret := make(map[int]bool)
+	for i := 0; i < int(e.ColumnCount); i++ {
+		ret[i] = e.VisibilityBitmap[p/8]&(1<<uint(7-p%8)) != 0
```
atercattus, line 753: nit: Maybe it's a little easier to understand the [original code](https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/libbinlogevents/src/rows_event.cpp#L316) with countdown instead of `1<<uint(7-p%8)`

dongwook-chan: @atercattus 
I appreciate your review. Also, I agree with you and I can get this fixed.
Actually [`VisibilityMap`](https://github.com/dongwook-chan/go-mysql/blob/ccabd266aaa188de9046ee782c144fec36fc7dab/replication/row_event.go#L743-L757) is copy and paste of existing [`UnsignedMap()`](https://github.com/dongwook-chan/go-mysql/blob/ccabd266aaa188de9046ee782c144fec36fc7dab/replication/row_event.go#L614-L631). Should I fix [`UnsignedMap()`](https://github.com/dongwook-chan/go-mysql/blob/ccabd266aaa188de9046ee782c144fec36fc7dab/replication/row_event.go#L614-L631) too?

atercattus: Oh. @lance6716 what do you think about this? Apply the Boy Scout Rule or not?

dongwook-chan: If we decide to improve, the code will look like below.
```go
func (e *TableMapEvent) VisibilityMap() map[int]bool {
	if len(e.VisibilityBitmap) == 0 {
		return nil
	}
	ret := make(map[int]bool)
	for _, field := range e.VisibilityBitmap {
		for c, i := 0x80, 0; c != 0; c, i = c>>1, i+1 {
			ret[i] = field&byte(c) != 0
		}
	}
	return ret
}
```

lance6716: The old code is really not easy to understand 😢 I also hope we start to improve the readability. As for UnsignedMap, I'm OK that you also modify it in this PR or another PR. WDYT @atercattus 

Your new code snippet seems not optimal. The status of `i` should be kept when iterate bytes of `VisibilityBitmap`. And we can early stop when the column count is not a multiple of 8 (though the [original code](https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/libbinlogevents/src/rows_event.cpp#L316) seems not optimize about this). Maybe we can

```
...
i := 0
for e.VisibilityBitmap {
  for c := 0x80; ... {
    ret[i] = ...
    i++
    if i >= e.ColumnCount { return }
  }
}
...
```
Welcome to find different code

replication/row_event.go
```diff
@@ -730,6 +740,22 @@ func (e *TableMapEvent) GeometryTypeMap() map[int]uint64 {
 	return ret
 }
 
+// VisibilityMap returns a map: column index -> visiblity.
+// Invisible column was introduced in MySQL 8.0.23
+// nil is returned if not available.
+func (e *TableMapEvent) VisibilityMap() map[int]bool {
```
lance6716, line 746: As for testing, I think maybe you can capture/construct some `VisibilityBitmap` bytes and test the result of TableMapEvent.VisibilityMap is expected.

replication/row_event.go
dongwook-chan: My bad. `i` should indeed be declared outside of the inner loop.

replication/row_event.go
dongwook-chan: @lance6716 
Sure. I think I should create a separate function. Just as `TestTableMapOptMetaPrimaryKey` is defined solely for primary key type meta data. At first, I thought about squeezing into `TestTableMapOptMetaNames` or `TestTableMapHelperMaps`, but it's too packed and I saw no room for any addition. Your advice would be help me a lot. 😀

dongwook-chan:
I added test case for 8.0 only, because the keyword `INVISIBLE` to create invisible column is not available in any other vendors or versions.

lance6716 (approved):

lance6716:
ptal @atercattus

dongwook-chan:
But on a second thought, I would like to add case 2 where invisible columns do not exist at all! This case will be applicable to MySQL 5.7, MariaDB 10.4 and MairaDB 10.5.

lance6716:
lgtm now. let's wait @atercattus

lance6716:
@dongwook-chan seems atercattus is a bit busy. Do you want me to merge it?

atercattus (approved):

dongwook-chan:
@lance6716 No need to wait any more!
@atercattus Thank you for the approval!

PR #817, opened 2023-08-31
Add function to parse 'extradata’ in rows event

cucuridas:
Hello, I am a developer contributing to py-mysql-replication. 

While checking the source code, we developed it because the parsing-related part of 'rows event' to 'extra data' was not developed in the go-mysql package
(This feature has been [added](https://github.com/mysql/mysql-server/commit/e11a540fc559dfbed72ba4ec4677638ac917454f) since Mysql 8.0.16)

Please leave a comment if there are any modifications or additions to the code!

## D**eveloped**

- Added function 'decodeExtraData' that in file 'replication.row_event.go’
- The function is executed when the DecodeHeader function is executed and the dataLen value is greater than 2
- The extra data to be parsed is as follows
    
    ```bash
    				+----------+--------------------------------------+
            |type_code |        extra_row_ndb_info            |
            +--- ------+--------------------------------------+
            | NDB      |Len of ndb_info |Format |ndb_data     |
            | 1 byte   |1 byte          |1 byte |len - 2 byte |
            +----------+----------------+-------+-------------+
    
            In case of INSERT/DELETE
            +-----------+----------------+
            | type_code | partition_info |
            +-----------+----------------+
            |   PART    |  partition_id  |
            | (1 byte)  |     2 byte     |
            +-----------+----------------+
    
            In case of UPDATE
            +-----------+------------------------------------+
            | type_code |        partition_info              |
            +-----------+--------------+---------------------+
            |   PART    | partition_id | source_partition_id |
            | (1 byte)  |    2 byte    |       2 byte        |
            +-----------+--------------+---------------------+
    ```


replication/row_event.go
```diff
@@ -878,7 +878,13 @@ type RowsEvent struct {
 	Flags uint16
 
 	// if version == 2
-	ExtraData []byte
+	// Use when DataLen value is greater than 2
+	NdbLength uint8
```
lance6716, line 882: seems it's complaining that the type of variables are not vertically aligned.

lance6716 (commented):
Hi, can you mimic https://github.com/go-mysql-org/go-mysql/blob/cfe601218a91b198994d7aad75ef09bcd96dbd2b/replication/row_event_test.go#L942 to dump some binlog content and test the new fields work as expected?

replication/row_event.go
```diff
@@ -955,8 +960,12 @@ func (e *RowsEvent) DecodeHeader(data []byte) (int, error) {
 	if e.Version == 2 {
 		dataLen := binary.LittleEndian.Uint16(data[pos:])
 		pos += 2
-
-		e.ExtraData = data[pos : pos+int(dataLen-2)]
+		if dataLen > 2 {
+			err := e.decodeExtraData(pos, data)
```
lance6716, line 964: (not important, just a code style) this project more often changes the start of slice at caller, rather than passes the index to function. In other words, I prefer `e.decodeExtraData(data[pos:])`

replication/row_event.go
```diff
@@ -985,6 +994,28 @@ func (e *RowsEvent) DecodeHeader(data []byte) (int, error) {
 	return pos, nil
 }
 
+func (e *RowsEvent) decodeExtraData(pos int, data []byte) (err2 error) {
+	extraDataType := data[pos]
+	pos += 1
+	switch extraDataType {
+	case 0:
```
lance6716, line 1001: it's better to add a const for `0`, like ENUM_EXTRA_ROW_INFO_TYPECODE_NDB or something. Same for below `1` case.

replication/row_event.go
```diff
@@ -985,6 +994,28 @@ func (e *RowsEvent) DecodeHeader(data []byte) (int, error) {
...
+		var ndbLength int = int(data[pos])
+		pos += 1
+		e.NdbFormat = data[pos]
+		pos += 1
+		e.NdbData = data[pos : pos+ndbLength-2]
+	case 1:
+		if e.eventType == UPDATE_ROWS_EVENTv2 {
```
lance6716, line 1008: MySQL checked more event types

https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/libbinlogevents/src/rows_event.cpp#L443-L454

lance6716 (commented):
rest lgtm

replication/row_event.go
```diff
@@ -985,6 +994,29 @@ func (e *RowsEvent) DecodeHeader(data []byte) (int, error) {
...
+	pos += 1
+	switch extraDataType {
+	case ENUM_EXTRA_ROW_INFO_TYPECODE_NDB:
+		var ndbLength int = int(data[pos])
+		pos += 1
+		e.NdbFormat = data[pos]
+		pos += 1
+		e.NdbData = data[pos : pos+ndbLength-2]
+	case ENUM_EXTRA_ROW_INFO_TYPECODE_PARTITION:
+		if e.eventType == UPDATE_ROWS_EVENTv2 || e.eventType == PARTIAL_UPDATE_ROWS_EVENT {
```
lance6716, line 1009: I'm still not sure about it. MySQL enables it for `UPDATE_ROWS_EVENT`, `UPDATE_ROWS_EVENT_V1` and `PARTIAL_UPDATE_ROWS_EVENT`, and we do it for `UPDATE_ROWS_EVENTv2` and `PARTIAL_UPDATE_ROWS_EVENT`. I remember `UPDATE_ROWS_EVENT` and `UPDATE_ROWS_EVENT_V1` has been deprecated so we really don't need to care about them. I'll check it later.

And it's helpful that you can also help me check and provide some reference 😄

lance6716: As you says in PR description, this feature is added in v8.0.16, and [log_bin_use_v1_row_events](https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_log_bin_use_v1_row_events) is deprecated in v8.0.18, so we should also consider UPDATE_ROWS_EVENT_V1.

And MySQL's UPDATE_ROWS_EVENT (`31`) is our UPDATE_ROWS_EVENTv2

lance6716 (approved):

PR #818, opened 2023-09-01
Add GTIDEvent interface to get the SequenceNumber and LastCommitted

cameron-p-m:
## Problem

We've implemented MySQL Parallel Replication on our application that uses a fork of this library, and would like to merge the changes upstream.
Here are some details explaining it: https://github.com/isotopp/mysql-dev-examples/blob/master/mysql-binlog-parallel/parallel.md 

We need to get the SequenceNumber and LastCommitted in order to do so. This is what I thought was the best way. Will have to be part of the next major release because it changes the interface of the handler.


replication/event.go
```diff
@@ -356,7 +356,7 @@ func (e *QueryEvent) Dump(w io.Writer) {
 	fmt.Fprintln(w)
 }
 
-type GTIDEvent struct {
+type MySQLGTIDEvent struct {
```
lance6716, line 359: I guess you rename this struct to avoid using the same name with the new interface `GTIDEvent`? But it will break more compatibility. I'll think if we have a better way to handle it later

mysql/gtid.go
```diff
@@ -29,3 +29,8 @@ func ParseGTIDSet(flavor string, s string) (GTIDSet, error) {
 		return nil, errors.Errorf("invalid flavor %s", flavor)
 	}
 }
+
+type GTIDEvent interface {
+	GTIDNext() (GTIDSet, error)
+	Decode(data []byte) error
```
lance6716, line 35: How do we use `Decode`? I see in the two implementations of `GTIDEvent` `Decode` changes the receiver, but `GTIDEvent` is an interface we know nothing about the actual receiver without type cast. And if we use type cast, we can directly call `MySQLGTIDEvent.Decode`, no need to add it to the interface.

lance6716:
> We need to get the SequenceNumber and LastCommitted in order to do so

How about we expose `SequenceNumber` and `LastCommitted` as the common methods of  `GTIDEvent` interface? Seems MariaDB does not have `LastCommitted` so we can use `SequenceNumber-1` (I will learn it later)

And also an `ID() string` method which returns server UUID / domain ID to distinguish binlog events from different primary node that can be applied concurrently.

replication/event.go
cameron-p-m: I could call it GenericGTIDEvent or something?

Wouldn't be what we would probably name it originally, but would break less compatibility.

cameron-p-m: I went with BinlogGTIDEvent. Let me know what you think.

lance6716: lgtm for the name. I'm a bit busy these days, will check the whole PR later

cameron-p-m:
>How about we expose SequenceNumber and LastCommitted as the common methods of GTIDEvent interface? Seems MariaDB does not have LastCommitted so we can use SequenceNumber-1 (I will learn it later)

I did this, but it doesn't really work because of MariaDBs unsigned int64 vs MySQL signed int64. The changes are pushed so you can take a look.

lance6716:
seems `sequence_number` can be a negative value 😢 

https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/sql/rpl_trx_tracking.cc#L151-L163

I'll check MariaDB later. Maybe unify them is not a good idea.

cameron-p-m:
@lance6716 I updated the code and added some tests for the GTID event.

lance6716:
> but it doesn't really work because of MariaDBs unsigned int64 vs MySQL signed int64

Sorry I didn't realised it before I recommend to put the `SequenceNumber` and `LastCommitted` into the interface. So maybe 
```
type BinlogGTIDEvent interface {
	GTIDNext() (GTIDSet, error)
}
```
is the final version.

And IMO the `BinlogGTIDEvent` seems a bit unnatural because it only has one method 😂 Do you think it's better that the logic of "checking GTID dependency" can be encapsulated in this interface? like 
```
type BinlogGTIDEvent interface {
...
	DependOn(previous BinlogGTIDEvent) bool
}
```
But in opposition to it, this library should only care about parsing and leave the dependency calculation to the upper application layer.

Do you have any ideas?

lance6716 (approved):

cameron-p-m:
> And IMO the BinlogGTIDEvent seems a bit unnatural because it only has one method 😂 Do you think it's better that the logic of "checking GTID dependency" can be encapsulated in this interface? like
> 
> type BinlogGTIDEvent interface {
> ...
> 	DependOn(previous BinlogGTIDEvent) bool
> }
> But in opposition to it, this library should only care about parsing and leave the dependency calculation to the upper application layer.
> 
> Do you have any ideas?

I think the dependency calculation should be up to the application. One issue with this is that I don't think it could work in MariaDB and MySQL. 

The dependency calculation also changes if there is filtering, so I think the application has to decide anyways.

PR #819, opened 2023-09-04
fix: prevent panic on malformed handshake

btoonk:
This PR resolves a panic in decodeFirstPart when the functions receives malformed data.
```
panic: runtime error: slice bounds out of range [:2] with capacity 0
```


lance6716 (approved):

lance6716:
thanks!

Issue #821, opened 2023-09-18
add value string output


zhujintao:
rowdata.go
```
if isNull {
	data[i].Type = FieldValueTypeNull
} else {
	isUnsigned := f[i].Flag&UNSIGNED_FLAG != 0
	data[i].str = v  //add here
```

lance6716:
Hi, can you explain in which use case you meet the problem?

zhujintao:
simple and fast combination of string SQL statements, no need to determine the type. 
```
values := make([]string, 0, len(row))
for _, v := range row {
	if v.Type == mysql.FieldValueTypeNull {
		values = append(values, "NULL")
	} else {
		values = append(values, fmt.Sprintf("'%s'", v.AsString()))
	}
}
// INSERT INTO xx VALUES ('a','b','1','2.0',NULL)
```

lance6716:
I see that not all MySQL types implements `AsString` (for example, numbers uses `FieldValue.value`). So it does not totally solve the problem. And generally concatenating the SQL as string is a bad idea, it has security problems compared to letting SQL driver interpolate the arguments

zhujintao:
can also use like ```String()``` methods to output the raw value ```v```, ```FieldValue.str```

lance6716:
> can also use like `String() Raw()` methods to output the raw value `v`, `FieldValue.str`

LGTM. Do you have time to write a PR for it? 😄

PR #823, opened 2023-09-19
add value string output

zhujintao:
#821

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -213,7 +213,6 @@ func (s *clientTestSuite) TestConn_Select2() {
 			require.Equal(s.T(), v.String(), ss[idx])
 		}
 	}
-
 }
 
 func (s *clientTestSuite) TestConn_Escape() {
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -203,26 +203,6 @@ func (s *clientTestSuite) TestConn_Select() {
 	require.Equal(s.T(), "test1", e)
 }
 
-func (s *clientTestSuite) TestConn_Select2() {
-	str := `insert into test.mixer_test_conn (id, str, f, e, u, i,j) values (6, "a", 3.14, "test1", 255, -127,'{"key1": 1, "key2": null}'),(7, NULL,NULL, NULL, NULL, NULL,NULL)`
-	result, err := s.c.Execute(str)
-	require.NoError(s.T(), err)
-	require.Equal(s.T(), uint64(2), result.AffectedRows)
-
-	str = `select * from mixer_test_conn where id in (6,7)`
-	ss := [][]string{{"6", "a", "3.14", "test1", "255", "-127", "{\"key1\": 1, \"key2\": null}"}, {"7", "", "", "", "", "", ""}}
-	result, err = s.c.Execute(str)
-	require.NoError(s.T(), err)
-
-	require.Len(s.T(), result.Values, 2)
-
-	for i, row := range result.Values {
-		for idx, v := range row {
-			require.Equal(s.T(), v.String(), ss[i][idx])
-		}
-	}
-}
-
 func (s *clientTestSuite) TestConn_Escape() {
 	e := `""''\abc`
 	str := fmt.Sprintf(`insert into mixer_test_conn (id, str) values(5, "%s")`,
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -208,7 +208,6 @@ func (s *clientTestSuite) TestConn_Select2() {
 	ss := []string{"1", "a", "3.14", "test1"}
 	result, err := s.c.Execute(str)
 	require.NoError(s.T(), err)
-
 	for _, row := range result.Values {
 		for idx, v := range row {
 			require.Equal(s.T(), v.String(), ss[idx])
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -203,6 +203,20 @@ func (s *clientTestSuite) TestConn_Select() {
 	require.Equal(s.T(), "test1", e)
 }
 
+func (s *clientTestSuite) TestConn_Select2() {
+	str := `select id, str, f, e from mixer_test_conn where id = 1`
+	ss := []string{"1", "a", "3.14", "test1"}
+	result, err := s.c.Execute(str)
+	require.NoError(s.T(), err)
+
+	for _, row := range result.Values {
+		for idx, v := range row {
+			require.Equal(s.T(), v.String(), ss[idx])
+		}
+	}
+
+}
+
 func (s *clientTestSuite) TestConn_Escape() {
 	e := `""''\abc`
 	str := fmt.Sprintf(`insert into mixer_test_conn (id, str) values(5, "%s")`,
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -204,7 +204,6 @@ func (s *clientTestSuite) TestConn_Select() {
 }
 
 func (s *clientTestSuite) TestConn_Select2() {
-
 	str := `insert into test.mixer_test_conn (id, str, f, e, u, i,j) values (6, "a", 3.14, "test1", 255, -127,'{"key1": 1, "key2": null}'),(7, NULL,NULL, NULL, NULL, NULL,NULL)`
 	result, err := s.c.Execute(str)
 	require.NoError(s.T(), err)
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -204,13 +204,22 @@ func (s *clientTestSuite) TestConn_Select() {
 }
 
 func (s *clientTestSuite) TestConn_Select2() {
-	str := `select id, str, f, e from mixer_test_conn where id = 1`
-	ss := []string{"1", "a", "3.14", "test1"}
+
+	str := `insert into test.mixer_test_conn (id, str, f, e, u, i,j) values (6, "a", 3.14, "test1", 255, -127,'{"key1": 1, "key2": null}'),(7, NULL,NULL, NULL, NULL, NULL,NULL)`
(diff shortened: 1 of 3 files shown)


lance6716:
Hi, as discussed in #821 I supposed we are going to add a method that for all field types it will return string representation. Just double check is it your plan?

zhujintao:
recheck plan，has been modified to ```String()``` return string.

zhujintao:
> Hi, as discussed in #821 I supposed we are going to add a method that for all field types it will return string representation. Just double check is it your plan?


submitted, can it be merged?

mysql/rowdata.go
```diff
@@ -40,6 +40,7 @@ func (p RowData) ParseText(f []*Field, dst []FieldValue) ([]FieldValue, error) {
 			data[i].Type = FieldValueTypeNull
 		} else {
 			isUnsigned := f[i].Flag&UNSIGNED_FLAG != 0
+			data[i].str = append(data[i].str[:0], v...)
```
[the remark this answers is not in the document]
zhujintao: unit test submitted.  and delete  type is default  ```str``` repeated assignment.

client/client_test.go
```diff
@@ -203,6 +203,18 @@ func (s *clientTestSuite) TestConn_Select() {
 	require.Equal(s.T(), "test1", e)
 }
 
+func (s *clientTestSuite) TestConn_Select2() {
```
lance6716, line 206: you can start a MySQL in your local machine and run this test. This test case will use the DSN in line 33 to connect MySQL. In fact the test is invalid because `result.Values` has no element, please check you really test some row. 

And I think the original issue is about NULL? As least you should test NULL, and insert some rows to mixer_test_conn and check all columns

zhujintao: the use cases have been completed and tested locally

```
[root@mysql003 ~]# go  test -v --host 192.168.10.44 
=== RUN   TestConnGenAttributes
--- PASS: TestConnGenAttributes (0.00s)
=== RUN   TestClientSuite
=== RUN   TestClientSuite/TestConn_Escape
=== RUN   TestClientSuite/TestConn_Insert
=== RUN   TestClientSuite/TestConn_Insert2
=== RUN   TestClientSuite/TestConn_Ping
=== RUN   TestClientSuite/TestConn_Select
=== RUN   TestClientSuite/TestConn_Select2
=== RUN   TestClientSuite/TestConn_SetCapability
=== RUN   TestClientSuite/TestConn_SetCharset
=== RUN   TestClientSuite/TestConn_TLS_Certificate
=== RUN   TestClientSuite/TestConn_TLS_Skip_Verify
=== RUN   TestClientSuite/TestConn_TLS_Verify
=== RUN   TestClientSuite/TestStmt_Delete
=== RUN   TestClientSuite/TestStmt_Insert
=== RUN   TestClientSuite/TestStmt_NULL
=== RUN   TestClientSuite/TestStmt_Select
=== RUN   TestClientSuite/TestStmt_Signed
=== RUN   TestClientSuite/TestStmt_Trans
=== RUN   TestClientSuite/TestStmt_Unsigned
--- PASS: TestClientSuite (0.39s)
    --- PASS: TestClientSuite/TestConn_Escape (0.01s)
    --- PASS: TestClientSuite/TestConn_Insert (0.01s)
    --- PASS: TestClientSuite/TestConn_Insert2 (0.00s)
    --- PASS: TestClientSuite/TestConn_Ping (0.00s)
    --- PASS: TestClientSuite/TestConn_Select (0.00s)
    --- PASS: TestClientSuite/TestConn_Select2 (0.01s)
    --- PASS: TestClientSuite/TestConn_SetCapability (0.00s)
    --- PASS: TestClientSuite/TestConn_SetCharset (0.00s)
    --- PASS: TestClientSuite/TestConn_TLS_Certificate (0.00s)
    --- PASS: TestClientSuite/TestConn_TLS_Skip_Verify (0.01s)
    --- PASS: TestClientSuite/TestConn_TLS_Verify (0.00s)
    --- PASS: TestClientSuite/TestStmt_Delete (0.00s)
    --- PASS: TestClientSuite/TestStmt_Insert (0.01s)
    --- PASS: TestClientSuite/TestStmt_NULL (0.01s)
    --- PASS: TestClientSuite/TestStmt_Select (0.00s)
    --- PASS: TestClientSuite/TestStmt_Signed (0.02s)
    --- PASS: TestClientSuite/TestStmt_Trans (0.02s)
    --- PASS: TestClientSuite/TestStmt_Unsigned (0.01s)
=== RUN   TestConnSuite
=== RUN   TestConnSuite/TestAttributes
=== RUN   TestConnSuite/TestExecuteMultiple
=== RUN   TestConnSuite/TestExecuteSelectStreaming
=== RUN   TestConnSuite/TestFieldList
--- PASS: TestConnSuite (0.25s)
    --- PASS: TestConnSuite/TestAttributes (0.00s)
    --- PASS: TestConnSuite/TestExecuteMultiple (0.01s)
    --- PASS: TestConnSuite/TestExecuteSelectStreaming (0.00s)
    --- PASS: TestConnSuite/TestFieldList (0.00s)
=== RUN   TestPoolSuite
=== RUN   TestPoolSuite/TestPool_Close
--- PASS: TestPoolSuite (0.01s)
    --- PASS: TestPoolSuite/TestPool_Close (0.01s)
PASS
ok      github.com/go-mysql-org/go-mysql/client 0.893s
```

client/client_test.go
```diff
@@ -203,6 +203,27 @@ func (s *clientTestSuite) TestConn_Select() {
...
+
+	str := `insert into test.mixer_test_conn (id, str, f, e, u, i,j) values (6, "a", 3.14, "test1", 255, -127,'{"key1": 1, "key2": null}'),(7, NULL,NULL, NULL, NULL, NULL,NULL)`
+	result, err := s.c.Execute(str)
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), uint64(2), result.AffectedRows)
+
+	str = `select * from mixer_test_conn where id in (6,7)`
+	ss := [][]string{{"6", "a", "3.14", "test1", "255", "-127", "{\"key1\": 1, \"key2\": null}"}, {"7", "", "", "", "", "", ""}}
```
lance6716, line 213: Thanks for adding test. LGTM now.

And I remember in https://github.com/go-mysql-org/go-mysql/issues/821#issuecomment-1722831730 you still need to wrap `'` to `String()` or use `NULL`. Do you think it's easier to use we implement these logic inside `String()`?

like

```golang
func (fv *FieldValue) String() string {
        if fv.Type == mysql.FieldValueTypeNull {
		return "NULL"
	}
	return "'" + string(fv.str) + "'"
}
```

lance6716: and we also check if the Type is number, we don't need to wrap `'`

zhujintao: not so good, ```String()``` only needs to give the raw value. usually users will judge the ```NULL``` type and combine strings

like

```go
c.ExecuteSelectStreaming(mysql_select_all_query, &result, func(row []mysql.FieldValue) error {
	values := make([]string, 0, len(row))
	for _, v := range row {
		if v.Type == mysql.FieldValueTypeNull {
			values = append(values, "NULL")
		} else {
			values = append(values, fmt.Sprintf("'%s'", v.String()))
		}
	}
	r := "INSERT INTO xx VALUES (" + strings.Join(values, ",") + ")"
	fmt.Println(r)

	return nil
}, nil)
```

zhujintao: insert into null value don't need to wrap '

lance6716: I think we don't need to insist "String() only needs to give the raw value". This is the comment of fmt.Stringer

```
// Stringer is implemented by any value that has a String method,
// which defines the “native” format for that value.
// The String method is used to print values passed as an operand
// to any format that accepts a string or to an unformatted printer
// such as Print.
```

I noticed that "String method is used to print values", so it better we directly let developer uses String() to build the SQL string rather than need to deal with NULL. As for what's "the 'native' format for that value", I think the String() should return a valid MySQL liternal https://dev.mysql.com/doc/refman/8.0/en/literals.html

lance6716: > insert into null value don't need to wrap '

I didn't say it need to wrap `'`.

zhujintao: "String() only needs to give the raw value" raw value is ```LengthEncodedString(p[pos:])``` return value

lance6716: "String() only needs to give the raw value" is only your definition, I still prefer "String() should return a valid MySQL liternal https://dev.mysql.com/doc/refman/8.0/en/literals.html"

zhujintao: using a different name?  just return the string value ```FieldValue.str```

lance6716: OK, and I think I can implement String() later, and then your application can directly use

```golang
	values := make([]string, 0, len(row))
	for _, v := range row {
		values = append(values, v.String())
	}
	r := "INSERT INTO xx VALUES (" + strings.Join(values, ",") + ")"
```

Maybe we can revert to

```golang
func (fv *FieldValue) Raw() []byte {
	return fv.str
}
```

lance6716: I guess I can implement String() in this or next weekend. Can you wait for me? If you are in hurry to support your application I can merge this PR after you change to `func (fv *FieldValue) Raw() []byte` function signature.

zhujintao: thank! wait you.

lance6716:
Hi @zhujintao , can you check https://github.com/go-mysql-org/go-mysql/pull/825?

zhujintao:
> Hi @zhujintao , can you check #825?

good

PR #825, opened 2023-09-22
add String() for FieldValue

lance6716:
close https://github.com/go-mysql-org/go-mysql/issues/821

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -458,7 +458,7 @@ CREATE TABLE field_value_test (
 INSERT INTO field_value_test VALUES (
     1, 2, 3, 4.5, 6.7, 
     '2019-01-01', '2019-01-01 01:01:01.123', '2019-01-01 01:01:01.1234', '01:01:01.12345', 2019,
-    'char', 'varchar', 'binary', 'varbinary', 'blob', 'a', 'a,b', '{"a": 1}',
+    'cha\'r', 'varchar', 'binary', 'varbinary', 'blob', 'a', 'a,b', '{"a": 1}',
     NULL
 )`)
 	require.NoError(s.T(), err)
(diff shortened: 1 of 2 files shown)


client/client_test.go
```diff
@@ -423,3 +423,71 @@ func (s *clientTestSuite) TestStmt_Trans() {
...
+		`1`, "'\x02'", `3`, `'4.50000'`, `6.7`,
+		`'2019-01-01'`, `'2019-01-01 01:01:01.123'`, `'2019-01-01 01:01:01.1234'`, `'01:01:01.12345'`, `2019`,
+		`'char'`, `'varchar'`, "'binary\x00\x00\x00\x00'", `'varbinary'`, `'blob'`, `'a'`, `'a,b'`, `'{"a": 1}'`,
+		`NULL`,
+	}
+	for i, v := range result.Values[0] {
+		require.Equal(s.T(), expected[i], v.String())
+	}
+
+	// test can directly use to build a SQL, through it's not safe in most cases
```
lance6716, line 479: @zhujintao do you think it's OK to use?

lance6716 (commented):
@atercattus PTAL, not sure if you have spare time

client/client_test.go
lance6716:
```suggestion
	// test can directly use to build a SQL, though it's not safe in most cases
```

client/client_test.go
```diff
@@ -423,3 +423,71 @@ func (s *clientTestSuite) TestStmt_Trans() {
...
+		`'2019-01-01'`, `'2019-01-01 01:01:01.123'`, `'2019-01-01 01:01:01.1234'`, `'01:01:01.12345'`, `2019`,
+		`'cha\'r'`, `'varchar'`, "'binary\x00\x00\x00\x00'", `'varbinary'`, `'blob'`, `'a'`, `'a,b'`, `'{"a": 1}'`,
+		`NULL`,
+	}
+	for i, v := range result.Values[0] {
+		require.Equal(s.T(), expected[i], v.String())
+	}
+
+	// test can directly use to build a SQL, though it's not safe in most cases
+	sql := fmt.Sprintf("INSERT INTO field_value_test VALUES (%s)", strings.Join(expected, ","))
```
lance6716, line 480: @zhujintao do you think it's OK to use?

zhujintao: yes, it works very well

PR #826, opened 2023-10-04
feat:add code comments & refactor mysql config constant

CocaineCong:
## add code comments:

I found the code comments of this program are not enough which make the program hardly to understand...
so I add some code comments through my understanding of the program.

## refactor mysql config constant
I just remove the mysql default constant to the same file.

### DIFF
--- a/canal/config.go
+++ b/canal/config.go
@@ -122,7 +122,7 @@ func NewConfig(data string) (*Config, error) {
 	return &c, nil
 }
 
-// NewDefaultConfig Init some default config for Canal
+// NewDefaultConfig initiates some default config for Canal
 func NewDefaultConfig() *Config {
 	c := new(Config)
 
--- a/canal/config.go
+++ b/canal/config.go
@@ -8,11 +8,12 @@ import (
 	"time"
 
 	"github.com/BurntSushi/toml"
-	"github.com/go-mysql-org/go-mysql/client"
-	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
 	"github.com/siddontang/go-log/log"
 	"github.com/siddontang/go-log/loggers"
+
+	"github.com/go-mysql-org/go-mysql/client"
+	"github.com/go-mysql-org/go-mysql/mysql"
 )
 
 type DumpConfig struct {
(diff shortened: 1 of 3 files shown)


canal/config.go
```diff
@@ -121,19 +122,18 @@ func NewConfig(data string) (*Config, error) {
 	return &c, nil
 }
 
+// NewDefaultConfig Init some default config for Canal
```
lance6716, line 125:
```suggestion
// NewDefaultConfig initiates some default config for Canal
```

mysql/position.go
```diff
@@ -6,12 +6,22 @@ import (
...
+// Compare the position information between the p and o,
+// if p > o return 1 means the position of p is further back than o.
+// example is following
+// id | project
+// 1  | math
+// 2  | cs
+// 3  | ee
+// p at id 3 and o at id 1.
+// so this way we can determine from which position we should start syncing the data
+// so that it is not consumed repeatedly.
```
lance6716, line 24:
```suggestion
// Compare the position information between the p and o,
// if p > o return 1 means the position of p is further back than o.
```

because the example is too simple for binlog replication scenario, I think adding that example will cause more trouble. For example, multiple row changes may have same position, and not all position can (re)start a binlog replication.

mysql/position.go
```diff
@@ -32,6 +42,13 @@ func (p Position) String() string {
 	return fmt.Sprintf("(%s, %d)", p.Name, p.Pos)
 }
 
+// CompareBinlogFileName in this func, we'll compare the position between a and b by their filenames.
```
lance6716, line 45:
```suggestion
// CompareBinlogFileName compares the binlog filename of a and b.
```

mysql/position.go
```diff
@@ -32,6 +42,13 @@ func (p Position) String() string {
...
+// if a>b will return 1.
+// if b>a will return -1.
+// the binlog filename consists of two parts (just like this --> [basename.00000x], for example bin_log.000001),
+// one part is the `bin_log_basename` from the MySQL configuration file,
+// and the other part is the number of the file's serial number, incrementing from 000001.
+// you can use `show variables like 'log_%';` or `show variables like 'binlog%';` to show your configuration about binlog.
```
lance6716, line 51:
```suggestion
```

can you link the MySQL document about binlog filenames in the comment?

CocaineCong:
PTAL

lance6716 (approved):
Don't worry, any contribution is welcome

PR #827, opened 2023-10-06
feat: handle subevents in transaction payload event

hjweddie:
When using canal to handle MySQL 8.0 Compressed binlog, I find that it could not get any row.
Hope that we can support handle TransactionPayloadEvent in canal, not just in replication package

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -242,6 +242,12 @@ func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
 
 	t, err := c.GetTable(schemaName, tableName)
 	if err != nil {
+		e := errors.Cause(err)
+		// ignore errors below
+		if e == ErrExcludedTable || e == schema.ErrTableNotExist || e == schema.ErrMissingTableMeta {
+			err = nil
+		}
+
 		return err
 	}
 	var action string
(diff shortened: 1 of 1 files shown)


canal/sync.go
```diff
@@ -106,6 +106,23 @@ func (c *Canal) runSyncBinlog() error {
 				}
 			}
 			continue
+		case *replication.TransactionPayloadEvent:
+			// handle subevent row by row
+			ev := ev.Event.(*replication.TransactionPayloadEvent)
+			for _, subEvent := range ev.Events {
+				err = c.handleRowsEvent(subEvent)
```
lance6716, line 113: Thanks for contributing!
these lines seems almost same with line 97~105, do you have time to unify them? like create a closure for one `*replication.RowsEvent`

hjweddie: I move some error check into handleRowsEvent, it may be less duplicated code

canal/sync.go
```diff
@@ -96,12 +96,17 @@ func (c *Canal) runSyncBinlog() error {
 			// we only focus row based event
 			err = c.handleRowsEvent(ev)
 			if err != nil {
-				e := errors.Cause(err)
-				// if error is not ErrExcludedTable or ErrTableNotExist or ErrMissingTableMeta, stop canal
-				if e != ErrExcludedTable &&
-					e != schema.ErrTableNotExist &&
```
lance6716, line 102: in the old code, the error may come from any `return` branch in `handleRowsEvent`, but in your change we only checked the last `return`. I made a quick check that the three errors of

```
		if e != ErrExcludedTable &&
			e != schema.ErrTableNotExist &&
			e != schema.ErrMissingTableMeta {
```

seems only come from `c.GetTable`

hjweddie: I agree with you. In this case, I think the old code means to ignore three errors: ErrExcludedTable,schema.ErrTableNotExist, schema.ErrMissingTableMeta.  So we can do this just after GetTable in handleRowsEvent

lance6716 (approved):

PR #828, opened 2023-10-09
Add comment about golang type of RowsEvent.Rows

lance6716:
We met a problem that use `==` to compare the elements of `RowsEvent.Rows`. So I think we'd better use comments to notify the user that there are various types and should take care all of them.

https://github.com/pingcap/tiflow/issues/9834


lance6716:
ptal @atercattus

dveeden (approved):

Issue #829, opened 2023-10-11
add a EventCacheSize variable to BinlogSyncerConfig to limit the streamer's memory usage


zing22845:
When the consumption speed of the event chan in streamer cannot catch up with its production speed, leading to an accumulation of events, the current fixed channel size of 10240 might occupy significant memory, potentially triggering an Out Of Memory (OOM) condition. Making the size of the event chan configurable would allow for controlling the memory usage of the streamer. https://github.com/go-mysql-org/go-mysql/blob/f0f861769f2f746cdcd3d3f7c3b1025ceedd0c9e/replication/binlogstreamer.go#L90

PR #830, opened 2023-10-11
Add EventCacheCount as member of BinlogSyncerConfig to limit streamer's event channel size

zing22845:
#829 
I have tested on my own MySQL instance with an 8GB size Binlog, when the channel size is the default 10240, during the Binlog transmission, occasional nearly 2GB memory usage will be seen. 
Additionally, I have written a test program to calculate the theoretical maximum memory usage size under different channel size situations.  The theoretical result match the test. 

This is the theoretical usage calc function: 
`
func (c *MySQLClient) GetMaxWindowSize(ctx context.Context, streamerLength uint16) (es *EventStatistic, err error) {
	es = new(EventStatistic)
	es.StreamerChannelLength = streamerLength
	startTime := time.Now()

	// get streamer
	_, streamer, err := c.GetBinlogStreamer(true, streamerLength)
	if err != nil {
		return nil, err
	}
	// defer syncer.Close()
	defer func() {
		es.Duration = time.Since(startTime)
	}()

	// move window
	var window []uint32
	var currentWindowSize uint32
	var ev *replication.BinlogEvent

	var filename string
	var offset uint32

	var f *os.File
	defer func() {
		if f != nil {
			f.Close()
		}
	}()

	// get event time
	parser := replication.NewBinlogParser()
	parser.SetRawMode(true)
	for {
		eventCTX, cancel := context.WithTimeout(ctx, 5*time.Second)
		ev, err = streamer.GetEvent(eventCTX)
		cancel()
		if err == context.DeadlineExceeded {
			return es, nil
		}
		if ev == nil {
			return es, nil
		}
		es.Count++
		// calc the event size
		eventSize := ev.Header.EventSize
		es.TotalSize += uint64(eventSize)

		// add event to the window
		window = append(window, eventSize)
		currentWindowSize += eventSize

		// move event out of window
		if uint16(len(window)) > streamerLength {
			currentWindowSize -= window[0]
			window = window[1:]
		}

		if currentWindowSize > es.MaxStreamerChannelSize {
			es.MaxStreamerChannelSize = currentWindowSize
		}

		offset = ev.Header.LogPos

		if ev.Header.EventType == replication.ROTATE_EVENT {
			rotateEvent := ev.Event.(*replication.RotateEvent)
			filename = string(rotateEvent.NextLogName)

			if ev.Header.Timestamp == 0 || offset == 0 {
				// fake rotate event
				continue
			}
		} else if ev.Header.EventType == replication.FORMAT_DESCRIPTION_EVENT {
			// FormateDescriptionEvent is the first event in binlog, we will close old one and create a new

			if f != nil {
				f.Close()
			}

			if len(filename) == 0 {
				return nil, errors.Errorf("empty binlog filename for FormateDescriptionEvent")
			}

			f, err = os.OpenFile(path.Join("./", filename), os.O_CREATE|os.O_WRONLY, 0644)
			if err != nil {
				return nil, errors.Trace(err)
			}

			// write binlog header fe'bin'
			if _, err = f.Write(replication.BinLogFileHeader); err != nil {
				return nil, errors.Trace(err)
			}
		}

		if n, err := f.Write(ev.RawData); err != nil {
			return nil, errors.Trace(err)
		} else if n != len(ev.RawData) {
			return nil, errors.Trace(io.ErrShortWrite)
		}
		if es.Count > 1 && ev.Header.EventType == replication.ROTATE_EVENT {
			return es, nil
		}
	}
}`

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -123,7 +123,7 @@ type BinlogSyncerConfig struct {
 
 	DiscardGTIDSet bool
 
-	StreamerChanSize int
+	EventCacheSize int
 }
 
 // BinlogSyncer syncs binlog event from server.
@@ -168,8 +168,8 @@ func NewBinlogSyncer(cfg BinlogSyncerConfig) *BinlogSyncer {
 		dialer := &net.Dialer{}
 		cfg.Dialer = dialer.DialContext
 	}
-	if cfg.StreamerChanSize == 0 {
-		cfg.StreamerChanSize = 10240
+	if cfg.EventCacheSize == 0 {
+		cfg.EventCacheSize = 10240
 	}
 
 	// Clear the Password to avoid outputing it in log.
@@ -398,7 +398,7 @@ func (b *BinlogSyncer) prepare() error {
 func (b *BinlogSyncer) startDumpStream() *BinlogStreamer {
 	b.running = true
 
-	s := NewBinlogStreamerWithChanSize(b.cfg.StreamerChanSize)
+	s := NewBinlogStreamerWithChanSize(b.cfg.EventCacheSize)
 
 	b.wg.Add(1)
 	go b.onStream(s)
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -122,6 +122,8 @@ type BinlogSyncerConfig struct {
 	RowsEventDecodeFunc func(*RowsEvent, []byte) error
 
 	DiscardGTIDSet bool
+
+	StreamerChanSize int
 }
 
 // BinlogSyncer syncs binlog event from server.
@@ -165,6 +167,9 @@ func NewBinlogSyncer(cfg BinlogSyncerConfig) *BinlogSyncer {
 	if cfg.Dialer == nil {
 		dialer := &net.Dialer{}
 		cfg.Dialer = dialer.DialContext
+	}
+	if cfg.StreamerChanSize == 0 {
+		cfg.StreamerChanSize = 10240
 	}
 
 	// Clear the Password to avoid outputing it in log.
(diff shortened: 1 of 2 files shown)


replication/binlogsyncer.go
```diff
@@ -122,6 +122,8 @@ type BinlogSyncerConfig struct {
 	RowsEventDecodeFunc func(*RowsEvent, []byte) error
 
 	DiscardGTIDSet bool
+
+	EventCacheSize int
```
lance6716, line 126: The name "XXXSize" makes me think it's using the unit of memory, (EventCacheSize = 1024 means the cache will not exceed 1KB), maybe "EventCacheCount" is a better name.

lance6716 (commented):
rest LGTM. 

This PR is good enough. And in future if you find it's hard to set a fixed value of cache count due to binlog events vary in event size, we can introduce another memory-based limit. For example, in `GetEvent` and `parseEvent` we maintain the sum of approximate memory consumption in cache queue, and block the streamer from reading the binlog when it costs too much memory.

replication/binlogsyncer.go
zing22845: Agreed! I'll work on the fixes.

zing22845: Directly limiting memory usage requires dynamically calculating the memory occupancy of the Event Cache during the streaming process, which might lead to a decrease in overall efficiency. 
In scenarios where the network conditions are not too poor, even setting the EventCacheCount to a single digit won't have much worse performance compared to 10240, while at the same time, the memory of the Event Cache will be limited to the size of a few Events, which is enough to achieve a balance between efficiency and resource occupancy.

lance6716 (approved):

PR #831, opened 2023-10-11
Skip decoding of excluded table rows

mcuelenaere:
NOTE: this PR is untested atm, hence why it is a draft

This PR changes `canal` to skip decoding `RowsEvent` events for tables that it is not interested in (cfr `IncludeTableRegex`, `ExcludeTableRegex`).

The rationale for implementing this, is that we noticed that quite a bit of compute went to parsing JSON fields, even though we were not listening to any tables containing JSON fields. This PR should fix that.

### DIFF
--- a/replication/row_event.go
+++ b/replication/row_event.go
@@ -1018,6 +1018,14 @@ func (e *RowsEvent) decodeExtraData(data []byte) (err2 error) {
 }
 
 func (e *RowsEvent) DecodeData(pos int, data []byte) (err2 error) {
+	if e.compressed {
+		data, err2 = DecompressMariadbData(data[pos:])
+		if err2 != nil {
+			//nolint:nakedret
+			return
+		}
+	}
+
 	// Rows_log_event::print_verbose()
 
 	var (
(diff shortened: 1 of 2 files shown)


replication/row_event.go
```diff
@@ -1018,6 +1018,13 @@ func (e *RowsEvent) decodeExtraData(data []byte) (err2 error) {
 }
 
 func (e *RowsEvent) DecodeData(pos int, data []byte) (err2 error) {
+	if e.compressed {
+		data, err2 = DecompressMariadbData(data[pos:])
+		if err2 != nil {
+			return
```
lance6716, line 1024: sorry we have enabled aggressive linter rules, you should be able to add `//nolint:nakedret` in above line to ignore it.

lance6716:
Hi, don't forget to change the PR state from draft to ready when you think it's OK

mcuelenaere:
Okay, after doing some brief testing this seems to work! I was unable to verify the `DecompressMariadbData` change however, as this was tested against MySQL 5.7 (but it should be safe)

lance6716 (approved):
thanks!

PR #833, opened 2023-10-19
Add a optional function to decode optional meta of table map event

zing22845:
I have a hacked version of MySQL which has a UDF optional meta in table map event. I managed to add a TableMapOptionalMetaDecodeFunc to decode the optional meta.

### DIFF
--- a/replication/binlogstreamer.go
+++ b/replication/binlogstreamer.go
@@ -87,7 +87,7 @@ func (s *BinlogStreamer) closeWithError(err error) {
 func NewBinlogStreamer() *BinlogStreamer {
 	s := new(BinlogStreamer)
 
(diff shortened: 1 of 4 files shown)


zing22845:
PTAL @lance6716

lance6716 (commented):
lgtm with a quick glance, will check again tomorrow.

Would you mind to share with us what's the provider of the hacked MySQL? Is it a cloud RDS?

zing22845:
It's ok.

zing22845:
We added the column info for rows events in earlier 5.6 version. Unfortunately these instances can not be upgrade and our DTS depends on those infos in optional meta.

lance6716 (approved):

PR #834, opened 2023-10-27
Update pingcap/tidb parser to new location

morgo:
The tidb parser moved in https://github.com/pingcap/tidb/pull/47123

This updates the dependency to the new location.


lance6716:
> Error: ../../../go/pkg/mod/github.com/pingcap/tidb/pkg/parser@v0.0.0-20231027131104-30288c77c7e8/charset/charset.go:17:2: package cmp is not in GOROOT (/opt/hostedtoolcache/go/1.20.10/x64/src/cmp)

Oh TiDB's parser seems have used go 1.21 standard libraries, but this library doesn't. And I'm a bit worried more users of this library have not preprared to upgrade to go 1.21. @morgo Do you need the latest parser or we can choose a go 1.20 commit ?

morgo:
> Oh TiDB's parser seems have used go 1.21 standard libraries, but this library doesn't. And I'm a bit worried more users of this library have not preprared to upgrade to go 1.21. @morgo Do you need the latest parser or we can choose a go 1.20 commit ?

Sorry, I didn't test early than go 1.21. I don't need the latest, and can chose a go 1.20 commit.

My motivation was just to bump it because tidb only recently started officially advertising 8.0 compatibility. I know the parser was more or less there for a while, but there have been some syntax changes even within 8.0's release, so I wanted to make sure it was based on something closer to master.

.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ "1.20", "1.19", "1.18" ]
+        go: [ "1.21", "1.20", "1.19", "1.18" ]
```
lance6716, line 8:
```suggestion
        go: [ "1.21", "1.20", "1.19", "1.18" ]
```

lance6716 (approved):

Issue #843, opened 2024-01-17
canal and sync can't send DDL sql


deadspacewii:

deadspacewii:
canal and sync can't send DDL sql like create index xx on tablename and drop index xx on tablename,how can I get it

deadspacewii:
![image](https://github.com/go-mysql-org/go-mysql/assets/17780173/a283394f-765e-4222-b791-8bacc39c25e9)
parseStmt do not support *ast.DropIndexStmt and *ast.CreateIndexStmt

PR #844, opened 2024-01-17
add support for createIndexStmt and dropIndexStmt

deadspacewii:
fix create index indexName on tableName and drop index on DDL statment


deadspacewii:
This PR addresses the changes requested in #843

dveeden:
Would it be possible to add some tests?

deadspacewii:
> Would it be possible to add some tests?
it's hard to repeat this situation, but i can show you:

deadspacewii:
> Would it be possible to add some tests?
I have create three tests，you can see and run

canal/canal_test.go
```diff
@@ -323,3 +323,93 @@ func TestWithoutSchemeExp(t *testing.T) {
...
+
+	baseTable := "test"
+	db := "test"
+	pr := parser.New()
+	for _, s := range cases {
+		stmts, _, err := pr.Parse(s, "", "")
+		require.NoError(t, err)
+		for _, st := range stmts {
+			nodes := parseStmt(st)
+			if len(nodes) == 0 {
```
lance6716, line 355: I prefer we use `require` package rather than the raw `testing` package to keep the coding style. In this line, you should be able to use `require.NotZero(t, len(nodes))`

canal/canal_test.go
```diff
@@ -323,3 +323,93 @@ func TestWithoutSchemeExp(t *testing.T) {
...
+				t.Fatalf("TestCreateIndexExp: parse fail\n")
+				continue
+			}
+			for _, node := range nodes {
+				rdb := node.db
+				rtable := node.table
+				if (len(rdb) != 0 && rdb != db) || rtable != baseTable {
```
lance6716, line 362: and here, you can use `require.Equal(t, db, rdb)` and `require.Equal(t, baseTable, rtable)`

lance6716 (commented):
thanks for your PR. please check my comment and apply it for the DROP INDEX test cases

canal/canal_test.go
deadspacewii: I have apply require package to test

lance6716 (approved):

PR #846, opened 2024-01-22
Update client pool logic

atercattus:
client/pool does not report connection problems properly: in cases of authorization errors, version incompatibility, incorrect database address, network problems, etc.
Also, we have https://github.com/go-mysql-org/go-mysql/pull/701 with a request for new functionality in the same case.

This PR updates the pool logic a bit: more user-friendly function `NewPoolWithOptions`, and option `WithNewPoolPingTimeout` for doing connect&ping during pool initialization.

Also, this fixes `make test-local` after https://github.com/go-mysql-org/go-mysql/pull/803.

### DIFF
--- a/client/pool.go
+++ b/client/pool.go
@@ -77,43 +77,45 @@ var (
 	MaxNewConnectionAtOnce = 5
 )
 
-// NewPool initializes new connection pool and uses params: addr, user, password, dbName and options.
-// minAlive specifies the minimum number of open connections that the pool will try to maintain.
-// maxAlive specifies the maximum number of open connections (for internal reasons,
-// may be greater by 1 inside newConnectionProducer).
-// maxIdle specifies the maximum number of idle connections (see DefaultIdleTimeout).
-func NewPool(
-	logFunc LogFunc,
-	minAlive int,
-	maxAlive int,
-	maxIdle int,
+// NewPoolWithOptions initializes new connection pool and uses params: addr, user, password, dbName and options.
+func NewPoolWithOptions(
 	addr string,
 	user string,
 	password string,
 	dbName string,
-	options ...func(conn *Conn),
-) *Pool {
-	if minAlive > maxAlive {
-		minAlive = maxAlive
+	options ...PoolOption,
+) (*Pool, error) {
+	po := poolOptions{
+		addr:     addr,
+		user:     user,
+		password: password,
+		dbName:   dbName,
 	}
-	if maxIdle > maxAlive {
-		maxIdle = maxAlive
+	for _, o := range options {
+		o(&po)
 	}
-	if maxIdle <= minAlive {
-		maxIdle = minAlive
+
+	if po.minAlive > po.maxAlive {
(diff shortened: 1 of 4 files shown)


client/pool.go
```diff
@@ -512,3 +579,34 @@ func (pool *Pool) Close() {
 	pool.synchro.idleConnections = nil
 	pool.synchro.Unlock()
 }
+
+// checkConnection tries to connect and ping DB server
+func (pool *Pool) checkConnection(ctx context.Context) error {
+	errChan := make(chan error, 1)
+
+	go func() {
```
lance6716, line 587: I guess we need an extra goroutine because `pool.connect` does not use `pool`'s context. If we modify `pool.connect` to

```
connect: func() (*Conn, error) {
  ctx, cancel := context.WithTimeout(pool.ctx, ...)
  defer cancel()
  return ConnectWithDialer(ctx, ...)
}
```

we don't need this extra goroutine. Also it's OK to not change.

client/pool_test.go
```diff
@@ -35,8 +42,40 @@ func (s *poolTestSuite) TestPool_Close() {
...
+func (s *poolTestSuite) TestPool_WrongPassword() {
+	addr := fmt.Sprintf("%s:%s", *test_util.MysqlHost, s.port)
+
+	_, err := NewPoolWithOptions(addr, *testUser, "wrong-password", "",
+		WithPoolLimits(5, 10, 5),
+		WithLogFunc(log.Debugf),
+		WithNewPoolPingTimeout(time.Second),
+	)
+
+	if (err == nil) || !strings.Contains(err.Error(), "ERROR 1045 (28000): Access denied for user") {
```
lance6716, line 60: this can be replaced by `require.ErrorContains`

lance6716 (commented):
rest lgtm

client/pool.go
atercattus: Unfortunately, for a test case like this
```go
l, _ := net.Listen("tcp4", "127.0.0.1:0")
laddr, _ := l.Addr().(*net.TCPAddr)
//_ = l.Close() // do NOT close!

_, err = NewPoolWithOptions(laddr.String(), *testUser, *testPassword, "",
	WithNewPoolPingTimeout(time.Second),
)
```

this dialling hangs forever:
```go
ctx, cancel := context.WithTimeout(context.Background(), time.Second)
dialer := &net.Dialer{}
conn, err := ConnectWithDialer(ctx, "", addr, user, password, dbName, dialer.DialContext, po.connOptions...)
cancel()
```
This behaviour needs to be investigated separately.

lance6716 (approved):

atercattus:
@lance6716 sorry, one more time :)

Together we didn't notice that I accidentally [committed the wrong thing](https://github.com/go-mysql-org/go-mysql/pull/846/commits/b1d7cd8603217a890cbd85981625998be5d5f708) to the Makefile:
image

lance6716 (approved):
I also double checked that `make test-local` can run

PR #849, opened 2024-02-01
Refactor StartBackup to Support Custom Backup Handlers

doraemonkeys:
This pull request introduces a significant refactor to the StartBackup method within the BinlogSyncer type. The primary goal of this change is to enhance flexibility and extensibility by allowing custom handlers for the backup process, rather than directly interacting with the file system within the method.

### DIFF
--- a/replication/backup.go
+++ b/replication/backup.go
@@ -45,7 +45,7 @@ func (b *BinlogSyncer) StartBackupToFile(backupDir string, p Position, timeout t
 // The function writes the raw data of each event to the current file and handles errors such as context deadline exceeded (timeout),
 // write errors, or short writes.
 func (b *BinlogSyncer) StartBackup(p Position, timeout time.Duration,
-	handler func(filename string) (io.WriteCloser, error)) (retErr error) {
+	handler func(binlogFilename string) (io.WriteCloser, error)) (retErr error) {
 	if timeout == 0 {
 		// a very long timeout here
 		timeout = 30 * 3600 * 24 * time.Second
--- a/replication/backup.go
+++ b/replication/backup.go
@@ -45,7 +45,7 @@ func (b *BinlogSyncer) StartBackupToFile(backupDir string, p Position, timeout t
 // The function writes the raw data of each event to the current file and handles errors such as context deadline exceeded (timeout),
 // write errors, or short writes.
 func (b *BinlogSyncer) StartBackup(p Position, timeout time.Duration,
-	handler func(filename string) (io.WriteCloser, error)) error {
+	handler func(filename string) (io.WriteCloser, error)) (retErr error) {
 	if timeout == 0 {
 		// a very long timeout here
 		timeout = 30 * 3600 * 24 * time.Second
@@ -64,8 +64,12 @@ func (b *BinlogSyncer) StartBackup(p Position, timeout time.Duration,
 
 	var w io.WriteCloser
 	defer func() {
+		var closeErr error
 		if w != nil {
-			w.Close()
+			closeErr = w.Close()
+		}
+		if retErr == nil {
+			retErr = closeErr
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


replication/backup.go
```diff
@@ -13,7 +13,39 @@ import (
...
+// and enters a loop to continuously fetch and write events.
+//
+// For each event, it checks the event type. If it's a ROTATE_EVENT, it updates the filename to the next log file as indicated by the event.
+// If it's a FORMAT_DESCRIPTION_EVENT, it signifies the start of a new binlog file, and the function closes the current file (if open) and opens
+// a new one using the handler function. It also writes the binlog file header to the new file.
+//
+// The function writes the raw data of each event to the current file and handles errors such as context deadline exceeded (timeout),
+// write errors, or short writes.
+func (b *BinlogSyncer) StartBackup(p Position, timeout time.Duration,
+	handler func(filename string) (io.WriteCloser, error)) error {
```
lance6716, line 48:
```suggestion
	handler func(binlogFilename string) (io.WriteCloser, error)) error {
```

It will be more clear about the meaning of filename. Maybe also update the comment of this function.

replication/backup.go
```diff
@@ -13,7 +13,39 @@ import (
...
+// The backup process can be controlled with a timeout duration, after which the backup will stop if no new events are received.
+//
+// Parameters:
+//   - p Position: The starting position in the binlog from which to begin the backup.
+//   - timeout time.Duration: The maximum duration to wait for new binlog events before stopping the backup process.
+//     If set to 0, a default very long timeout (30 days) is used instead.
+//   - handler func(filename string) (io.WriteCloser, error): A function provided by the caller to handle file creation and writing.
+//     This function is expected to return an io.WriteCloser for the specified filename, which will be used to write binlog events.
+//
+// The function first checks if a timeout is specified, setting a default if not. It then enables raw mode parsing for binlog events
```
lance6716, line 37: From this paragraph till the end of comments, they are explain the implementation of the function, I think it's better to move them inside the function (maybe the first line in the function body?), rather than stay here as a function-level comment.

Just my own opinion, I want to see a short explanaion, the meaning of arguments and return values, calling conventions and short examples of the function in function-level comments. If I want to check the implementation details, I can go inside the function to see source codes and more context-related comments

replication/backup.go
```diff
@@ -67,26 +95,26 @@ func (b *BinlogSyncer) StartBackup(backupDir string, p Position, timeout time.Du
 		} else if e.Header.EventType == FORMAT_DESCRIPTION_EVENT {
 			// FormateDescriptionEvent is the first event in binlog, we will close old one and create a new
 
-			if f != nil {
-				f.Close()
+			if w != nil {
+				_ = w.Close()
```
lance6716, line 99: I think the error of `Close` is important, for example, some implementations only flush the data when `Close`. In other words, `Write` may only store data in memory and can not guarantee the persistence. 

also for line 68

doraemonkeys:
@lance6716 Your opinions are good and I have changed the code

replication/backup.go
```diff
@@ -13,7 +13,26 @@ import (
...
+
+// StartBackup starts the backup process for the binary log using the specified position and handler.
+// The process will continue until the timeout is reached or an error occurs.
+//
+// Parameters:
+//   - p: The starting position in the binlog from which to begin the backup.
+//   - timeout: The maximum duration to wait for new binlog events before stopping the backup process.
+//     If set to 0, a default very long timeout (30 days) is used instead.
+//   - handler: A function that takes a binlog filename and returns an WriteCloser for writing raw events to.
+func (b *BinlogSyncer) StartBackup(p Position, timeout time.Duration,
```
lance6716, line 34: Would you consider to rename this function to `StartBackupWithHandler`, and let the existing function keep the name `StartBackup`? Just for the compatibility. 

It's not a problem because other PRs also modify the public APIs and we can make a major version release to declare the breaking changes, but I think renaming your new function is also acceptable? Please leave your opinions.

rest lgtm

doraemonkeys: Renaming this function is acceptable for compatibility purposes

lance6716 (approved):

PR #860, opened 2024-04-26
allow setting the collation in auth handshake

dvilaverde:
This PR is to allow the collation to be set during the auth handshake in order to avoid requiring calls to `SET NAMES` post connection.

### DIFF
--- a/client/auth.go
+++ b/client/auth.go
@@ -269,7 +269,7 @@ func (c *Conn) writeAuthHandshake() error {
 	data[11] = 0x00
 
 	// Charset [1 byte]
-	// use default collation id 33 here, is utf-8
+	// use default collation id 33 here, is `utf8mb3_general_ci`
 	collationName := c.collation
 	if len(collationName) == 0 {
 		collationName = DEFAULT_COLLATION_NAME
--- a/client/auth.go
+++ b/client/auth.go
@@ -279,7 +279,14 @@ func (c *Conn) writeAuthHandshake() error {
 		return fmt.Errorf("invalid collation name %s", collationName)
 	}
 
-	data[12] = byte(collation.ID)
+	// the MySQL protocol calls for the collation id to be sent as 1, where only the
+	// lower 8 bits are used in this field. But wireshark shows that the first by of
+	// the 23 bytes of filler is used to send the upper 8 bits of the collation id.
+	// see https://github.com/mysql/mysql-server/pull/541
+	data[12] = byte(collation.ID & 0xff)
+	if collation.ID > 255 {
+		data[13] = byte(collation.ID >> 8)
+	}
 
 	// SSL Connection Request Packet
 	// http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::SSLRequest
@@ -302,6 +309,11 @@ func (c *Conn) writeAuthHandshake() error {
 
 	// Filler [23 bytes] (all 0x00)
 	pos := 13
+	if collation.ID > 255 {
+		// skip setting the first byte of the filler to 0x00 since it is used to
+		// send the upper 8 bits of the collation id
+		pos++
+	}
 	for ; pos < 13+23; pos++ {
 		data[pos] = 0
 	}
--- a/client/auth.go
+++ b/client/auth.go
@@ -281,12 +281,12 @@ func (c *Conn) writeAuthHandshake() error {
 
 	// the MySQL protocol calls for the collation id to be sent as 1, where only the
 	// lower 8 bits are used in this field. But wireshark shows that the first by of
-	// the 23 bytes of filler is used to send the upper 8 bits of the collation id.
+	// the 23 bytes of filler is used to send the right middle 8 bits of the collation id.
 	// see https://github.com/mysql/mysql-server/pull/541
 	data[12] = byte(collation.ID & 0xff)
-	if collation.ID > 255 {
-		data[13] = byte(collation.ID >> 8)
-	}
+	// if the collation ID is <= 255 the middle 8 bits are 0s so this is the equivalent of
+	// padding the filler with a 0.
+	data[13] = byte((collation.ID & 0xff00) >> 8)
 
 	// SSL Connection Request Packet
 	// http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::SSLRequest
@@ -308,13 +308,12 @@ func (c *Conn) writeAuthHandshake() error {
 	}
 
 	// Filler [23 bytes] (all 0x00)
-	pos := 13
-	if collation.ID > 255 {
-		// skip setting the first byte of the filler to 0x00 since it is used to
-		// send the upper 8 bits of the collation id
-		pos++
-	}
-	for ; pos < 13+23; pos++ {
+	// the filler starts at position 13, but the first byte of the filler
+	// maybe have been set by the collation id earlier. So we only position 13
+	// will be either 0x00 or the right middle 8 bits of the collation id. Therefore
+	// here we start at position 14 and fill the remaining 22 bytes with 0x00.
+	pos := 14
+	for ; pos < 14+22; pos++ {
 		data[pos] = 0
 	}
 
--- a/client/auth.go
+++ b/client/auth.go
@@ -5,11 +5,11 @@ import (
 	"crypto/tls"
 	"encoding/binary"
 	"fmt"
-	"github.com/pingcap/tidb/pkg/parser/charset"
 
 	. "github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/packet"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/pkg/parser/charset"
 )
 
 const defaultAuthPluginName = AUTH_NATIVE_PASSWORD
--- a/client/auth.go
+++ b/client/auth.go
@@ -310,7 +310,7 @@ func (c *Conn) writeAuthHandshake() error {
 
 	// Filler [23 bytes] (all 0x00)
 	// the filler starts at position 13, but the first byte of the filler
-	// has been set earlier with collaction id earlier, so position 13 at this point
+	// has been set earlier with collation id earlier, so position 13 at this point
 	// will be either 0x00 or the right middle 8 bits of the collation id.
 	// Therefore, we start at position 14 and fill the remaining 22 bytes with 0x00.
 	pos := 14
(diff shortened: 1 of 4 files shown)


client/auth.go
```diff
@@ -269,7 +270,16 @@ func (c *Conn) writeAuthHandshake() error {
 
 	// Charset [1 byte]
 	// use default collation id 33 here, is utf-8
```
dveeden, line 272:
```suggestion
	// use default collation id 33 here, is `utf8mb3_general_ci`
```

```
mysql> SHOW COLLATION WHERE id=33;
+--------------------+---------+----+---------+----------+---------+---------------+
| Collation          | Charset | Id | Default | Compiled | Sortlen | Pad_attribute |
+--------------------+---------+----+---------+----------+---------+---------------+
| utf8mb3_general_ci | utf8mb3 | 33 | Yes     | Yes      |       1 | PAD SPACE     |
+--------------------+---------+----+---------+----------+---------+---------------+
1 row in set (0.00 sec)
```

client/auth.go
```diff
@@ -269,7 +270,16 @@ func (c *Conn) writeAuthHandshake() error {
...
+	collationName := c.collation
+	if len(collationName) == 0 {
+		collationName = DEFAULT_COLLATION_NAME
+	}
+	collation, err := charset.GetCollationByName(collationName)
+	if err != nil {
+		return fmt.Errorf("invalid collation name %s", collationName)
+	}
+
+	data[12] = byte(collation.ID)
```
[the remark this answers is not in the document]
dveeden: Weirdly enough the protocol says this is 1 byte and that only the low 8-bits are put in this field. Not sure how that's going to work.

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_response.html

image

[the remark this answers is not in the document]
dveeden: A MySQL 8.0 Client only allows the user to set the charset, not the collation:
```
  --default-character-set=name 
                      Set the default character set.
```

And as all default collations are in the 0-255 range this works with the protocol.


```
mysql> SELECT MAX(ID) FROM information_schema.COLLATIONS WHERE IS_DEFAULT='Yes';
+---------+
| MAX(ID) |
+---------+
|     255 |
+---------+
1 row in set (0.00 sec)
```

dveeden:
So this needs testing for collations that have an id that's >255.

Also what is the reason for doing this? Is it connection setup performance? Are you always using an UTF-8 collation or something else?

dveeden:
```
mysql> SELECT *,ID&0x00ff FROM information_schema.COLLATIONS WHERE COLLATION_NAME IN('utf8mb4_ja_0900_as_cs','latin1_bin') ORDER BY ID;
+-----------------------+--------------------+-----+------------+-------------+---------+---------------+-----------+
| COLLATION_NAME        | CHARACTER_SET_NAME | ID  | IS_DEFAULT | IS_COMPILED | SORTLEN | PAD_ATTRIBUTE | ID&0x00ff |
+-----------------------+--------------------+-----+------------+-------------+---------+---------------+-----------+
| latin1_bin            | latin1             |  47 |            | Yes         |       1 | PAD SPACE     |        47 |
| utf8mb4_ja_0900_as_cs | utf8mb4            | 303 |            | Yes         |       0 | NO PAD        |        47 |
+-----------------------+--------------------+-----+------------+-------------+---------+---------------+-----------+
2 rows in set (0.00 sec)
```
So if only 1 byte gets send, both these collations will have the same value.

```python
In [24]: c = mysql.connector.connect(host='127.0.0.1',port=3306,user="test",password="test",collation="utf8mb4_unicode_520_ci", ssl_disabled=True)

In [25]: cur = c.cursor()

In [26]: cur.execute("SHOW SESSION VARIABLES LIKE '%collation%'");

In [27]: cur.fetchall()
Out[27]: 
[('collation_connection', 'utf8mb4_unicode_520_ci'),
 ('collation_database', 'utf8mb4_0900_ai_ci'),
 ('collation_server', 'utf8mb4_0900_ai_ci'),
 ('default_collation_for_utf8mb4', 'utf8mb4_0900_ai_ci')]

In [28]: cur.close()
Out[28]: True

In [29]: c.close()

In [30]: c = mysql.connector.connect(host='127.0.0.1',port=3306,user="test",password="test",collation="utf8mb4_ja_0900_as_cs", ssl_disabled=True)

In [31]: cur = c.cursor()

In [32]: cur.execute("SHOW SESSION VARIABLES LIKE '%collation%'");

In [33]: cur.fetchall()
Out[33]: 
[('collation_connection', 'utf8mb4_ja_0900_as_cs'),
 ('collation_database', 'utf8mb4_0900_ai_ci'),
 ('collation_server', 'utf8mb4_0900_ai_ci'),
 ('default_collation_for_utf8mb4', 'utf8mb4_0900_ai_ci')]

In [34]: cur.close()
Out[34]: True

In [35]: c.close()
```

```
In [40]: c = mysql.connector.connect(host='127.0.0.1',port=3306,user="test",password="test",collation="latin1_bin", ssl_disabled=True)

In [41]: cur = c.cursor()

In [42]: cur.execute("SHOW SESSION VARIABLES LIKE '%collation%'");

In [43]: cur.fetchall()
Out[43]: 
[('collation_connection', 'latin1_bin'),
 ('collation_database', 'utf8mb4_0900_ai_ci'),
 ('collation_server', 'utf8mb4_0900_ai_ci'),
 ('default_collation_for_utf8mb4', 'utf8mb4_0900_ai_ci')]
```


image



image


So, the protocol docs are wrong and Wireshark is also not decoding this properly. I'll follow up on both.

This can actually be two bytes (see the "Unused", which isn't all zeros in one of the cases)

dveeden:
https://github.com/mysql/mysql-server/pull/541

dveeden:
https://gitlab.com/wireshark/wireshark/-/merge_requests/15464

dvilaverde:
> So this needs testing for collations that have an id that's >255.
> 
> Also what is the reason for doing this? Is it connection setup performance? Are you always using an UTF-8 collation or something else?

Hi yes our reason to do this is we have an application, using `latin1_general_ci` and MariaDB 5.5.x, that is sensitive to latency and we setup and tear down connections a a lot, primarily because there is a large number of MariaDB instances that application needs to interact with. Ideally we'd like reduce the connection setup time as much as possible.

dvilaverde:
I believe I've addressed the > 255 collation ID and added a test to assert the bytes are correctly set in the auth handshake. I am hoping I've understood your comments and feedback correctly on this PR and addressed all the issues. Please let me know if I've misunderstood and you were looking for something else.

dvilaverde:
Let me know if you prefer to follow the documentation for the protocol and limit the use of collation in the auth handshake to collations with an ID < 255. This would allow users to set the mysql 8.x `utf8mb4_0900_ai_ci` default if needed while remaining true to the protocol.

client/auth.go
```diff
@@ -268,8 +269,24 @@ func (c *Conn) writeAuthHandshake() error {
...
+	if err != nil {
+		return fmt.Errorf("invalid collation name %s", collationName)
+	}
+
+	// the MySQL protocol calls for the collation id to be sent as 1, where only the
+	// lower 8 bits are used in this field. But wireshark shows that the first by of
+	// the 23 bytes of filler is used to send the upper 8 bits of the collation id.
+	// see https://github.com/mysql/mysql-server/pull/541
+	data[12] = byte(collation.ID & 0xff)
+	if collation.ID > 255 {
```
dveeden, line 287: It might be possible always encode the `collation.ID` to a 2 byte integer, if the right endianness is used. That might simplify the code a bit

client/auth.go
```diff
@@ -292,6 +309,11 @@ func (c *Conn) writeAuthHandshake() error {
 
 	// Filler [23 bytes] (all 0x00)
 	pos := 13
+	if collation.ID > 255 {
+		// skip setting the first byte of the filler to 0x00 since it is used to
+		// send the upper 8 bits of the collation id
+		pos++
+	}
 	for ; pos < 13+23; pos++ {
```
dveeden, line 317: Is the 23 here correct? Or should it be 22 if the collation is 2 bytes?

dveeden:
@lance6716 PTAL

client/auth.go
dvilaverde: > Is the 23 here correct? Or should it be 22 if the collation is 2 bytes?

yes you are correct, should be 22

lance6716:
I'll take a look tomorrow

client/auth_test.go
```diff
@@ -34,3 +38,73 @@ func TestConnGenAttributes(t *testing.T) {
...
+		}
+
+		// sanity check: validate the 22 bytes of filler with value 0x00 are set correctly
+		for i := 14; i < 14+22; i++ {
+			require.Equal(t, byte(0x00), handShakeResponse[i])
+		}
+
+		// and finally the username
+		password := string(handShakeResponse[36:40])
+		require.Equal(t, "test", password)
```
lance6716, line 84:
```suggestion
		username := string(handShakeResponse[36:40])
		require.Equal(t, "test", username)
```

client/auth_test.go
```diff
@@ -34,3 +38,73 @@ func TestConnGenAttributes(t *testing.T) {
...
+		db:             "test",
+		password:       "test",
+		proto:          "tcp",
+		collation:      collation,
+		salt:           ([]byte)("123456781234567812345678"),
+	}
+
+	go func() {
+		err := c.writeAuthHandshake()
+		require.NoError(t, err)
```
lance6716, line 107: we can also close the `c` here

client/client_test.go
```diff
@@ -228,6 +232,21 @@ func (s *clientTestSuite) TestConn_SetCharset() {
 	require.NoError(s.T(), err)
 }
 
+func (s *clientTestSuite) TestConn_SetCollationAfterConnect() {
+	err := s.c.SetCollation("latin1_swedish_ci")
+	require.Error(s.T(), err)
```
lance6716, line 237: we can also check the error message is readable by `require.ErrorContains`. Not a big problem

client/conn.go
```diff
@@ -357,6 +363,20 @@ func (c *Conn) SetCharset(charset string) error {
 	}
 }
 
+func (c *Conn) SetCollation(collation string) error {
+	if c.status == 0 {
```
lance6716, line 367: I'm a bit worried that after handshake `status` can also be `0` (will check it later). It's safer to use other fields like `c.serverVersion`?

lance6716 (commented):
rest lgtm

dvilaverde:
> rest lgtm

I've addressed the feedback from the review.

dveeden (approved):

lance6716 (approved):

Issue #862, opened 2024-05-02
Enabling compression on client connection results in "connection was bad" error


dvilaverde:
When compression is enabled on the client my application fails with `connection was bad`, the nested exception is:

```
io.CopyN failed. err unexpected EOF, copied 0, expected 29
```

But when disabling compression the application works correctly.

### Repro Steps

Start MariaDB in Docker:

```
docker run --rm -e MYSQL_ROOT_PASSWORD=root -p 3306:3306 mariadb:10.3.39
```

Create the DB and table found in the attached file

[db_schema.txt](https://github.com/go-mysql-org/go-mysql/files/15192830/db_schema.txt)

Run this test:

```
import (
  "github.com/go-mysql-org/go-mysql/client"
  "github.com/go-mysql-org/go-mysql/mysql"
  "github.com/stretchr/testify/assert"
)

func TestServer_ShowFieldTable1(t *testing.T) {
	a := assert.New(t)

	c, err := client.Connect("localhost:3306", "root", "root", "mydb", func(conn *client.Conn) {
		conn.SetCapability(mysql.CLIENT_COMPRESS) // Disable compression and the test will pass
	})
	a.NoError(err)

	execute, err := c.Execute("SHOW FIELDS FROM mydb.table1")
	if a.NoError(err) {
		rs := execute.Resultset
		a.Equal(575, len(rs.RowDatas))
	}
}

```

dvilaverde:
Some more context in case it helps. I notice that setting `net_buffer_length` on the server causes the above test case to pass, for example:

```
set global net_buffer_length=32768;
```

then resetting it lot the original value will revert to the previous failing behavior.

```
set global net_buffer_length=16384;
```

dveeden:
Thanks for the bugreport. I'll try to reproduce this.

dvilaverde:
I've created a PR that seems to fix this in my fork: https://github.com/go-mysql-org/go-mysql/pull/863

dvilaverde:
Merged https://github.com/go-mysql-org/go-mysql/pull/863

PR #863, opened 2024-05-03
fixing bad connection error when reading large compressed packets

dvilaverde:
This is a fix for https://github.com/go-mysql-org/go-mysql/issues/862

### DIFF
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -206,7 +206,7 @@ func (c *Conn) copyN(dst io.Writer, src io.Reader, n int64) (written int64, err 
 			}
 
 			// now read the remaining bytes into the buffer containing the first read bytes
-			rd, err = io.ReadAtLeast(c.compressedReader, buf[rd:], bcap-rd)
+			rd, err = io.ReadAtLeast(c.currentPacketReader(), buf[rd:], bcap-rd)
 			n -= int64(rd)
 		}
 
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -66,8 +66,6 @@ type Conn struct {
 
 	compressedHeader [7]byte
 
-	compressedReaderActive bool
-
 	compressedReader io.Reader
 }
 
@@ -108,13 +106,12 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 	}()
 
 	if c.Compression != MYSQL_COMPRESS_NONE {
-		if !c.compressedReaderActive {
+		if c.compressedReader == nil {
 			var err error
 			c.compressedReader, err = c.newCompressedPacketReader()
 			if err != nil {
 				return nil, err
 			}
-			c.compressedReaderActive = true
 		}
 	}
 
@@ -315,7 +312,6 @@ func (c *Conn) WritePacket(data []byte) error {
 			return errors.Wrapf(ErrBadConn, "Write failed. only %v bytes written, while %v expected", n, len(data))
 		}
 		c.compressedReader = nil
-		c.compressedReaderActive = false
 	default:
 		return errors.Wrapf(ErrBadConn, "Write failed. Unsuppored compression algorithm set")
 	}
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -119,11 +119,11 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 	}
 
 	if c.compressedReader != nil {
-		if err := c.ReadPacketTo(buf, c.compressedReader); err != nil {
+		if err := c.ReadPacketTo(buf); err != nil {
 			return nil, errors.Trace(err)
 		}
 	} else {
-		if err := c.ReadPacketTo(buf, c.reader); err != nil {
+		if err := c.ReadPacketTo(buf); err != nil {
 			return nil, errors.Trace(err)
 		}
 	}
@@ -177,6 +177,14 @@ func (c *Conn) newCompressedPacketReader() (io.Reader, error) {
 	return nil, nil
 }
 
+func (c *Conn) currentPacketReader() io.Reader {
+	if c.Compression == MYSQL_COMPRESS_NONE && c.compressedReader == nil {
+		return c.reader
+	} else {
+		return c.compressedReader
+	}
+}
+
 func (c *Conn) copyN(dst io.Writer, src io.Reader, n int64) (written int64, err error) {
 	for n > 0 {
 		bcap := cap(c.copyNBuf)
@@ -193,7 +201,8 @@ func (c *Conn) copyN(dst io.Writer, src io.Reader, n int64) (written int64, err 
 		// if we've read to EOF, and we have compression then advance the sequence number
 		// and reset the compressed reader to continue reading the remaining bytes
 		// in the next compressed packet.
-		if c.Compression != MYSQL_COMPRESS_NONE && rd < bcap && goErrors.Is(err, io.ErrUnexpectedEOF) {
+		if c.Compression != MYSQL_COMPRESS_NONE && rd < bcap &&
+			(goErrors.Is(err, io.ErrUnexpectedEOF) || goErrors.Is(err, io.EOF)) {
 			// we have read to EOF and read an incomplete uncompressed packet
 			// so advance the compressed sequence number and reset the compressed reader
 			// to get the remaining unread uncompressed bytes from the next compressed packet.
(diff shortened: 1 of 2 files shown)


jnewmano:
I tried running your fork and got a panic, I'll work on getting test case that I can share to reproduce.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x49d587]

goroutine 1 [running]:
io.ReadAtLeast({0x0, 0x0}, {0xc000000000, 0x4, 0x4000}, 0x4)
        /usr/local/go/src/io/io.go:335 +0x67
github.com/go-mysql-org/go-mysql/packet.(*Conn).copyN(0xc00013e070, {0x8bd0e0, 0xc001f0e030}, {0x7fe96f835c00, 0xc00007a550}, 0x1ac000?)
        /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:209 +0x206
github.com/go-mysql-org/go-mysql/packet.(*Conn).ReadPacketTo(0xc00013e070, {0x8bd0e0, 0xc001f0e000})
        /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:237 +0xc5
github.com/go-mysql-org/go-mysql/packet.(*Conn).ReadPacketReuseMem(0xc00013e070, {0xc000b8e000, 0x1570cb, 0x1ac000})
        /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:121 +0x165
github.com/go-mysql-org/go-mysql/client.(*Conn).readResultRows(0xc0000a0340, 0xc000106060, 0x1)
```

dvilaverde:
> I tried running your fork and got a panic, I'll work on getting a test case that I can share to reproduce.
> 
> ```
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x49d587]
> 
> goroutine 1 [running]:
> io.ReadAtLeast({0x0, 0x0}, {0xc000000000, 0x4, 0x4000}, 0x4)
>         /usr/local/go/src/io/io.go:335 +0x67
> github.com/go-mysql-org/go-mysql/packet.(*Conn).copyN(0xc00013e070, {0x8bd0e0, 0xc001f0e030}, {0x7fe96f835c00, 0xc00007a550}, 0x1ac000?)
>         /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:209 +0x206
> github.com/go-mysql-org/go-mysql/packet.(*Conn).ReadPacketTo(0xc00013e070, {0x8bd0e0, 0xc001f0e000})
>         /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:237 +0xc5
> github.com/go-mysql-org/go-mysql/packet.(*Conn).ReadPacketReuseMem(0xc00013e070, {0xc000b8e000, 0x1570cb, 0x1ac000})
>         /go/pkg/mod/github.com/dvilaverde/go-mysql@v0.0.0-20240503125055-5895dda79aff/packet/conn.go:121 +0x165
> github.com/go-mysql-org/go-mysql/client.(*Conn).readResultRows(0xc0000a0340, 0xc000106060, 0x1)
> ```

That would be great. I'll take a look when you provide that.

dvilaverde:
I've pushed one more change that may address the panic you found in your testing.  

Also not sure why `golangci-lint` checks are failing now, they pass when i run locally using version `1.57.2`. The GitHub Action is using `1.58.0` and when i tried with that version it never completes and seems to consume 60gb+ of my system memory.

jnewmano:
The latest commit fixed the panic I was seeing. Thank you.

dvilaverde:
Fixed the golangci-lint issue as well. The latest version 1.58.0 removed 3 linters, so i removed those from the config file.

packet/conn.go
```diff
@@ -175,9 +187,29 @@ func (c *Conn) copyN(dst io.Writer, src io.Reader, n int64) (written int64, err
...
-		rd, err := io.ReadAtLeast(src, buf, bcap)
+		var rd int
+		rd, err = io.ReadAtLeast(src, buf, bcap)
+
 		n -= int64(rd)
 
+		// if we've read to EOF, and we have compression then advance the sequence number
+		// and reset the compressed reader to continue reading the remaining bytes
+		// in the next compressed packet.
+		if c.Compression != MYSQL_COMPRESS_NONE && rd < bcap &&
```
lance6716, line 198: I just read the comment of `ReadAtLeast`, seems for `goErrors.Is(err, io.ErrUnexpectedEOF) || goErrors.Is(err, io.EOF)` we don't need to check `rd < bcap`

packet/conn.go
```diff
@@ -175,9 +187,29 @@ func (c *Conn) copyN(dst io.Writer, src io.Reader, n int64) (written int64, err
...
+			// we have read to EOF and read an incomplete uncompressed packet
+			// so advance the compressed sequence number and reset the compressed reader
+			// to get the remaining unread uncompressed bytes from the next compressed packet.
+			c.CompressedSequence++
+			if c.compressedReader, err = c.newCompressedPacketReader(); err != nil {
+				return written, errors.Trace(err)
+			}
+
+			// now read the remaining bytes into the buffer containing the first read bytes
+			rd, err = io.ReadAtLeast(c.currentPacketReader(), buf[rd:], bcap-rd)
```
lance6716, line 209: should we delete this reading and let the outer loop read it? because it may still meet the EOF error like line 191.

packet/conn.go
```diff
@@ -167,6 +144,41 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
...
+			return zlib.NewReader(limitedReader)
+		case MYSQL_COMPRESS_ZSTD:
+			return zstd.NewReader(limitedReader)
+		}
+	}
+
+	return nil, nil
+}
+
+func (c *Conn) currentPacketReader() io.Reader {
```
lance6716, line 174: there're

```
	compressedReaderActive bool

	compressedReader io.Reader
```

in `Conn`. Seems we can directly check `c.compressedReader == nil` as the returned reader for `currentPacketReader`. And `compressedReaderActive` always has the same value for `c.compressedReader == nil` so we can delete it.

lance6716 (commented):
will review later

packet/conn.go
dvilaverde: When I attempted to delete `compressedReaderActive` the tests in the `client` package all began failing when I ran them with compression enabled. I think this is because `compressedReaderActive` is reset to false in `WritePacket` after writing the compressed packet. So I don't think I can delete it, or at least I feel deleting it is out of scope for this PR.

packet/conn.go
dvilaverde: Yes agree with you after reviewing the `ReadAtLeast` documentation. I've applied the change suggested.

packet/conn.go
dvilaverde: Yes terrific suggestion, i've applied this change.

packet/conn.go
dvilaverde: I've updated this PR with your suggestion. It was 2am for me and I wasn't thinking clearly, but after more sleep, I realized I could easily remove the `compressedReaderActive` boolean property from `Conn`.

lance6716: take care of your health ❤️

lance6716 (approved):

jnewmano:
One of the commits after `22dbb42` has led to broken connections when using compression with the error message `invalid compressed sequence 2 != 1`

dvilaverde:
> One of the commits after `22dbb42` has led to broken connections when using compression with the error message `invalid compressed sequence 2 != 1`

Do you happen to have a way to reproduce? Any other info that could help in diagnosing the issue?

jnewmano:
> > One of the commits after `22dbb42` has led to broken connections when using compression with the error message `invalid compressed sequence 2 != 1`
> 
> Do you happen to have a way to reproduce? Any other info that could help in diagnosing the issue?

I'll get more information together and open an issue.

jnewmano:
> > One of the commits after `22dbb42` has led to broken connections when using compression with the error message `invalid compressed sequence 2 != 1`
> 
> Do you happen to have a way to reproduce? Any other info that could help in diagnosing the issue?

Details posted in issue #871

Issue #864, opened 2024-05-06
Setting collations need adjustment


dveeden:
Follow-up for #860

See also: 
- https://gitlab.com/wireshark/wireshark/-/merge_requests/15539
- https://bugs.mysql.com/bug.php?id=114818 / http://github.com/mysql/mysql-server/pull/541
- https://bugs.mysql.com/bug.php?id=114857 / http://github.com/mysql/mysql-connector-python/pull/96
- https://bugs.mysql.com/bug.php?id=114832

What is needed is that we set the collation in the handshake as 1 byte if the collation <=255 and that we send a `SET NAMES ... COLLATE ...` if the collation >255.

cc @lance6716  @dvilaverde

dvilaverde:
I hope https://github.com/go-mysql-org/go-mysql/pull/865 addresses this issue.

dveeden:
Thanks @dvilaverde !

PR #865, opened 2024-05-06
collation in auth handshake are 1-byte

dvilaverde:
Fix for https://github.com/go-mysql-org/go-mysql/issues/864

This adjustment reverts an earlier change that made the collation in the auth handshake 2 bytes. Now it follows the MySQL packet specification and is 1-byte. If the collation has a ID of > 255 then `SET NAMES ...` is called after establishing the connection.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,13 +2,10 @@ linters:
   disable-all: true
   enable:
     # All code is ready for:
-    - deadcode
     - errcheck
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     - misspell
     - nolintlint
     - goimports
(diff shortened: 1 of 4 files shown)


.golangci.yml
dveeden, line 1: These don't look related to the rest of the PR?

dveeden (approved):

.golangci.yml
dveeden:
```
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [linters_context] structcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
```

Ok. So they are deprecated. Indeed good to remove them. Thanks.

lance6716 (approved):

Issue #873, opened 2024-05-08
to exclude databases with ExcludeTableRegex


gaojijun:
I want to exclude some database when do mysql replication. So I set ExcludeTableRegex to ["mysql\\..*"], and IncludeTableRegex is nil. As a result, non databases are synced. Is this expected?

```
// IncludeTableRegex or ExcludeTableRegex should contain database name
	// Only a table which matches IncludeTableRegex and dismatches ExcludeTableRegex will be processed
	// eg, IncludeTableRegex : [".*\\.canal"], ExcludeTableRegex : ["mysql\\..*"]
	//     this will include all database's 'canal' table, except database 'mysql'
	// Default IncludeTableRegex and ExcludeTableRegex are empty, this will include all tables
	IncludeTableRegex []string `toml:"include_table_regex"`
	ExcludeTableRegex []string `toml:"exclude_table_regex"`
```

PR #874, opened 2024-05-08
Fix no table is replicated when excludeTableRegex is set while includeTableRegex is nil

gaojijun:
fix https://github.com/go-mysql-org/go-mysql/issues/873


lance6716:
Thanks, please fix the lint error

```
  canal/canal.go:299: File is not `goimports`-ed (goimports)
```

And can you update the `canal_test.go` like `TestCanalFilter` to add an unit test for it?

gaojijun:
> Thanks, please fix the lint error
> 
> ```
>   canal/canal.go:299: File is not `goimports`-ed (goimports)
> ```
> 
> And can you update the `canal_test.go` like `TestCanalFilter` to add an unit test for it?

Done

lance6716:
Hi, I mean add a test to explain your change. I manually deleted line 302~303 and the test `TestIncludeExcludeTableRegex` can still pass. I want a failed test without the change as issue said

lance6716:
Can you also update the comment of `IncludeTableRegex` and `ExcludeTableRegex` to explain the behaviout when either is not set? 

My last concern is by Hyrum's Law we should not change the behaviour of API, but I think this change does no harm. Because the old behaviour will replicate nothing so there should be no user using it. As a user, I directly use `BinlogSyncer` and don't use canal package, @froot can you give your opinion?

rest lgtm

gaojijun:
// Default IncludeTableRegex and ExcludeTableRegex are empty, this will include all tables

@lance6716 Is this comment good enough? It's already in the code.

lance6716:
> // Default IncludeTableRegex and ExcludeTableRegex are empty, this will include all tables
> 
> @lance6716 Is this comment good enough? It's already in the code.

It means both are empty, not one of them is empty

gaojijun:
> > // Default IncludeTableRegex and ExcludeTableRegex are empty, this will include all tables
> > @lance6716 Is this comment good enough? It's already in the code.
> 
> It means both are empty, not one of them is empty

updated

lance6716 (approved):

PR #880, opened 2024-05-17
reduce memory allocations when zlib compression is enabled

dvilaverde:
When enabling compression in our production application we're seeing a fairly large increase in memory allocations and GC duration. Based on some profiling I was able to narrow it down to the new creation of the zlib writers and readers that are thrown away after every packet.

I've implemented re-use of the writers using `sync.Pool` along with a few other changes, such as not creating a zilb Writer when the packet data size is less then minCompressionLength to help alleviate the pressure we see on the garbage collector. 

Here is a benchmark for before my changes:

```
go test ./pkg/database/maria/testserver -cpu 1,2,4 -bench=. -run=^# -benchtime=1s -benchmem
goos: darwin
goarch: arm64
pkg: git.company.com/server/test/pkg/database/maria/testserver
BenchmarkInsert_TestServer           664           1762367 ns/op         1529787 B/op        161 allocs/op
BenchmarkInsert_TestServer-2        2130            515707 ns/op         1529917 B/op        162 allocs/op
BenchmarkInsert_TestServer-4        2086            501147 ns/op         1530396 B/op        163 allocs/op
PASS
ok      git.company.com/server/test/pkg/database/maria/testserver     5.038s
```

and this is the benchmark after the changes.

```
go test ./pkg/database/maria/testserver -cpu 1,2,4 -bench=. -run=^# -benchtime=1s -benchmem
goos: darwin
goarch: arm64
pkg: git.company.com/server/test/pkg/database/maria/testserver
BenchmarkInsert_TestServer          2968            346505 ns/op            8734 B/op        101 allocs/op
BenchmarkInsert_TestServer-2        3057            364719 ns/op            9099 B/op        101 allocs/op
BenchmarkInsert_TestServer-4        3032            375451 ns/op           13800 B/op        101 allocs/op
PASS
ok      git.company.com/server/test/pkg/database/maria/testserver     4.951s
```

I've been running my for in production for a few days now making SQL calls at a rate of about 1000/s and the application has been much more stable and we've been able to scale down our cluster size due to the fewer CPU cycles we're spending doing GC.


client/req.go
```diff
@@ -39,19 +39,23 @@ func (c *Conn) writeCommandStr(command byte, arg string) error {
 func (c *Conn) writeCommandUint32(command byte, arg uint32) error {
 	c.ResetSequence()
 
-	return c.WritePacket([]byte{
-		0x05, //5 bytes long
-		0x00,
-		0x00,
-		0x00, //sequence
+	buf := utils.ByteSliceGet(9)
```
dvilaverde, line 42: re-use of byte slices and byte buffers was done on the packet reading path but not so much on the packet writing path. Our application is write heavy and as a result we saw lots of allocations when writing. Some of the changes you'll see in this PR are to re-use byte slices and buffers where it makes sense.

packet/conn.go
```diff
@@ -1,58 +1,32 @@
...
+	"github.com/go-mysql-org/go-mysql/compress"
 	. "github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/utils"
 	"github.com/klauspost/compress/zstd"
 	"github.com/pingcap/errors"
+	"io"
+	"net"
 )
 
-type BufPool struct {
```
dvilaverde, line 23: I didn't see any references to this BufPool

packet/conn.go
```diff
@@ -75,11 +49,9 @@ func NewConn(conn net.Conn) *Conn {
 	c := new(Conn)
 	c.Conn = conn
 
-	c.bufPool = NewBufPool()
-	c.br = bufio.NewReaderSize(c, 65536) // 64kb
-	c.reader = c.br
+	c.reader = c
```
dvilaverde, line 52: ok so not sure about this, but the NewTLSConn doesn't create a buffered reader for this, so I mirrored that here as well. If there is a reason we need the buffered reader here and I can restore it with some comments for future reference.

packet/conn.go
```diff
@@ -320,8 +290,14 @@ func (c *Conn) WritePacket(data []byte) error {
 		} else if n != len(data) {
 			return errors.Wrapf(ErrBadConn, "Write failed. only %v bytes written, while %v expected", n, len(data))
 		}
-		c.compressedReader = nil
+
 		c.compressedReaderActive = false
+		if c.compressedReader != nil {
+			if _, ok := c.compressedReader.(io.ReadCloser); ok {
+				_ = c.compressedReader.(io.ReadCloser).Close()
```
dvilaverde, line 297: make sure the reader gets back into the pool

packet/conn.go
```diff
@@ -331,41 +307,45 @@ func (c *Conn) WritePacket(data []byte) error {
...
-	var w io.WriteCloser
-	minCompressLength := 50
-	compressedHeader := make([]byte, 7)
+	var (
+		compressedLength, uncompressedLength int
+		payload                              *bytes.Buffer
+		compressedHeader                     = make([]byte, 7)
+	)
 
-	switch c.Compression {
```
dvilaverde, line 340: moving this switch so that it only creates a compression writer if `len(data) > MinCompressionLength` otherwise all the allocations that come with creating the writer happen only to have the writer thrown away because compression won't occur.

lance6716:
thanks! I'll take a look next week.

packet/conn.go
```diff
@@ -452,7 +439,7 @@ func (c *Conn) WriteEncryptedPassword(password string, seed []byte, pub *rsa.Pub
 	return errors.Wrap(c.WriteAuthSwitchPacket(enc, false), "WriteAuthSwitchPacket failed")
 }
 
-// http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchResponse
+// WriteAuthSwitchPacket see http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchResponse
```
dveeden, line 442: This should be updated to the doxygen docs: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_auth_switch_response.html

We probably have to do this in multiple places.

dveeden (commented):
Looks like a good change. Only looked at the code, didn't try it yet.

packet/conn.go
```diff
@@ -331,41 +311,45 @@ func (c *Conn) WritePacket(data []byte) error {
...
-		w, err = zlib.NewWriterLevel(&payload, zlib.HuffmanOnly)
-	case MYSQL_COMPRESS_ZSTD:
-		w, err = zstd.NewWriter(&payload)
-	}
-	if err != nil {
-		return 0, err
-	}
+	if len(data) > MinCompressionLength {
+		var w io.WriteCloser
+		payload = utils.BytesBufferGet()
```
lance6716, line 322: we can add a `defer` here to make sure in every branch `payload` is put back

packet/conn.go
```diff
@@ -331,41 +311,45 @@ func (c *Conn) WritePacket(data []byte) error {
...
-	if len(data) > minCompressLength {
-		compressedLength = len(payload.Bytes())
+		compressedLength = payload.Len()
 	} else {
 		compressedLength = len(data)
 	}
 
 	c.CompressedSequence = 0
+	// write the compressed packet header
+	compressedPacket := utils.BytesBufferGet()
```
lance6716, line 352: also this one

lance6716 (commented):
rest lgtm. Thanks!

dvilaverde:
> rest lgtm. Thanks!

I've address your feedback. Thanks for taking the time to look at this.

lance6716 (approved):

Issue #859, opened 2024-04-24
Driver Compression Enablement


kvolmar:
Is there a way to enable compression when using the Driver rather than the Client?

kvolmar:
Specifically I've seen https://github.com/go-mysql-org/go-mysql/blob/7c31dc4e4e5bac6f8f9a73c04efad3e0af69e606/client/conn.go#L124-L128. Where is ccaps set and is there a way to configure them when using the Driver?

lance6716:
> Specifically I've seen
> 
> https://github.com/go-mysql-org/go-mysql/blob/7c31dc4e4e5bac6f8f9a73c04efad3e0af69e606/client/conn.go#L124-L128
> 
> . Where is ccaps set and is there a way to configure them when using the Driver?

https://github.com/go-mysql-org/go-mysql/blob/7c31dc4e4e5bac6f8f9a73c04efad3e0af69e606/client/conn.go#L177-L180

kvolmar:
I am currently using the go-sql-driver/mysql and interested in migrating to go-mysql-org/go-mysql specifically for the driver. I am evaluating the feasibility of using the database/sql library with your driver and enabling compression. I'm currently importing the driver with side effects and able to write to MySQL, but without compression. Are the compression settings configurable with env vars or in the DSN?

lance6716:
Sorry I think this library does not have this way to configurate compression. Welcome to help us implement it, related files should be under `driver` folder.

dvilaverde:
@kvolmar this has been implemented, can you close this issue?

PR #885, opened 2024-06-03
Additional Driver args for compression and connection read/write timeouts 

dvilaverde:
I'm proposing this PR to address the issue https://github.com/go-mysql-org/go-mysql/issues/859 which asks for the ability to enable compression in the connection DSN. I've also added a few other properties:

- compress
- readTimeout
- timeout
- writeTimeout

and I've updated the README.md.


README.md
```diff
@@ -360,6 +360,58 @@ func main() {
...
+[user[:password]@]addr[/db[?param=X]]
+```
+
+#### `compress`
+
+Enable zlib compression between the client and the server.
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| boolean   | false     | user:pass@localhost/mydb?compress=true      |
```
dveeden, line 377: I think it might be better to have something like `compress=zlib` / `compress=zstd` to allow selection of the compression algorithm.

MySQL Client has deprecated `--compress` in favor of `--compression-algorithms=` and `--zstd-compression-level=`.

```
$ mysql --help
...
  -C, --compress      Use compression in server/client protocol.
...
  --compression-algorithms=name 
                      Use compression algorithm in server/client protocol.
                      Valid values are any combination of
                      'zstd','zlib','uncompressed'.
  --zstd-compression-level=# 
                      Use this compression level in the client/server protocol,
                      in case --compression-algorithms=zstd. Valid range is
                      between 1 and 22, inclusive. Default is 3.
```

README.md
```diff
@@ -360,6 +360,58 @@ func main() {
...
+
+#### `readTimeout`
+
+I/O read timeout.
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| duration  | 0         | user:pass@localhost/mydb?readTimeout=10s    |
+
+#### `ssl`
```
dveeden, line 387: Could we rename this to `tls` or would that cause any issues?

README.md
```diff
@@ -360,6 +360,58 @@ func main() {
...
+
+Enable zlib compression between the client and the server.
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| boolean   | false     | user:pass@localhost/mydb?compress=true      |
+
+#### `readTimeout`
+
+I/O read timeout.
```
dveeden, line 381:
```suggestion
I/O read timeout. 0 means no timeout.
```

README.md
dvilaverde: we can, but this is the existing `ssl` configuration so it would break existing clients. I can make it support `tls` and `ssl` options so that it's backward compatible with existing DSN's and going forward we document only the `tis` param in the readme.

dveeden: In that case, I would suggest to leave it as-is.

README.md
```diff
@@ -360,6 +360,56 @@ func main() {
...
+| --------- | ------------- | --------------------------------------- |
+| string    | uncompressed  | user:pass@localhost/mydb?compress=zlib  |
+
+#### `readTimeout`
+
+I/O read timeout. 0 means no timeout.
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| duration  | 0         | user:pass@localhost/mydb?readTimeout=10s    |
```
lance6716, line 385: I see that there's the unit in argument. It's friendly to mention we support golang's https://pkg.go.dev/time#ParseDuration format.

README.md
```diff
@@ -360,6 +360,56 @@ func main() {
...
+
+#### `ssl`
+
+Enable TLS between client and server. Valid values are `true` or `custom`.
```
lance6716, line 389: Also mention that `custom` will use the TLS configuration set by `SetCustomTLSConfig` matching this host

README.md
```diff
@@ -360,6 +360,56 @@ func main() {
...
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| string    |           | user:pass@localhost/mydb?ssl=true           |
+
+#### `timeout`
+
+Timeout is the maximum amount of time a dial will wait for a connect to complete.  0 means no timeout.
```
lance6716, line 397: Also https://pkg.go.dev/time#ParseDuration

README.md
```diff
@@ -360,6 +360,56 @@ func main() {
...
+
+| Type      | Default   | Example                                     |
+| --------- | --------- | ------------------------------------------- |
+| duration  | 0         | user:pass@localhost/mydb?timeout=1m         |
+
+#### `writeTimeout`
+
+I/O write timeout. 0 means no timeout.
```
lance6716, line 405: ditto

client/conn.go
```diff
@@ -108,19 +116,21 @@ func ConnectWithDialer(ctx context.Context, network string, addr string, user st
...
 	c.proto = network
-	c.Conn = packet.NewConn(conn)
 
 	// use default charset here, utf-8
 	c.charset = DEFAULT_CHARSET
 
 	// Apply configuration functions.
-	for i := range options {
-		options[i](c)
+	for _, option := range options {
```
lance6716, line 124: I just found we forget to close the `conn` in line 110 when error happens 😂 Do you think we can fix it in this PR or a separate one?

driver/driver.go
```diff
@@ -92,26 +96,52 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
...
+					})
+				default:
+					return nil, errors.Errorf("Supported options are ssl=true or ssl=custom")
+				}
+			} else if key == "timeout" && len(value) > 0 {
+				if timeout, err = time.ParseDuration(value[0]); err != nil {
+					return nil, errors.Wrap(err, "invalid duration value for timeout option")
+				}
+			} else {
+				if option, ok := options[key]; ok {
```
lance6716, line 128: Do you think we can expose a `SetDSNOptions` to let user add their own modifications to `Conn`?

packet/conn.go
```diff
@@ -310,6 +338,16 @@ func (c *Conn) WritePacket(data []byte) error {
 	return nil
 }
 
+func (c *Conn) writeWithTimeout(b []byte) (n int, err error) {
```
lance6716, line 341: in line 415, the `c.Write` should also be replace by this function

lance6716 (commented):
rest lgtm

client/conn.go
dvilaverde: added `Close()` to 3 additional spots in this function.

driver/driver.go
dvilaverde: I've added `SetDsnOptions()`, hopefully I've understood your request and implemented as you intended.

packet/conn.go
dvilaverde: done

driver/driver.go
```diff
@@ -324,3 +359,19 @@ func SetCustomTLSConfig(dsn string, caPem []byte, certPem []byte, keyPem []byte,
...
+// It requires a full import of the driver (not by side-effects only).
+// Example of supplying a custom option:
+//
+//	driver.SetDsnOptions(map[string]DriverOption{
+//			"my_option": func(c *client.Conn, value string) error {
+//				c.SetCapability(mysql.CLIENT_CAN_HANDLE_EXPIRED_PASSWORDS)
+//				return nil
+//			},
+//		})
+func SetDsnOptions(customOptions map[string]DriverOption) {
```
lance6716, line 373: I recommend name it SetDSNOptions, as the golang's style

https://google.github.io/styleguide/go/decisions.html#initialisms

Don't forget to update README's reference to this function name

lance6716 (approved):
rest lgtm. Thanks

driver/driver.go
dvilaverde: done

Issue #886, opened 2024-06-07
converting argument $1 type: uint64 values with high bit set are not supported


dvilaverde:
When attempting a query that requires an `uint64`, the driver fails with this error:

`sql: converting argument $1 type: uint64 values with high bit set are not supported`

Example:

```
conn, err := sql.Open("mysql", "root@127.0.0.1:3306/test")
defer conn.Close()

stmt, err := conn.Prepare("select a, b from tableA where uint64 = ?")
defer stmt.Close()

var val uint64 = math.MaxUint64
result, err := stmt.Query(val)
```

I don't see a way to use the https://pkg.go.dev/database/sql/driver#NamedValueChecker to allow for `uint64` handling, or any custom Value types for that matter.

dvilaverde:
I've added PR https://github.com/go-mysql-org/go-mysql/pull/887 to address this issue.

PR #887, opened 2024-06-07
add support for driver.NamedValueChecker on driver connection

dvilaverde:
This PR is to address issue https://github.com/go-mysql-org/go-mysql/issues/886, which prevents a client from using custom value types, specifically `uint64` in prepared statements.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -6,6 +6,7 @@ import (
 	"crypto/tls"
 	"database/sql"
 	sqldriver "database/sql/driver"
+	goErrors "errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -25,6 +26,10 @@ var customTLSMutex sync.Mutex
 var (
 	customTLSConfigMap = make(map[string]*tls.Config)
 	options            = make(map[string]DriverOption)
+
+	// can be provided by clients to allow more control in handling Go and database
+	// types beyond the default Value types allowed
+	namedValueCheckers []CheckNamedValueFunc
 )
 
 type driver struct {
@@ -154,8 +159,30 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
 	return &conn{c}, nil
 }
 
+type CheckNamedValueFunc func(*sqldriver.NamedValue) error
+
+var _ sqldriver.NamedValueChecker = &conn{}
+
 type conn struct {
 	*client.Conn
+}
+
+func (c *conn) CheckNamedValue(nv *sqldriver.NamedValue) error {
+	for _, nvChecker := range namedValueCheckers {
+		err := nvChecker(nv)
+		if err == nil {
+			// we've found a CheckNamedValueFunc that handled this named value
+			// no need to keep looking
+			return nil
+		} else {
+			// we've found an error, if the error is driver.ErrSkip then
+			// keep looking otherwise return the unknown error
+			if !goErrors.Is(sqldriver.ErrSkip, err) {
+				return err
+			}
+		}
+	}
+	return sqldriver.ErrSkip
 }
 
 func (c *conn) Prepare(query string) (sqldriver.Stmt, error) {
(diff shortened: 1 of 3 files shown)


driver/driver_options_test.go
```diff
@@ -94,16 +97,50 @@ func TestDriverOptions_writeTimeout(t *testing.T) {
 	srv := CreateMockServer(t)
 	defer srv.Stop()
 
-	conn, err := sql.Open("mysql", "root@127.0.0.1:3307/test?writeTimeout=10")
```
lance6716, line 97: I see that in the old code, `10` is an illegal value and should return error. So line 102 catched it. Now what's the purpose to change it to `1ns`?

driver/driver.go
```diff
@@ -375,3 +402,11 @@ func SetDSNOptions(customOptions map[string]DriverOption) {
 		options[o] = f
 	}
 }
+
+// AddNamedValueChecker sets a custom NamedValueChecker for the driver connection which
+// allows for more control in handling Go and database types beyond the default Value types.
+// See https://pkg.go.dev/database/sql/driver#NamedValueChecker
+// Usage requires a full import of the driver (not by side-effects only).
+func AddNamedValueChecker(nvCheckFunc ...CheckNamedValueFunc) {
```
lance6716, line 410: In function comment, also mention it's not concurrent-safe

lance6716 (commented):
And just for #886, I'm not sure to handle the large unit64 value, simply skipping the checker can work well. (will the driver write legal payload?)

dvilaverde:
> And just for #886, I'm not sure to handle the large unit64 value, simply skipping the checker can work well. (will the driver write legal payload?)

Based on the documentation calling ErrSkip will cause the [DefaultParameterConverter](https://pkg.go.dev/database/sql/driver#DefaultParameterConverter) to be used instead.

driver/driver_options_test.go
dvilaverde: > I see that in the old code, `10` is an illegal value and should return error. So line 102 catched it. Now what's the purpose to change it to `1ns`?

I'm going to add some comments into the test to explain. But you're right in that i unintentionally altered what was being tested so I'll fix that as well.

driver/driver_options_test.go
```diff
@@ -94,16 +97,85 @@ func TestDriverOptions_writeTimeout(t *testing.T) {
...
+	log.SetLevel(log.LevelDebug)
+	srv := CreateMockServer(t)
+	defer srv.Stop()
+	conn, err := sql.Open("mysql", "root@127.0.0.1:3307/test?writeTimeout=1s")
+	require.NoError(t, err)
+	defer conn.Close()
+
+	// the NamedValueChecker will return ErrSkip for types that are NOT uint64, so make
+	// sure those make it to the server ok first.
+	int32Stmt, err := conn.Prepare("select ?")
```
dvilaverde, line 149: lines 149-159 tests that the nameValueChecker falls back to the default name value checker

driver/driver_options_test.go
```diff
@@ -94,16 +97,85 @@ func TestDriverOptions_writeTimeout(t *testing.T) {
...
 	require.Error(t, err)
+	require.Contains(t, err.Error(), "i/o timeout")
+	require.Nil(t, result)
 
 	conn.Close()
 }
 
+func TestDriverOptions_namedValueChecker(t *testing.T) {
+	AddNamedValueChecker(func(nv *sqlDriver.NamedValue) error {
+		rv := reflect.ValueOf(nv.Value)
```
lance6716, line 131: I also checked it can be written as

```
		_, ok := nv.Value.(uint64)
		if !ok {
			return sqlDriver.ErrSkip
		}

		return nil
```

Maybe it's faster to use golang's built-in type cast rather than reflection in your application.

lance6716 (approved):

driver/driver_options_test.go
dvilaverde: Oh yea thanks. This is much better. 

Unrelated question, how often do you, or what is the criteria to, create a tag for a new version?

lance6716: As a user, our project always uses "master" branch 😂 So generally when other contributors ask for a release I ping @atercattus to release it.

PR #888, opened 2024-06-12
Support DSN without 'user:password@'

dvilaverde:
I noticed that a DSN like this:

```
db.example.com:3307/db?ssl=true
```

would incorrectly have the property `db` in the `connInfo` object set to `ssl=true` when it should have been set to `db`.

According to the README the DSN format allows for an optional `user:password@`.

```
[user[:password]@]addr[/db[?param=X]]
```

Adding unit tests for the `ParseDSN` function to cover those cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,14 @@ Configuration options can be provided by the standard DSN (Data Source Name).
 [user[:password]@]addr[/db[?param=X]]
 ```
 
+#### `collation`
+
+Set a collation during the Auth handshake.
+
+| Type      | Default         | Example                                               |
+| --------- | --------------- | ----------------------------------------------------- |
+| string    | utf8_general_ci | user:pass@localhost/mydb?collation=latin1_general_ci  |
+
 #### `compress`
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -368,6 +368,14 @@ Configuration options can be provided by the standard DSN (Data Source Name).
 [user[:password]@]addr[/db[?param=X]]
 ```
 
+#### `collation`
```
dvilaverde, line 371: i forgot to add this connection param in my previous commits

client/auth.go
```diff
@@ -285,10 +285,8 @@ func (c *Conn) writeAuthHandshake() error {
 		return fmt.Errorf("invalid collation name %s", collationName)
 	}
 
-	// the MySQL protocol calls for the collation id to be sent as 1, where only the
```
dvilaverde, line 288: updating this comment which was incorrect regarding the first byte of the 23 bytes of filter being used to store the right 8 bits of the collation id.

lance6716 (approved):

PR #891, opened 2024-06-17
time column type string formatting and test coverage

jnewmano:
This PR improves the string formatting of time type columns, `FormatBinaryTime` does not return properly formatted time strings in all cases. 

In the zero value case, it was incorrectly returning a `date` formatted string instead of the expected zero `time` of `00:00:00`. For positive times there was a leading null byte in the returned []byte.

`CREATE TABLE time_test (ts TIME);`
`INSERT INTO time_test (ts) VALUES ("00:00:00"), ("06:12:10");`

Reading the values out (using a statement as direct queries don't use FormatBinaryTime) 

`SELECT * FROM %s WHERE ts > ?`
with query parameter `-5` or some other number

resulted in

```
"0000-00-00"
"\x0006:12:10"
```

when the expected result is
```
"00:00:00"
"06:12:10"
```

### DIFF
--- a/mysql/util_test.go
+++ b/mysql/util_test.go
@@ -23,3 +23,33 @@ func TestCompareServerVersions(t *testing.T) {
 		require.Equal(t, test.Expect, got)
 	}
 }
+
+func TestFormatBinaryTime(t *testing.T) {
+	tests := []struct {
+		Data   []byte
+		Expect string
+		Error  bool
+	}{
+		{Data: []byte{}, Expect: "00:00:00"},
+		{Data: []byte{0, 0, 0, 0, 0, 0, 0, 10}, Expect: "00:00:10"},
+		{Data: []byte{0, 0, 0, 0, 0, 0, 1, 40}, Expect: "00:01:40"},
+		{Data: []byte{1, 0, 0, 0, 0, 0, 1, 40}, Expect: "-00:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40}, Expect: "-25:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40, 1, 2, 3, 0}, Expect: "-25:01:40.197121"},
+		{Data: []byte{0}, Error: true},
+	}
+
+	for _, test := range tests {
+
+		n := len(test.Data)
+
+		// binary.LittleEndian.PutUint32(data[1:], uint64(test.Seconds))
+		got, err := FormatBinaryTime(n, test.Data)
+		if test.Error {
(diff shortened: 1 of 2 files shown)


mysql/util_test.go
```diff
@@ -23,3 +23,33 @@ func TestCompareServerVersions(t *testing.T) {
...
+		{Data: []byte{0, 0, 0, 0, 0, 0, 0, 10}, Expect: "00:00:10"},
+		{Data: []byte{0, 0, 0, 0, 0, 0, 1, 40}, Expect: "00:01:40"},
+		{Data: []byte{1, 0, 0, 0, 0, 0, 1, 40}, Expect: "-00:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40}, Expect: "-25:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40, 1, 2, 3, 0}, Expect: "-25:01:40.197121"},
+		{Data: []byte{0}, Error: true},
+	}
+
+	for _, test := range tests {
+
```
lance6716, line 43:
```
  mysql/util_test.go:42:30: unnecessary leading newline (whitespace)
```

from CI ⬆️

mysql/util_test.go
```diff
@@ -23,3 +23,32 @@ func TestCompareServerVersions(t *testing.T) {
...
+		{Data: []byte{1, 0, 0, 0, 0, 0, 1, 40}, Expect: "-00:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40}, Expect: "-25:01:40"},
+		{Data: []byte{1, 1, 0, 0, 0, 1, 1, 40, 1, 2, 3, 0}, Expect: "-25:01:40.197121"},
+		{Data: []byte{0}, Error: true},
+	}
+
+	for _, test := range tests {
+		n := len(test.Data)
+
+		// binary.LittleEndian.PutUint32(data[1:], uint64(test.Seconds))
```
lance6716, line 45: we can remove this comment?

mysql/util_test.go
```diff
@@ -23,3 +23,32 @@ func TestCompareServerVersions(t *testing.T) {
...
+		got, err := FormatBinaryTime(n, test.Data)
+		if test.Error {
+			require.Error(t, err)
+		} else if err != nil {
```
lance6716, line 49: we can use 

... else { 
  require.NoError(t. err) 
}

here

lance6716 (commented):
rest lgtm. Thank you!

lance6716 (approved):

PR #892, opened 2024-06-21
allow packet.Conn buffer size to be adjustable

dvilaverde:
Our usage of this library is in an application that will maintain a few thousand active MySQL connections from a single instance and we've noticed that the application is holding a lot of memory in `packet.Conn`, specifically the `bufio.Reader` instances are holding about 500MB of the total 800MB heap.

This PR will allow our application to change the hardcoded `65536` buffer size to something more reasonable for our use case (maybe `4096`), which should reduce our heap by at least 50%.

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -18,6 +18,8 @@ import (
 	"github.com/go-mysql-org/go-mysql/utils"
 )
 
+const defaultBufferSize = 65536 // 64kb
+
 type Option func(*Conn) error
 
 type Conn struct {
@@ -32,6 +34,9 @@ type Conn struct {
 	// Connection read and write timeouts to set on the connection
 	ReadTimeout  time.Duration
 	WriteTimeout time.Duration
+
+	// The buffer size to use in the packet connection
+	BufferSize int
 
 	serverVersion string
 	// server capabilities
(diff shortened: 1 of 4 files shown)


driver/driver_options_test.go
```diff
@@ -73,6 +74,29 @@ func TestDriverOptions_ConnectTimeout(t *testing.T) {
 	conn.Close()
 }
 
+func TestDriverOptions_BufferSize(t *testing.T) {
+	log.SetLevel(log.LevelDebug)
+	srv := CreateMockServer(t)
+	defer srv.Stop()
+
+	SetDSNOptions(map[string]DriverOption{
```
dvilaverde, line 82: our application is already using `SetDSNOptions` so we're planning on using the same technique for the `BufferSize`. I didn't want to add more configuration to the public README of this project because I felt that the default for most people is probably fine and therefore no need to add more configuration options.

packet/conn.go
```diff
@@ -53,19 +53,23 @@ type Conn struct {
 }
 
 func NewConn(conn net.Conn) *Conn {
+	return NewBufferedConn(conn, 65536) // 64kb
```
dvilaverde, line 56: keeping this `NewConn` function for compatibility reasons. As of now just the server code and most unit tests use this. The client close will be using `NewBufferedConn`.

client/conn.go
```diff
@@ -94,6 +99,7 @@ type Dialer func(ctx context.Context, network, address string) (net.Conn, error)
 func ConnectWithDialer(ctx context.Context, network, addr, user, password, dbName string, dialer Dialer, options ...Option) (*Conn, error) {
 	c := new(Conn)
 
+	c.BufferSize = defaultBufferSize
```
dvilaverde, line 102: init the new packet `Conn` with the 64kb buffer size

lance6716 (approved):

PR #894, opened 2024-06-24
fix(canal): do not ignore events with LogPos=0, use timestamp to distinguish fake rotate events

### LINKED ISSUE #893: Canal module does not detect row updates in transactions after upgrading MariaDB to 11.4.1
## Description ##
After upgrading our MariaDB database to version 11.4.1, the canal module of the go-mysql library stopped detecting row updates within transactions. This issue seems to be linked to the handling of fake rotate events that I added a few years ago. The problem stems from an incorrect assumption that events with LogPos=0 can be ignored. However, it appears that the new version of MariaDB does not specify LogPos for certain events:
```
Events written through transaction or statement cache have log_pos set
to 0 so that they can be copied directly to the binlog without having
to compute the real end_log_pos.
```
Relevant code in MariaDB: [MariaDB log_event_server.cc](https://github.com/MariaDB/server/blob/7c5fdc9b6a45f4e8bad1ae5c7c93b457be9891fa/sql/log_event_server.cc#L858) - https://github.com/MariaDB/server/commit/b8f9f796ff8a1b0744cafde715c370db5a07adb7

**Steps to Reproduce:**
1. Setup MariaDB 10.7.7.
2. Configure the canal module to process binlog events.
3. Execute a transaction that updates rows in the database.
4. Check that the row updates are handled correctly.
5. Upgrade MariaDB to version 11.4.1.
6. Execute a transaction that updates rows in the database.
7. Observe that the canal module does not detect the row updates.

**Expected Behavior:**
The canal module should detect and process row updates within transactions, regardless of the LogPos value.

**Actual Behavior:**
The canal module ignores events with LogPos=0, causing it to miss row updates within transactions.

**Environment:**
* MariaDB version: 11.4.1
* go-mysql version: 1.8.0
* OS: Alpine 3.20

## Additional Context ##
The issue seems to be related to a change in [how MariaDB handles LogPos for events written through transaction or statement cache](https://jira.mariadb.org/browse/MDEV-31273). To resolve this, the canal module needs to be updated to handle events with LogPos=0 appropriately.

## Proposed Solution ##
Consider fake rotate events only for events of type ROTATE_EVENT with timestamp = 0. This aligns with the comment in the MySQL codebase indicating that the timestamp is set to 0 to distinguish FAKE_ROTATE_EVENT from ROTATE_EVENT and conforms to the MariaDB documentation on [fake rotate events](https://mariadb.com/kb/en/fake-rotate_event/).

## Related ##
https://github.com/go-mysql-org/go-mysql/issues/396 https://github.com/go-mysql-org/go-mysql/issues/406 https://github.com/go-mysql-org/go-mysql/issues/752

mefcorvi:
After upgrading to MariaDB 11.4.1, the canal module stopped detecting row updates within transactions due to incorrect handling of fake rotate events. MariaDB 11.4.1 does not set LogPos for certain events, causing these events to be ignored. This fix modifies the handling to consider fake rotate events only for ROTATE_EVENTs with timestamp = 0, aligning with MariaDB and MySQL documentation.

Consequences of this change:

* Events with LogPos=0 that are not of type ROTATE_EVENT will no longer be ignored.
* If a database does not set timestamp=0 for fake rotate events, these events will be processed as real rotate events.

For MariaDB 11.4, I have confirmed that the timestamp for fake rotate events is set to 0. Based on the MariaDB and MySQL source code and documentation, I believe we can rely on this behavior.

I am not aware of any other events with LogPos=0 that should be ignored. Originally, this code was intended exclusively for handling fake rotate events and was not meant to ignore other events. Therefore, I believe this change is safe and will not negatively impact the processing of other binlog events.

Fixes #893

@lintanghui @csuzhangxc @gaojijun

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -47,22 +47,22 @@ func (c *Canal) runSyncBinlog() error {
 		// Update the delay between the Canal and the Master before the handler hooks are called
 		c.updateReplicationDelay(ev)
 
-		// If log pos equals zero then the received event is a fake rotate event and
-		// contains only a name of the next binlog file
-		// See https://github.com/mysql/mysql-server/blob/8e797a5d6eb3a87f16498edcb7261a75897babae/sql/rpl_binlog_sender.h#L235
-		// and https://github.com/mysql/mysql-server/blob/8cc757da3d87bf4a1f07dcfb2d3c96fed3806870/sql/rpl_binlog_sender.cc#L899
-		if ev.Header.LogPos == 0 {
-			switch e := ev.Event.(type) {
-			case *replication.RotateEvent:
+		switch e := ev.Event.(type) {
+		case *replication.RotateEvent:
+			// If the timestamp equals zero, the received rotate event is a fake rotate event
+			// and contains only the name of the next binlog file. Its log position should be
+			// ignored.
+			// See https://github.com/mysql/mysql-server/blob/8e797a5d6eb3a87f16498edcb7261a75897babae/sql/rpl_binlog_sender.h#L235
+			// and https://github.com/mysql/mysql-server/blob/8cc757da3d87bf4a1f07dcfb2d3c96fed3806870/sql/rpl_binlog_sender.cc#L899
+			if ev.Header.Timestamp == 0 {
 				fakeRotateLogName := string(e.NextLogName)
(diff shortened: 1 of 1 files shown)


lance6716 (commented):
Thanks! I'll review my application using this library and double check how to handle it soon.

lance6716 (approved):

PR #899, opened 2024-07-24
allow disabling the default golang database/sql retry behavior

dvilaverde:
We've run into a problem caused by the Golangs `database/sql` package [default retry behavior](https://github.com/golang/go/blob/c5430dc1d8d96cfd0c85a6d760262676d798a6a9/src/database/sql/sql.go#L1663) (which can't be disabled by the way) when an [ErrBadConn](https://pkg.go.dev/database/sql/driver@go1.22.5#ErrBadConn) from the `database/sql/driver` is returned by the driver. We're looking to have our application fail fast when an issue occurs reaching the DB, but as of now we're seeing 3 retries, causing our application to hang up longer than we'd like.

Seems like others in the community have been asking for a way to disable retries, see https://github.com/golang/go/issues/52886 but the go developers have decided to not allow developers to configure that behavior.

This PR makes use of the [Validator](https://pkg.go.dev/database/sql/driver@go1.22.5#Validator) interface, along with introducing a new Driver argument called `retries` to determine if [ErrBadConn](https://pkg.go.dev/database/sql/driver@go1.22.5#ErrBadConn) or the go-mysql-org native [ErrBadConn](https://github.com/go-mysql-org/go-mysql/blob/3b1dd0e7703815062f7999fd7d503df0061807a9/mysql/error.go#L10) should be returned.

This takes advantage of the fact that retry won't occur unless the database/sql ErrBadConn is returned.  Per go docs:

> To prevent duplicate operations, ErrBadConn should NOT be returned if there's a possibility that the database server might have performed the operation. Even if the server sends back an error, you shouldn't return ErrBadConn.

And to maintain the behavior of discarding bad connection, the Validator interface will return false for isValid(), preventing the bad connection from being cached for reuse in sql.DB.

Thoughts on this are appreciated.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -97,7 +97,11 @@ func parseDSN(dsn string) (connInfo, error) {
 // Open takes a supplied DSN string and opens a connection
 // See ParseDSN for more information on the form of the DSN
 func (d driver) Open(dsn string) (sqldriver.Conn, error) {
-	var c *client.Conn
+	var (
+		c *client.Conn
+		// by default database/sql driver retries will be enabled
+		retries = true
+	)
 
 	ci, err := parseDSN(dsn)
 
@@ -134,6 +138,10 @@ func (d driver) Open(dsn string) (sqldriver.Conn, error) {
 				if timeout, err = time.ParseDuration(value[0]); err != nil {
 					return nil, errors.Wrap(err, "invalid duration value for timeout option")
 				}
+			} else if key == "retries" && len(value) > 0 {
+				// by default keep the golang database/sql retry behavior enabled unless
+				// the retries driver option is explicitly set to 'false'
+				retries = !strings.EqualFold(value[0], "off")
 			} else {
 				if option, ok := options[key]; ok {
 					opt := func(o DriverOption, v string) client.Option {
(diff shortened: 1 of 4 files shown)


driver/driver_options_test.go
```diff
@@ -31,6 +31,51 @@ type testServer struct {
...
+
+	rows, err := conn.QueryContext(context.TODO(), "select * from slow;")
+	require.Nil(t, rows)
+
+	// we want to get a golang database/sql/driver ErrBadConn
+	require.ErrorIs(t, err, sqlDriver.ErrBadConn)
+
+	// here we issue assert that even though we only issued 1 query, that the retries
+	// remained on and there were 3 calls to the DB.
+	require.Equal(t, 3, srv.handler.queryCount)
```
dvilaverde, line 57: here we had 3 calls to the server mockHandler because the database/sql package retried 3 times and retries were NOT turned off.

driver/driver_options_test.go
```diff
@@ -31,6 +31,51 @@ type testServer struct {
...
+	require.NoError(t, err)
+
+	rows, err := conn.QueryContext(context.TODO(), "select * from slow;")
+	require.Nil(t, rows)
+	// we want the native error from this driver implementation
+	require.ErrorIs(t, err, mysql.ErrBadConn)
+
+	// here we issue assert that even though we only issued 1 query, that the retries
+	// remained on and there were 3 calls to the DB.
+	require.Equal(t, 1, srv.handler.queryCount)
```
dvilaverde, line 78: same test as before but the retries were off and now there is only 1 call to the server mockHandler

lance6716 (commented):
rest lgtm

driver/driver.go
```diff
@@ -222,10 +246,16 @@ func buildArgs(args []sqldriver.Value) []interface{} {
 	return a
 }
 
-func replyError(err error) error {
-	if mysql.ErrorEqual(err, mysql.ErrBadConn) {
+func (st *state) replyError(err error) error {
+	isBadConnection := mysql.ErrorEqual(err, mysql.ErrBadConn)
+
+	if !st.useDriverErrors && isBadConnection {
```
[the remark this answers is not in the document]
dvilaverde: Oh great advice, I will certainly apply this change. I felt the same way but couldn't come up with a way to make it better. So thanks!

lance6716 (approved):

PR #900, opened 2024-07-24
Add `DisableFlushBinlogWhileWaiting` config and almost every query event triggers `OnPosSynced`

morgo:
`ANALYZE TABLE` was not being considered in DDL statements.

This is a problem because if it doesn't return as one of the nodes from `parseStmt()` then it won't set `savePos` to `true` here:

https://github.com/go-mysql-org/go-mysql/blob/3b1dd0e7703815062f7999fd7d503df0061807a9/canal/sync.go#L157-L182

Which means the binary log doesn't advance.

We rely on the binary log to advance to implement our own `BlockWait()` function, which is similar to `c.WaitUntilPos()`. We can't use `WaitUntilPos()` because it calls `FLUSH BINARY LOGS` which requires additional privileges.

We were previously injecting binlog noise to advance the binary log, but with this fix we shouldn't need to do that (and ideally `WaitUntilPos()` doesn't either, so it can catch bugs if statements are left off the list like this one.)

### DIFF
--- a/canal/canal_test.go
+++ b/canal/canal_test.go
@@ -133,8 +133,24 @@ func (s *canalTestSuite) TestCanal() {
 	)
 	s.execute("ALTER TABLE test.canal_test ADD `age` INT(5) NOT NULL AFTER `name`")
 	s.execute("INSERT INTO test.canal_test (name,age) VALUES (?,?)", "d", "18")
-	s.execute("ANALYZE TABLE test.canal_test")
 
+	err := s.c.CatchMasterPos(10 * time.Second)
+	require.NoError(s.T(), err)
+}
+
+func (s *canalTestSuite) TestAnalyzeAdvancesSyncedPos() {
+	<-s.c.WaitDumpDone()
+
+	// We should not need to use FLUSH BINARY LOGS
+	// An ANALYZE TABLE statement should advance the saved position.
+	// There are still cases that don't advance, such as
+	// statements that won't parse like [CREATE|DROP] TRIGGER.
+	s.c.cfg.DisableFlushBinlogWhileWaiting = true
+	defer func() {
+		s.c.cfg.DisableFlushBinlogWhileWaiting = false
+	}()
+
+	s.execute("ANALYZE TABLE test.canal_test")
 	err := s.c.CatchMasterPos(10 * time.Second)
 	require.NoError(s.T(), err)
 }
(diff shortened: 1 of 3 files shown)


dveeden (approved):
It would be good to add a test for this as well if possible.

canal/sync.go
```diff
@@ -235,6 +235,14 @@ func parseStmt(stmt ast.StmtNode) (ns []*node) {
 			table: t.Table.Name.String(),
 		}
 		ns = []*node{n}
+	case *ast.AnalyzeTableStmt:
+		ns = make([]*node, len(t.TableNames))
```
lance6716, line 239: I see that with this change, `OnTableChanged`, `OnDDL` of handler will be called. My concern is

1. ANALYZE will not change the table. I'm not a user of handler interface but I just read the comment of `OnTableChanged`, it said "You need to clear the associated data like cache with the table." I think with this PR ANALYZE will trigger unnecessary cache dropping.
2. ANALYZE is not a DDL, MySQL put it under "Database Administration Statements - Table Maintenance Statements" https://dev.mysql.com/doc/refman/8.4/en/analyze-table.html. 
 
So I recommend we extend the `EventHandler` interface to add a new callback for it?

morgo: What about just setting:
```
savePos = true 
force = true 
```

.. regardless of the len(nodes)?

This would solve my problem since I just need the `OnPosSynced` event to fire.

lance6716: > since I just need the OnPosSynced event to fire.

I think we can let `parseStmt` returns a `savePos`, and if any stmt has a `true` `savePos` we can enter https://github.com/go-mysql-org/go-mysql/blob/3b1dd0e7703815062f7999fd7d503df0061807a9/canal/sync.go#L173-L177

canal/canal_test.go
```diff
@@ -138,6 +138,23 @@ func (s *canalTestSuite) TestCanal() {
...
 }
 
+func (s *canalTestSuite) TestAnalyzeAdvancesSyncedPos() {
+	<-s.c.WaitDumpDone()
+
+	// We should not need to use FLUSH BINARY LOGS
+	// An ANALYZE TABLE statement should advance the saved position.
+	// There are still cases that don't advance, such as
+	// statements that won't parse like [CREATE|DROP] TRIGGER.
+	s.c.cfg.DisableFlushBinlogWhileWaiting = true
```
lance6716, line 148: The database user in this test has enough privilege for FLUSH BINARY LOGS, so it will not ensure FLUSH BINARY LOGS is skipped. I think we can get `s.c.GetMasterPos()` before and after `s.execute("ANALYZE TABLE test.canal_test")` and compare the binlog filename is not changed.

lance6716 (approved):

Issue #903, opened 2024-08-05
Schema load with multi-value index works with `NewTable`, fails with `NewTableFromSqlDB`


shauns:
Hi!

A heads-up that I am by no means a Go developer, but am happy to get my hands dirty on this 🙏 

I found that with tables using [multi-value indexes](https://dev.mysql.com/doc/refman/8.4/en/create-index.html#:~:text=A%20multi%2Dvalued%20index%20is,intended%20for%20indexing%20JSON%20arrays.), the `NewTable` function is able to load the table, but the `NewTableFromSqlDB` function is not. I'll put a concise failing test case below.

It looks like the issue is in loading the column names for the index. MySQL returns a null value for the column name, and in the `NewTableFromSqlDB` function, we fail with `converting NULL to string is unsupported` (In `NewTable`, it uses `""`).

If someone can point in the direction of what to fix, I'm happy to make a contrib. Or, let us know if there's a fundamental blocker that stops this working.

Here's the short failing test case -- the call to `NewTableFromSqlDB` fails.

```go
func (s *schemaTestSuite) TestSchemaWithMultiValueIndex() {
	_, err := s.conn.Execute(`DROP TABLE IF EXISTS multi_value_idx_test`)
	require.NoError(s.T(), err)

	str := `
        CREATE TABLE IF NOT EXISTS multi_value_idx_test (
            id INT,
            entries json,
            PRIMARY KEY(id)
        ) ENGINE = INNODB;
    `

	_, err = s.conn.Execute(str)
	require.NoError(s.T(), err)

	str = `CREATE INDEX idx_entries ON multi_value_idx_test((CAST((entries->'$') AS CHAR(64))));`
	_, err = s.conn.Execute(str)
	require.NoError(s.T(), err)

	ta, err := NewTable(s.conn, *schema, "multi_value_idx_test")
	require.NoError(s.T(), err)

	require.Len(s.T(), ta.Indexes, 2)

	require.Equal(s.T(), "PRIMARY", ta.Indexes[0].Name)
	require.Len(s.T(), ta.Indexes[0].Columns, 1)
	require.Equal(s.T(), "id", ta.Indexes[0].Columns[0])

	require.Equal(s.T(), "idx_entries", ta.Indexes[1].Name)
	require.Len(s.T(), ta.Indexes[1].Columns, 1)
	require.Equal(s.T(), "", ta.Indexes[1].Columns[0])

[...]

PR #904, opened 2024-08-05
Handle null index column names in sql.DB loader

shauns:
For developers using multi-value indexes, the column name as reported by `show index...` is null.

Closes https://github.com/go-mysql-org/go-mysql/issues/903

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -354,7 +354,8 @@ func (ta *Table) fetchIndexesViaSqlDB(conn *sql.DB) error {
 	var unusedVal interface{}
 
 	for r.Next() {
-		var indexName, colName string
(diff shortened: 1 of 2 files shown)


rogeryen:
Hi @lance6716 (or @atercattus), just wondering if there's an ETA on when this fix would be included in a new version/tag? TY

lance6716:
> Hi @lance6716 (or @atercattus), just wondering if there's an ETA on when this fix would be included in a new version/tag? TY

Currently you can use master version in go.mod file. The version release interval is mostly decided by @atercattus 😄

atercattus:
I don't do anything special. Just tag and update the documentation :)
Let's make a new tag.

PR #910, opened 2024-08-28
FIX:The problem of the ignoreTables parameter not taking effect in the go-mysqldump package

liuhuanxg:
FIX:The problem of the ignoreTables parameter not taking effect in the go-mysqldump package


lance6716:
Hi, the linter error should be fixed by https://github.com/go-mysql-org/go-mysql/pull/907 .

liuhuanxg:
ok, thanks

lance6716 (approved):

PR #912, opened 2024-09-01
fixing golangci-lint@1.60.3 errors of current build

dvilaverde:
Noticed that the current build had a failure during golangci-lint check. Simply addressing those.


mysql/util.go
```diff
@@ -130,10 +130,13 @@ func AppendLengthEncodedInteger(b []byte, n uint64) []byte {
 
 func RandomBuf(size int) []byte {
 	buf := make([]byte, size)
-	mrand.Seed(time.Now().UTC().UnixNano())
+	// When this project supports golang 1.20 as a minimum, then this mrand.New(...)
+	// line can be eliminated and the random number can be generated by simply
+	// calling mrand.Intn()
+	random := mrand.New(mrand.NewSource(time.Now().UTC().UnixNano()))
```
dvilaverde, line 136: Although this line is no longer needed with golang 1.20+ and simply the call to `mrand.Intn()` can be used, I've created a new random from a Source because this projects `go.mod` declares golang 1.18. But left a note in the code for future reference.

driver/driver.go
```diff
@@ -203,7 +203,7 @@ func (c *conn) CheckNamedValue(nv *sqldriver.NamedValue) error {
 		} else {
 			// we've found an error, if the error is driver.ErrSkip then
 			// keep looking otherwise return the unknown error
-			if !goErrors.Is(sqldriver.ErrSkip, err) {
+			if !goErrors.Is(err, sqldriver.ErrSkip) {
```
dvilaverde, line 206: the arguments were in the wrong order here.

lance6716 (approved):

Issue #911, opened 2024-08-30
Can't connect to mysql server 8.0 when password is larger than 19 characters 


erfanium:
Affected version: 1.8.0
Latest working version: 1.7.0

# Reproduction Steps
1- mysql server with complex password and caching_sha2_password as default
```yml
services:
  mysql_test:
    image: mysql:8.0.33
    environment:
      MYSQL_ROOT_PASSWORD: "*pl3$#&XiC4Tn*AP13l#OKUY"
      MYSQL_DATABASE: default
      MYSQL_USER: base-user
      MYSQL_PASSWORD: "*pl3$#&XiC4Tn*AP13l#OKUY"
    ports:
      - "3306:3306"
    command: --default-authentication-plugin=caching_sha2_password
```

2- 
```go
package main

import "github.com/go-mysql-org/go-mysql/client"

func main() {
	conn, err := client.Connect("127.0.0.1:3306", "root", "*pl3$#&XiC4Tn*AP13l#OKUY", "default")
	if err != nil {
		panic(err)
	}
	defer conn.Close()

	_, err = conn.Execute(`select 1;`)

	if err != nil {
		panic(err)
	}

	println("connected")
}
```

then you will see this error:
```
panic: handleAuthResult: ERROR 1045 (28000): Access denied for user 'root'@'172.23.0.1' (using password: YES)
```

lance6716:
Thank for reporting, I'll take a look next week

erfanium:
@lance6716 
Thanks! I'm new to golang, but I can contribute on this if you need.

lance6716:
@erfanium Thank you, currently I only tried to reproduce this problem, but I have not started to locate the root cause. This is my testing steps:

```bash
docker run --rm --name mysql-3306 -p 3306:3306 -e MYSQL_ROOT_PASSWORD='*pl3$#&XiC4Tn*AP13l#OKUY' mysql:8.0.33 --log-bin=mysql-bin --port=3306 --bind-address=0.0.0.0 --binlog-format=ROW --server-id=1 --gtid_mode=ON --enforce-gtid-consistency=true --default-authentication-plugin=caching_sha2_password
```

then add this unit test in `client` package

```go
func TestComplexPassword(t *testing.T) {
	conn, err := Connect("127.0.0.1:3306", "root", "*pl3$#&XiC4Tn*AP13l#OKUY", "default")
	require.NoError(t, err)
	defer conn.Close()

[...]

PR #914, opened 2024-09-07
Fix can't connect to MySQL 8.0 with long password

lance6716:
close https://github.com/go-mysql-org/go-mysql/issues/911

Seems we should not use the ending zero in the scramble


lance6716:
ptal @atercattus

GMHDBJD (approved):
LGTM

PR #916, opened 2024-09-11
Add getters to BinlogSyncer

dt8269:


replication/binlogsyncer.go
```diff
@@ -918,3 +918,24 @@ func (b *BinlogSyncer) killConnection(conn *client.Conn, id uint32) {
 	}
 	b.cfg.Logger.Infof("kill last connection id %d", id)
 }
+
+// GetSyncedGTIDSet returns the current GTID set.
+func (b *BinlogSyncer) GetSyncedGTIDSet() GTIDSet {
+	b.m.RLock()
+	defer b.m.RUnlock()
+	return b.currGset
```
lance6716, line 926: 1. I see the function name is "synced" GTID set, but `b.currGset` is not "synchronized to downsteam" GTID set, it just a "synchronized to BinlogSyncer" one. Maybe the function comment can describe it as "the GTID set that BinlogSyncer is fully fetched".
2. However, "fully fetched" does not mean the caller of BinlogSyncer can see it, because `GetEvent` is FIFO and there are backlog event in it. I think `GetSyncedGTIDSet` is meaningless to caller.

dt8269: Thanks for your feedback!

1. Could you clarify or provide documentation that explains what you mean by "synchronized to downstream" in this context?
2. My goal is to pause syncing, fully flush the stream, and then retrieve the resulting GTID set from the BinlogSyncer. In a scenario where there is no backlog of events, wouldn't `GetSyncedGTIDSet` serve a useful purpose?

lance6716: 1. I support your use case is reading the binlog events and synchronize them to elsewhere, like save in files, replicate to another database or message queue. I want to mention the "Synced" in new function name `GetSyncedGTIDSet` does not mean those things.
2. What do you mean by pause syncing? Close the `BinlogSyncer` or stop writing in upstream MySQL? I think some events may stuck in-flight in the network between MySQL and `BinlogSyncer`. So you can't make sure it's flushed. A better way is check events from `GetEvent` and record the GTID related event manually.

dt8269: 1. Yes, but "Synced" is used relative to the BinlogSyncer alone, as it's a function on the BinlogSyncer, right? From its perspective, what events is it aware of? Do you have a suggestion for a name that you think is more correct?
2. I'd like to call `Close()` then `StartSync(...)`. Are you suggesting skipping over `BinlogSyncer` in general and using `BinlogStreamer` directly to make sure I have a consistent stream of events?

lance6716: 2. I think your usage can not ensure correctness.

> I'd like to call Close() then StartSync(...)

How do you use the `BinlogStreamer`? I assume that you have a goroutine reading binlog events from `GetEvent` and process them. And your usage is 

t1: the application has processed many binlog events. 
t2: somehow you want to close the `BinlogStreamer` and restart it later, so record the GTID set using the new getter
t3: now close the `BinlogStreamer` and do some stuff
t4: start `BinlogStreamer` from the GTID set and the application continues.

There's a problem in above workflow. There are residue binlogevents in `BinlogStreamer.ch` that are skipped, because they have not been processed in the time <= t1 and the GTID set get in t2 is larger than them.

lance6716: Sorry I can't understand your use case after these communications. Can you provide code snippet with more details?

dt8269: Yes, here's a rough example:
```
syncer := replication.NewBinlogSyncer(config)
streamer := syncer.StartSync(4)
time.Sleep(5 * time.Second)
syncer.Close()
for {
		select {
		case <-ctx.Done():
			return
		default:
			event, err := streamer.GetEvent(ctx)
			if err != nil {
				log.Println("Error getting event:", err)
				return
			}
		}
	}
syncedGTIDSet := syncer.GetSyncedGTIDSet()
otherReplicaGTIDSet := getOtherGTIDSet()
difference := getGTIDSetDifference(syncedGTIDSet, otherReplicaGTIDSet)
```

lance6716: in this example, `syncedGTIDSet` does not equal the sum of `event` get from `event, err := streamer.GetEvent(ctx)`. Because `s.ech` is set to `ErrSyncClosed` so `s.ch` is not run win in the `select`

https://github.com/go-mysql-org/go-mysql/blob/5c82c1f91206c9cf4535acfafef88353dacaa2f4/replication/binlogstreamer.go#L70-L76

https://github.com/go-mysql-org/go-mysql/blob/5c82c1f91206c9cf4535acfafef88353dacaa2f4/replication/binlogstreamer.go#L30-L34

I still recommend you check the type of every `event` and update the last seen GTID set. For example, `XIDEvent` has a `GSet` field.

dt8269: At a high level, we want a service that participates in semisync and stores the binlogs durably. In some cases, we need to retrieve the full set of transactions that have been processed by a syncer. This should only include transactions that the syncer has ACKed, not those that will be processed in the future.
Currently, the backup and ACKing logic are not ordered in a way that ensures the ACK is sent only after the data is backed up to disk.
In summary:
- We want to ACK after writing to disk.
- We want the ability to retrieve the full GTID set that has been written to disk.

While looking through this, we noticed that `BinlogSyncer.parseEvent` should close before acking when `needStop` and `needACK` are both true. Should I open a separate issue for that?

lance6716: > While looking through this, we noticed that BinlogSyncer.parseEvent should close before acking when needStop and needACK are both true. Should I open a separate issue for that?

Yes, thanks for finding this problem. When `needStop` is true, the event is not sent to `s.ch` so `BinlogSyncer` should not ACK

> We want to ACK after writing to disk.
> We want the ability to retrieve the full GTID set that has been written to disk.

I still recommend you type cast the `event`s get from `event, err := streamer.GetEvent(ctx)` and get the GTID set from some events like `XIDEvent`. The `event` is really the stuff the caller has written to disk so the GTID set has the same meaning. The new getter does not have the same meaning considering `BinlogSyncer` has inner channel buffer, it's error-prone to have extra action to match the two GTID set.

dt8269: According to the MySQL documentation:

> The replica acknowledges receipt of a transaction's events only after they have been written to its relay log and flushed to disk.

Is achieving parity with the MySQL protocol a goal here? The asynchronous ordering of these two events in this library appears to violate that behavior.

lance6716: That's a requirement for MySQL relay log, not this package. If you want to implement the behaviour like MySQL relay log, more adjustment should be made.

dt8269: Ah, right okay. We have that requirement so I'm going to close this PR and see how we can implement that invariant

dt8269: thank you, by the way, for your attention and responsiveness

PR #919, opened 2024-09-27
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of 4 rather than initializing the length of this slice.


dveeden:
Can you explain the issue you're trying to fix? Is this for functionality and/or performance?

lance6716:
Hi, please fix ci

dveeden:
@cuishuang could you fix the failing checks?

cuishuang:
> @cuishuang could you fix the failing checks?

Sorry, it seems that my changes were incorrect. I found that my detection tool has a high false positive rate when the slice type is byte.

Sorry for wasting your time.

PR #920, opened 2024-09-30
Ensure ACKs are sent after backup

dt8269:
## Summary
This PR addresses an issue where acknowledgments (ACKs) were sometimes sent to the master before binlog events were fully written and fsynced to disk during backup operations. Sending ACKs prematurely in semi-synchronous replication could lead to data loss if the replica fails after sending the ACK but before persisting the event.
### ACK after backup
- Ensured that ACKs are sent only after the event has been fully processed and fsynced by sending the ACK *after* event handling
###  Event handling
- Introduced an `EventHandler` interface with a `HandleEvent` method for processing binlog events. This allows custom event handling logic to be injected into the replication stream
- Accordingly, added an `eventHandler` field and `SetEventHandler` method to `BinlogSyncer`.
- Implemented `BackupEventHandler` which writes binlog events to disk and ensures that each event is fsynced before returning. This ensures data durability before ACKs are sent.

### Separating parsing from handling
- Modified the `onStream` method in `BinlogSyncer` to separate event parsing (`parseEvent`) from event handling and ACK sending (`handleEventAndACK`). This adheres to the single-responsibility principle and makes the code cleaner
- Moved state updates (e.g., updating `b.nextPos`) and GTIDSet handling from `parseEvent` to `handleEventAndACK` to avoid side effects during parsing. Something called `parseEvent` should only be parsing, not modifying state or sending ACKs.

## Testing
Three new tests:
* `TestGTIDSetHandling` - This confirms the continued functionality of the existing GTIDSet logic after it was involved in a refactor
* `TestACKSentAfterFsync` - This confirms that ACKs are only send *after* the backup files are fsynced
* `TestBackupEventHandlerInvocation` -  This confirms that the (new) EventHandler works as expected

```
❯ go test
PASS
ok      github.com/go-mysql-org/go-mysql/replication    0.703s
```

### DIFF
--- a/replication/backup.go
+++ b/replication/backup.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"sync"
 	"time"
 
 	. "github.com/go-mysql-org/go-mysql/mysql"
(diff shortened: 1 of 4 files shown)


lance6716:
maybe we can limit the user to either use the new blocking handler or use the existing asynchronous streaming consumer. in your use case of persisting to disk, no need to bother draining the stream by GetEvent

to achieve this, don't send the event to the channel if handler is set. and refine other logic and comments.

Issue #866, opened 2024-05-07
Support MySQL 8.4.0


dveeden:
MySQL 8.4.0 has removed some of the deprecated "MASTER" and "SLAVE" terms.

```
--- FAIL: TestSyncerSuite (0.01s)
    --- FAIL: TestSyncerSuite/TestMysqlBinlogCodec (0.00s)
        replication_test.go:52: 
            	Error Trace:	/home/dvaneeden/dev/go-mysql/replication/replication_test.go:52
            	            				/home/dvaneeden/dev/go-mysql/replication/replication_test.go:409
            	Error:      	Received unexpected error:
            	            	ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'MASTER' at line 1
            	Test:       	TestSyncerSuite/TestMysqlBinlogCodec
    --- FAIL: TestSyncerSuite/TestMysqlPositionSync (0.00s)
        replication_test.go:308: 
            	Error Trace:	/home/dvaneeden/dev/go-mysql/replication/replication_test.go:308
            	            				/home/dvaneeden/dev/go-mysql/replication/replication_test.go:340
            	Error:      	Received unexpected error:
            	            	ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'MASTER STATUS' at line 1
            	Test:       	TestSyncerSuite/TestMysqlPositionSync
    --- FAIL: TestSyncerSuite/TestMysqlSemiPositionSync (0.00s)
        replication_test.go:308: 
            	Error Trace:	/home/dvaneeden/dev/go-mysql/replication/replication_test.go:308
            	            				/home/dvaneeden/dev/go-mysql/replication/replication_test.go:403
            	Error:      	Received unexpected error:
            	            	ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'MASTER STATUS' at line 1
            	Test:       	TestSyncerSuite/TestMysqlSemiPositionSync
    --- FAIL: TestSyncerSuite/TestStartBackupEndInGivenTime (0.00s)
        replication_test.go:52: 
            	Error Trace:	/home/dvaneeden/dev/go-mysql/replication/replication_test.go:52
            	            				/home/dvaneeden/dev/go-mysql/replication/backup_test.go:16
            	Error:      	Received unexpected error:
[...]

PR #933, opened 2024-10-26
dump: use --source-data based on version

dveeden:
Use `mysqldump --source-data ...` instead of `mysqldump --master-data ...` based on the version of `mysqldump`.

### Related

- #866


dveeden:
This would likely not work if `mariadb-dump` is used instead of `mysqldump`...

In the future MariaDB might remove `mysqldump`...

```
dvaneeden@dve-carbon:~$ ./opt/mysql/11.2.2/bin/mariadb-dump --help | head -1
./opt/mysql/11.2.2/bin/mariadb-dump from 11.2.2-MariaDB, client 10.19 for linux-systemd (x86_64)
dvaneeden@dve-carbon:~$ ./opt/mysql/11.2.2/bin/mysqldump --help | head -1
./opt/mysql/11.2.2/bin/mysqldump: Deprecated program name. It will be removed in a future release, use '/home/dvaneeden/opt/mysql/11.2.2/bin/mariadb-dump' instead
./opt/mysql/11.2.2/bin/mysqldump from 11.2.2-MariaDB, client 10.19 for linux-systemd (x86_64)
dvaneeden@dve-carbon:~$ 
```

lance6716 (approved):

PR #936, opened 2024-10-27
server: Improve example

dveeden:
1. Use the example from the `README.md` and turn it into a `go-mysqlserver` binary that users can run.
2. Add more logging to make it easier to understand what it is doing. This is done both in `go-mysqlserver` as well as the `EmptyHandler`
3. Remove the `server/example` as we already have a example now.
4. Support the minimal set of queries that MySQL Shell `mysqlsh` needs to be able to connect. (tested with MySQL Shell 9.1.0)
5. Change the default version from 5.7.0 to 8.0.11 (first 8.0 GA version)


server/command.go
```diff
@@ -186,39 +188,71 @@ type EmptyReplicationHandler struct {
...
+			return nil, err
+		} else {
+			return &mysql.Result{
+				Status:       0,
+				Warnings:     0,
+				InsertId:     0,
+				AffectedRows: 0,
+				Resultset:    r,
+			}, nil
+		}
```
lance6716, line 215:
```suggestion
		}
		return &mysql.Result{
			Status:       0,
			Warnings:     0,
			InsertId:     0,
			AffectedRows: 0,
			Resultset:    r,
		}, nil
```

server/command.go
```diff
@@ -186,39 +188,71 @@ type EmptyReplicationHandler struct {
...
+	}
+
 	return nil, fmt.Errorf("not supported now")
 }
 
 func (h EmptyHandler) HandleFieldList(table string, fieldWildcard string) ([]*Field, error) {
+	log.Println("Received: FieldList")
```
lance6716, line 222: It's helpful to print the function arguments? Also for below methods

lance6716 (commented):
rest lgtm

server/command.go
dveeden: Done.

lance6716 (approved):

PR #939, opened 2024-11-05
Add support for rowsEventDecodeFunc in Canal

cameron-p-m:


cameron-p-m:
Thank you, we deciding not to go this route, but I can open it back up later if its desired.

PR #943, opened 2024-11-08
canal: support ipv6 address

sinomoe:
canal: 
* address format support ipv6 with port like `[::1]:3306` in canal
* parse ipv6 address with optional port like `::1` or `[::1]:3306` in dumper

### DIFF
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -481,17 +481,25 @@ func (c *Canal) prepareSyncer() error {
 	if strings.Contains(c.cfg.Addr, "/") {
 		cfg.Host = c.cfg.Addr
 	} else {
-		seps := strings.Split(c.cfg.Addr, ":")
-		if len(seps) != 2 {
-			return errors.Errorf("invalid mysql addr format %s, must host:port", c.cfg.Addr)
+		ipv6 := strings.Count(c.cfg.Addr, ":") > 1
+		if ipv6 && !strings.ContainsAny(c.cfg.Addr, "[]") {
+			return errors.Errorf("invalid mysql ipv6 addr format %s, must [host]:port", c.cfg.Addr)
 		}
-
-		port, err := strconv.ParseUint(seps[1], 10, 16)
+		lastSep := strings.LastIndex(c.cfg.Addr, ":")
+		if !ipv6 && lastSep == -1 {
+			return errors.Errorf("invalid mysql ipv4 addr format %s, must host:port", c.cfg.Addr)
+		}
+		addr := strings.Trim(c.cfg.Addr[:lastSep], "[]")
+		ip := net.ParseIP(addr)
+		if ip == nil {
+			return errors.Errorf("invalid mysql ip format %s", addr)
+		}
+		port, err := strconv.ParseUint(c.cfg.Addr[lastSep+1:], 10, 16)
 		if err != nil {
 			return errors.Trace(err)
 		}
 
-		cfg.Host = seps[0]
+		cfg.Host = addr
 		cfg.Port = uint16(port)
 	}
 
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -481,26 +481,17 @@ func (c *Canal) prepareSyncer() error {
 	if strings.Contains(c.cfg.Addr, "/") {
 		cfg.Host = c.cfg.Addr
 	} else {
-		ipv6 := strings.Count(c.cfg.Addr, ":") > 1
-		if ipv6 && !strings.ContainsAny(c.cfg.Addr, "[]") {
-			return errors.Errorf("invalid mysql ipv6 addr format %s, must [host]:port", c.cfg.Addr)
+		host, port, err := net.SplitHostPort(c.cfg.Addr)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


canal/canal.go
```diff
@@ -481,17 +481,25 @@ func (c *Canal) prepareSyncer() error {
 	if strings.Contains(c.cfg.Addr, "/") {
 		cfg.Host = c.cfg.Addr
 	} else {
-		seps := strings.Split(c.cfg.Addr, ":")
```
lance6716, line 484: I think we can use https://pkg.go.dev/net#SplitHostPort

canal/canal.go
```diff
@@ -481,17 +481,25 @@ func (c *Canal) prepareSyncer() error {
...
-			return errors.Errorf("invalid mysql addr format %s, must host:port", c.cfg.Addr)
+		ipv6 := strings.Count(c.cfg.Addr, ":") > 1
+		if ipv6 && !strings.ContainsAny(c.cfg.Addr, "[]") {
+			return errors.Errorf("invalid mysql ipv6 addr format %s, must [host]:port", c.cfg.Addr)
 		}
-
-		port, err := strconv.ParseUint(seps[1], 10, 16)
+		lastSep := strings.LastIndex(c.cfg.Addr, ":")
+		if !ipv6 && lastSep == -1 {
+			return errors.Errorf("invalid mysql ipv4 addr format %s, must host:port", c.cfg.Addr)
```
dveeden, line 490:
```suggestion
			return errors.Errorf("invalid MySQL IPv4 addr format %s, must host:port", c.cfg.Addr)
```

canal/canal.go
```diff
@@ -481,17 +481,25 @@ func (c *Canal) prepareSyncer() error {
...
+		}
+		addr := strings.Trim(c.cfg.Addr[:lastSep], "[]")
+		ip := net.ParseIP(addr)
+		if ip == nil {
+			return errors.Errorf("invalid mysql ip format %s", addr)
```
dveeden, line 495:
```suggestion
			return errors.Errorf("invalid MySQL IP-address format %s", addr)
```

dveeden (commented):
Should we try to support things like `:3307` (e.g. only a port, not a host?)

canal/canal.go
sinomoe: Good idea, let me check this

canal/canal.go
```diff
@@ -481,17 +481,25 @@ func (c *Canal) prepareSyncer() error {
 	if strings.Contains(c.cfg.Addr, "/") {
 		cfg.Host = c.cfg.Addr
 	} else {
-		seps := strings.Split(c.cfg.Addr, ":")
-		if len(seps) != 2 {
-			return errors.Errorf("invalid mysql addr format %s, must host:port", c.cfg.Addr)
+		ipv6 := strings.Count(c.cfg.Addr, ":") > 1
+		if ipv6 && !strings.ContainsAny(c.cfg.Addr, "[]") {
+			return errors.Errorf("invalid mysql ipv6 addr format %s, must [host]:port", c.cfg.Addr)
```
dveeden, line 486:
```suggestion
			return errors.Errorf("invalid MySQL IPv6 address format %s, must [host]:port", c.cfg.Addr)
```

canal/canal.go
sinomoe: done

mysql/const.go
```diff
@@ -172,6 +172,7 @@ const (
 
 const (
 	DEFAULT_ADDR                  = "127.0.0.1:3306"
+	DEFAULT_IPv6_ADDR             = "[::1]:3306"
```
dveeden, line 175:
```suggestion
	DEFAULT_IPV4_ADDR                  = "127.0.0.1:3306"
	DEFAULT_IPV6_ADDR             = "[::1]:3306"
```

Not sure if we should uppercase the `V` here or not..

sinomoe: Changing `DEFAULT_ADDR` to `DEFAULT_IPv6_ADDR` might break compatibility, maybe it's better to leave it unchanged.

sinomoe:
> What about tests and CI? Anything we can do to test this and/or make it easier to test?

Test cases and CI have been completed.

dump/dumper.go
```diff
@@ -202,10 +202,25 @@ func (d *Dumper) Dump(w io.Writer) error {
 	if strings.Contains(d.Addr, "/") {
 		args = append(args, fmt.Sprintf("--socket=%s", d.Addr))
 	} else {
-		seps := strings.SplitN(d.Addr, ":", 2)
-		args = append(args, fmt.Sprintf("--host=%s", seps[0]))
-		if len(seps) > 1 {
-			args = append(args, fmt.Sprintf("--port=%s", seps[1]))
+		ipv6 := strings.Count(d.Addr, ":") > 1
```
lance6716, line 205: this part can also use `net.SplitHostPort`, just like canal/canal.go

lance6716 (commented):
rest lgtm

dump/dumper.go
sinomoe: done

lance6716 (approved):
ptal @dveeden

dveeden (approved):

PR #950, opened 2024-11-20
Update README

dveeden:
- More clearly state that this should also work for MariaDB
- Move migration info to the end of the README as there are probably not that many people anymore that need this
- Add more details to the "Content" section
- List examples
- Add a `Go Reference` badge

Open questions:
- Should we update the "Donate" and "Feedback" sections?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,33 @@
 # go-mysql
 
-A pure go library to handle MySQL network protocol and replication.
+A pure go library to handle MySQL network protocol and replication as used by MySQL and MariaDB.
 
 ![semver](https://img.shields.io/github/v/tag/go-mysql-org/go-mysql)
 ![example workflow](https://github.com/go-mysql-org/go-mysql/actions/workflows/ci.yml/badge.svg)
 ![gomod version](https://img.shields.io/github/go-mod/go-version/go-mysql-org/go-mysql/master)
-
-## How to migrate to this repo
-To change the used package in your repo it's enough to add this `replace` directive to your `go.mod`:
-```
-replace github.com/siddontang/go-mysql => github.com/go-mysql-org/go-mysql v1.10.0
-```
-
-v1.10.0 - is the last tag in repo, feel free to choose what you want.
+[![Go Reference](https://pkg.go.dev/badge/github.com/go-mysql-org/go-mysql.svg)](https://pkg.go.dev/github.com/go-mysql-org/go-mysql)
 
 ## Changelog
 This repo uses [Changelog](CHANGELOG.md).
 
 ---
 # Content
-* [Replication](#replication)
-* [Incremental dumping](#canal)
-* [Client](#client)
-* [Fake server](#server)
-* [database/sql like driver](#driver)
+* [Replication](#replication) - Process events from a binlog stream.
+* [Incremental dumping](#canal) - Sync from MySQL to Redis, Elasticsearch, etc.
+* [Client](#client) - Simple MySQL client.
+* [Fake server](#server) - server side of the MySQL protocol, as library.
+* [database/sql like driver](#driver) - An alternative `database/sql` driver for MySQL.
+* [Migration](#how-to-migrate-to-this-repo) - Information for how to migrate if you used the old location of this project.
+
+## Examples
+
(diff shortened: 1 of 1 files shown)


lance6716:
> Should we update the "Donate" and "Feedback" sections?

I think @siddontang does not need donation from this project now 😄 We can setup our donation link but I'm not familiar with it. For feedback I guess most users of this package are golang developers and they can open GitHub issues without the feedback guide

> Should the examples be listed next to the sections in the README or combined in a single section

The simple program under `/cmd` should also be helpful to new developers, I think we can only leave very short code snippet in README and link to `/cmd` for a more detailed usage. I'll think about this part later.

dveeden:
@lance6716 GitHub offers this: https://github.com/sponsors but not sure if that's only for individuals or also for organizations. I'm not sure we need this.

lance6716:
> @lance6716 GitHub offers this: https://github.com/sponsors but not sure if that's only for individuals or also for organizations. I'm not sure we need this.

I checked that organization can have sponsors, but it requires an email as contact and a bank account 😂 It seems more suitable for a real organization rather than our virtual team. I'm not capable to handle these finance stuffs, maybe you or other organization members can handle this or we don't need sponsors for now.

I'll review README changes soon.

README.md
```diff
@@ -1,29 +1,33 @@
...
 
 ---
 # Content
-* [Replication](#replication)
-* [Incremental dumping](#canal)
-* [Client](#client)
-* [Fake server](#server)
-* [database/sql like driver](#driver)
+* [Replication](#replication) - Process events from a binlog stream.
+* [Incremental dumping](#canal) - Sync from MySQL to Redis, Elasticsearch, etc.
```
lance6716, line 16: IMO the difference between this "Incremental dumping" and "Replication" item are developers need to implement interfaces (like `EventHandler`) or handle decoded binlog events. Maybe use "library mode" or "framework mode". I choose these two words to tell the difference like 

> When you use a library, you are in charge of the flow of the application. You are choosing when and where to call the library. When you use a framework, the framework is in charge of the flow. It provides some places for you to plug in your code, but it calls the code you plugged in as needed.

https://www.freecodecamp.org/news/the-difference-between-a-framework-and-a-library-bd133054023f/ (a random blog ranks top in google search "library vs framework")

README.md
```diff
@@ -1,29 +1,33 @@
...
+* [Fake server](#server) - server side of the MySQL protocol, as library.
+* [database/sql like driver](#driver) - An alternative `database/sql` driver for MySQL.
+* [Migration](#how-to-migrate-to-this-repo) - Information for how to migrate if you used the old location of this project.
+
+## Examples
+
+The `cmd` directory contains example applications that can be build by running `make build` in the root of the project. The resulting binaries will be places in `bin/`.
+
+- `go-binlogparser`: parses a binlog file at a given offset
+- `go-canal`: streams binlog events from a server to canal
```
lance6716, line 27: In fact I don't know what's `canal` 😂  maybe it's https://github.com/alibaba/canal?tab=readme-ov-file ? I'm not sure this library will output similar output or have similar usage as that one.

lance6716 (commented):
rest lgtm

README.md
dveeden: Not sure what you want me to change. Do you you want to clarify the difference between "Replication" and "Incremental dumping"?

README.md
dveeden: Looks like it might be similar, but I'm not sure it is the same or if it is compatible.

README.md
lance6716: yes, like

```
* [Use as a library](#replication) - Process events from a binlog stream and have maximum functionalities.
* [Use as a framework](#canal) - Implement `EventHandler` interface to subscribe binlog changes
```

lance6716: However I don't have confidence how to describe these two items. Will think about it later

lance6716:
Sorry I'm too busy this week and don't have enough time to check my comments. Because this PR definitely is an improvement, I think we can merge it. The unresolved comment can be fixed in future PR

lance6716 (approved):

Issue #845, opened 2024-01-18
MySQL 8.3 will introduce new GTID format


lance6716:
```
The new GTID format is UUID:<TAG>:NUMBER, in which <TAG> is an arbitrary string up to 8 characters long
```

https://dev.mysql.com/doc/relnotes/mysql/8.3/en/news-8-3-0.html

thanks for @dveeden to remind us

- [x] Support tagged gtid in `PreviousGTIDsEvent` #952 
- [x] Decode `GTID_TAGGED_LOG_EVENT`
- [ ] Support tagged GTID in `mysql.ParseGTIDSet(mysql.MySQLFlavor, ...)` / `mysql.ParseMysqlGTIDSet()`

dveeden:
An example event from `mysqlbinlog --hexdump`
```
# at 220825
#241112 21:51:23 server id 1  end_log_pos 220908 CRC32 0x71013dac 
# Position  Timestamp   Type   Source ID        Size      Source Pos    Flags 
# 00035e99 cb bf 33 67   2a   01 00 00 00   53 00 00 00   ec 5e 03 00   00 00
# 00035eac 02 78 00 00 00 02 aa aa  aa aa 88 88 66 66 44 44 |.x..........ffDD|
# 00035ebc 22 22 22 22 22 22 04 73  20 1b 06 14 73 65 63 6f |.......s....seco|
# 00035ecc 6e 64 74 65 73 74 08 c1  0e 0a d1 0e 0c 7f 23 35 |ndtest.........5|
# 00035edc 48 61 bd 26 06 10 59 04  12 a3 ff 0a ac 3d 01 71 |Ha....Y........q|
# 	GTID	last_committed=472	sequence_number=474	rbr_only=yes	original_committed_timestamp=1731444683060515	immediate_commit_timestamp=1731444683060515	transaction_length=278
/*!50718 SET TRANSACTION ISOLATION LEVEL READ COMMITTED*//*!*/;
# original_commit_timestamp=1731444683060515 (2024-11-12 21:51:23.060515 CET)
# immediate_commit_timestamp=1731444683060515 (2024-11-12 21:51:23.060515 CET)
/*!80001 SET @@session.original_commit_timestamp=1731444683060515*//*!*/;
/*!80014 SET @@session.original_server_version=90100*//*!*/;
/*!80014 SET @@session.immediate_server_version=90100*//*!*/;
[...]

PR #952, opened 2024-11-21
replication: Support GTID tag in PreviousGTIDsEvent

dveeden:
Issue: ref #845

The `PreviousGTIDsEvent` / `PREVIOUS_GTIDS_LOG_EVENT` has changed to work with tagged GTIDs.

First the `uuidCount` has changed, it encodes the GTID format. Here format 1 is tagged and format 0 is untagged.

Then each entry may have a tag. If there is a tag then the uuid itself isn't printed but the tag is appended to the last entry.

## Example 1
`896e7882-18fe-11ef-ab88-22222d34d411:1-3`

regular format, compatible with both formats

## Example 2
`896e7882-18fe-11ef-ab88-22222d34d411:1-4:aaaa:1`

tagged format.

Combination of 
- `896e7882-18fe-11ef-ab88-22222d34d411:1-4`
- `896e7882-18fe-11ef-ab88-22222d34d411:aaaa:1`

## Example 3

`896e7882-18fe-11ef-ab88-22222d34d411:1-4:aaaa:1:abc:1-3:bbbbb:1:bbbbbb:1:x:1,896e7882-18fe-11ef-ab88-22222d34d412:1-2`

Combination of:
```
896e7882-18fe-11ef-ab88-22222d34d411:1-4
                                    :aaaa:1
                                    :abc:1-3
                                    :bbbbb:1
                                    :bbbbbb:1
                                    :x:1,
896e7882-18fe-11ef-ab88-22222d34d412:1-2
```

Please also see: `mysqlbinlog --read-from-remote-server --hexdump $binlogfile` to see how MySQL encodes/decodes this.

See also:
- https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html


lance6716 (commented):
I'm not sure Example 3 in PR description has the same format as MySQL. in https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html , there's

> GTIDs originating from the same server but having different tags are treated in a manner similar to those originating from different servers, like this:

> 3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:15-21, 3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_2:8-52

rather than `3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:15-21:Domain_2:8-52`. I guess the reason is if `Domain_2` is replaced by tag `23`, and we will see `3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:15-21:23:8-52`. which is ambiguous.

replication/event.go
```diff
@@ -229,15 +229,59 @@ type PreviousGTIDsEvent struct {
...
 }
 
+type GtidFormat int
+
+const (
+	GtidFormatClassic = iota
+	GtidFormatTagged
+)
+
+func decodeSid(data []byte) (format GtidFormat, sidnr uint64) {
```
lance6716, line 239: can you put the referenced MySQL source code or document in comment, so other reviewers can double check the implementation?

dveeden:
And with the exact GTIDs from the docs:
```
mysql-9.1.0> reset binary logs and gtids;
Query OK, 0 rows affected (0.03 sec)

mysql-9.1.0> set global gtid_purged='3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:15-21, 
    '> 3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_2:8-52';
Query OK, 0 rows affected (0.02 sec)

mysql-9.1.0> show global variables like 'gtid%';
+----------------------------------+-----------------------------------------------------------------------+
| Variable_name                    | Value                                                                 |
+----------------------------------+-----------------------------------------------------------------------+
| gtid_executed                    | 3e11fa47-71ca-11e1-9e33-c80aa9429562:domain_1:1-3:15-21:domain_2:8-52 |
| gtid_executed_compression_period | 0                                                                     |
| gtid_mode                        | ON                                                                    |
| gtid_owned                       |                                                                       |
| gtid_purged                      | 3e11fa47-71ca-11e1-9e33-c80aa9429562:domain_1:1-3:15-21:domain_2:8-52 |
+----------------------------------+-----------------------------------------------------------------------+
5 rows in set (0.00 sec)
```

dveeden:
I've filed https://bugs.mysql.com/bug.php?id=116789 for this

dveeden:
I also filed https://bugs.mysql.com/bug.php?id=116747 some days ago

replication/event.go
```diff
@@ -229,15 +229,59 @@ type PreviousGTIDsEvent struct {
...
+	sidnr = binary.LittleEndian.Uint64(data)
+	return
+}
+
 func (e *PreviousGTIDsEvent) Decode(data []byte) error {
 	pos := 0
-	uuidCount := binary.LittleEndian.Uint16(data[pos : pos+8])
+
+	gtidinfo := make([]byte, 8)
+	copy(gtidinfo, data[:8])
```
lance6716, line 265: I see that the new copy is caused by `decodeSid` will modify data in-place. How about let `decodeSid` copy the needed data? So there's no risk that in future we need to change the code and caller forget to copy it.

replication/event.go
```diff
@@ -229,15 +229,59 @@ type PreviousGTIDsEvent struct {
...
+	format, uuidCount := decodeSid(gtidinfo)
 	pos += 8
 
 	previousGTIDSets := make([]string, uuidCount)
-	for i := range previousGTIDSets {
+	currentSetnr := 0
+	for _ = range previousGTIDSets {
```
lance6716, line 271:
```suggestion
	for range previousGTIDSets {
```

replication/event.go
```diff
@@ -229,15 +229,59 @@ type PreviousGTIDsEvent struct {
...
 		uuid := e.decodeUuid(data[pos : pos+16])
 		pos += 16
+		isTag := false
+		var tag string
+		if format == GtidFormatTagged {
+			tagLength := int(data[pos]) / 2
+			pos += 1
+			if tagLength > 0 {
```
lance6716, line 279: can the tag have zero length? please add comments

replication/event.go
```diff
@@ -254,9 +298,14 @@ func (e *PreviousGTIDsEvent) Decode(data []byte) error {
 			}
 			intervals[i] = interval
 		}
-		previousGTIDSets[i] = fmt.Sprintf("%s:%s", uuid, strings.Join(intervals, ":"))
+		if isTag {
+			previousGTIDSets[currentSetnr-1] += fmt.Sprintf(":%s:%s", tag, strings.Join(intervals, ":"))
```
lance6716, line 302: Because string is immutable, string `+=` will create a new string object each time. If there are too many tags the performance will drop. Maybe store every "tag+intervals" and join them once at all

lance6716 (commented):
Just a pure code-related review. I'll check GTID encoding and decoding soon

replication/event.go
dveeden:
```
mysql-9.1.0> set gtid_next = 'AUTOMATIC:abc';
Query OK, 0 rows affected (0.00 sec)

mysql-9.1.0> create table test.t9 (id int primary key);
Query OK, 0 rows affected (0.02 sec)

mysql-9.1.0> set gtid_next = 'AUTOMATIC:';
ERROR 1774 (HY000): Malformed GTID specification 'AUTOMATIC:'.
```

https://dev.mysql.com/doc/dev/mysql-server/latest/classmysql_1_1gtid_1_1Tag.html#details

> Representation of the GTID tag.
>
> Tag format: [a-z_]{a-z0-9_}{0,31} Tag may be created from a given text. Text may contain leading and trailing spaces which will be omitted during Tag creation. Text may also contain uppercase characters. Acceptable format for text is as follows: [:space:][a-zA-Z_][a-zA-Z0-9_]{0,31}[:space:]

So it is non-zero if it has a tag. Note the `{0,31}` in the regexp, which is only for the second character as the first can't be a number.

replication/event.go
dveeden: yes good one.

replication/event.go
dveeden: done

dveeden:
This could also be useful: https://databaseblog.myname.nl/2024/12/mysql-gtid-tags-and-binlog-events.html

replication/event.go
```diff
@@ -229,15 +229,69 @@ type PreviousGTIDsEvent struct {
...
 	previousGTIDSets := make([]string, uuidCount)
-	for i := range previousGTIDSets {
+	currentSetnr := 0
+	for range previousGTIDSets {
 		uuid := e.decodeUuid(data[pos : pos+16])
 		pos += 16
+		isTag := false
+		var tag string
+		if format == GtidFormatTagged {
+			tagLength := int(data[pos]) / 2
```
lance6716, line 287: I'm not sure how this `... / 2` happens in MySQL's source code. Seems it's https://github.com/mysql/mysql-server/blob/61a3a1d8ef15512396b4c2af46e922a19bf2b174/libs/mysql/gtid/tag.cpp#L84 and it further calls a complex `read_varlen_bytes`.

Just to double check with you, did you simplify `read_varlen_bytes` into `int(data[pos]) / 2`? If not, I think I can help to provide more test data with different tag length and check our calculation is correct

replication/event.go
```diff
@@ -229,15 +229,69 @@ type PreviousGTIDsEvent struct {
...
+	format, uuidCount := decodeSid(data)
 	pos += 8
 
 	previousGTIDSets := make([]string, uuidCount)
-	for i := range previousGTIDSets {
+	currentSetnr := 0
+	for range previousGTIDSets {
 		uuid := e.decodeUuid(data[pos : pos+16])
 		pos += 16
+		isTag := false
```
lance6716, line 284: seems `isTag` can be replaced by `len(tag) > 0`

replication/event.go
lance6716: maybe try this

```diff
diff --git a/replication/event.go b/replication/event.go
index 20c7b0c..0786516 100644
--- a/replication/event.go
+++ b/replication/event.go
@@ -277,7 +277,9 @@ func (e *PreviousGTIDsEvent) Decode(data []byte) error {
 	pos += 8
 
 	previousGTIDSets := make([]string, uuidCount)
+
 	currentSetnr := 0
+	var buf strings.Builder
 	for range previousGTIDSets {
 		uuid := e.decodeUuid(data[pos : pos+16])
 		pos += 16
@@ -292,30 +294,38 @@ func (e *PreviousGTIDsEvent) Decode(data []byte) error {
 				pos += tagLength
 			}
 		}
+
+		if isTag {
+			buf.WriteString(":")
+			buf.WriteString(tag)
+		} else {
+			if currentSetnr != 0 {
+				buf.WriteString(",")
+			}
+			buf.WriteString(uuid)
+			currentSetnr += 1
+		}
+
 		sliceCount := binary.LittleEndian.Uint16(data[pos : pos+8])
 		pos += 8
-		intervals := make([]string, sliceCount)
-		for i := range intervals {
+		for range sliceCount {
+			buf.WriteString(":")
+
 			start := e.decodeInterval(data[pos : pos+8])
 			pos += 8
 			stop := e.decodeInterval(data[pos : pos+8])
 			pos += 8
-			interval := ""
 			if stop == start+1 {
-				interval = fmt.Sprintf("%d", start)
+				fmt.Fprintf(&buf, "%d", start)
 			} else {
-				interval = fmt.Sprintf("%d-%d", start, stop-1)
+				fmt.Fprintf(&buf, "%d-%d", start, stop-1)
 			}
-			intervals[i] = interval
 		}
-		if isTag {
-			previousGTIDSets[currentSetnr-1] += fmt.Sprintf(":%s:%s", tag, strings.Join(intervals, ":"))
-		} else {
-			previousGTIDSets[currentSetnr] = fmt.Sprintf("%s:%s", uuid, strings.Join(intervals, ":"))
+		if !isTag {
 			currentSetnr += 1
 		}
 	}
-	e.GTIDSets = strings.Join(previousGTIDSets[:currentSetnr], ",")
+	e.GTIDSets = buf.String()
 	return nil
 }
 
```

lance6716 (commented):
rest lgtm

lance6716:
I can help to modify this PR with my comment. Do you want me to do that @dveeden

dveeden:
> I can help to modify this PR with my comment. Do you want me to do that @dveeden

Yes

replication/event.go
lance6716: I'll check this part later

PR #961, opened 2024-12-27
use syscall.Gettimeofday to get current time

hjweddie:
If we use syscall.Gettimeofday instead of time.Now, we could get a bit performance increasement.

Here are benchmark code and some result:

```source code
package main_test

import (
	"syscall"
	"testing"
	"time"
)

func Now() time.Time {
	var tv syscall.Timeval
	if err := syscall.Gettimeofday(&tv); nil != err {
		return time.Now()
	}

	return time.Unix(0, syscall.TimevalToNsec(tv))
}

func BenchmarkOld(t *testing.B) {
	t.ResetTimer()
	for n := 0; n < t.N; n++ {
		_ = time.Now()
	}
	t.StopTimer()
}

func BenchmarkNew(t *testing.B) {
	t.ResetTimer()
	for n := 0; n < t.N; n++ {
		_ = Now()
	}
	t.StopTimer()
}
```

```benchmark result
goos: linux
goarch: amd64
pkg: test
cpu: Intel(R) Xeon(R) Platinum 8575C
BenchmarkOld-4   	22126742	        54.39 ns/op
BenchmarkNew-4   	37930492	        31.78 ns/op
PASS
ok  	test	2.500s


goos: darwin
goarch: arm64
pkg: test
cpu: Apple M2
BenchmarkOld-8   	20453368	        54.23 ns/op
BenchmarkNew-8   	34531131	        34.67 ns/op
PASS
ok  	test	5.448s
```


dveeden (commented):
This might be useful background info: 
- https://tpaschalis.me/golang-time-now/
- https://pkg.go.dev/syscall#Gettimeofday
- https://man7.org/linux/man-pages/man2/gettimeofday.2.html

I think `utils.Now()` should probably have some tests.

dveeden:
What I don't really know is if `time.Now()` and `syscall.Gettimeofday()` behave the same for:
- Timezones (With system clock set to UTC or not, etc)
- Precision
- Monotonic increases
- etc

hjweddie:
According to https://tpaschalis.me/golang-time-now/, for linux amd64, when we call time.Now(), it uses runtime.vdsoClockgettimeSym, which makes a system call with context-switch overhead. 
When we call syscall.Gettimeofday, it is called directly by user application, acceing the memory without context-switch overhead. This is why syscall.Gettimeofday may be faster than time.Now in linux.

But for freebsd, https://github.com/golang/go/blob/master/src/runtime/vdso_freebsd.go#L108 , time.Now prefers to use vdsoClockgettimeSym, and the same as linux.  
When we call syscall.Gettimeofday, because theris no vdsoGettimeofdaySym in freebsd vdso. I think it just make a system call Gettimeofday without vdso, which has more interrupt handing overhead than vdsoGettimeofdaySym. This is why now.Now is faster than syscall.Gettimeofday for freebsd.

hjweddie:
> What I don't really know is if `time.Now()` and `syscall.Gettimeofday()` behave the same for:
> 
> * Timezones (With system clock set to UTC or not, etc)
> * Precision
> * Monotonic increases
> * etc

I agree with you that we should probably have some tests. 
As what we had found out, we could gain performance increasement in linux but not all os, we may focus on linux os first and just use syscall.Gettimeofday on linux os.

dveeden:
This also fixes support for Windows:
```
dvaneeden@dve-carbon:~/dev/go-mysql$ git status
On branch hjweddie/master
nothing to commit, working tree clean
dvaneeden@dve-carbon:~/dev/go-mysql$ make GOOS=windows build
go build -o bin/go-mysqlbinlog cmd/go-mysqlbinlog/main.go
go build -o bin/go-mysqldump cmd/go-mysqldump/main.go
go build -o bin/go-canal cmd/go-canal/main.go
go build -o bin/go-binlogparser cmd/go-binlogparser/main.go
go build -o bin/go-mysqlserver cmd/go-mysqlserver/main.go
dvaneeden@dve-carbon:~/dev/go-mysql$ git checkout HEAD^1
HEAD is now at 45f2ba0 use syscall.Gettimeofday to get current time
dvaneeden@dve-carbon:~/dev/go-mysql$ git status
HEAD detached at 45f2ba0
nothing to commit, working tree clean
dvaneeden@dve-carbon:~/dev/go-mysql$ make GOOS=windows build
go build -o bin/go-mysqlbinlog cmd/go-mysqlbinlog/main.go
# github.com/go-mysql-org/go-mysql/utils
utils/time.go:16:30: undefined: syscall.TimevalToNsec
make: *** [Makefile:8: build] Error 1
dvaneeden@dve-carbon:~/dev/go-mysql$ 
```

utils/now.go
dveeden, line 1: I would probably call this `now_unix.go` and the other one `now.go` or `now_generic.go` or so.

hjweddie: Good idea and I rename files name to now_unix.go and now.go

hjweddie:
> What I don't really know is if `time.Now()` and `syscall.Gettimeofday()` behave the same for:
> 
> * Timezones (With system clock set to UTC or not, etc)
> * Precision
> * Monotonic increases
> * etc



> What I don't really know is if `time.Now()` and `syscall.Gettimeofday()` behave the same for:
> 
> * Timezones (With system clock set to UTC or not, etc)
> * Precision
> * Monotonic increases
> * etc

According to man doc: https://man7.org/linux/man-pages/man2/gettimeofday.2.html
Gettimeofday percision is one millisecond, and I can confirm it in unittest code.
I think we should use time.Now with seeding to get a more random int. 
For the most use cases in go-mysql project millisecond precision may be enough.

utils/now_unix.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"syscall"
+	"time"
+)
+
+// Now is a faster method to get current time
+func Now() time.Time {
+	var tv syscall.Timeval
+	if err := syscall.Gettimeofday(&tv); nil != err {
```
lance6716, line 13:
```suggestion
	if err := syscall.Gettimeofday(&tv); err != nil {
```

nit. just a style preference. golang will not have `if err = nil {}` problems

utils/now_unix_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestCustomTimeNow(t *testing.T) {
+	precision := time.Millisecond
+
+	for i := 0; i < 1000; i++ {
+		timestamp := time.Now().UnixNano()
+		customTimestamp := Now().UnixNano()
+
+		// two timestamp should within 1 percistion
+		assert.Equal(t, timestamp+int64(precision) >= customTimestamp && timestamp-int64(precision) <= customTimestamp, true, fmt.Sprintf("Loop: %d: customTimestamp should within %s. timestamp: %d, customTimestamp: %d", i, precision.String(), timestamp, customTimestamp))
```
lance6716, line 22: you can use `assert.WithinRange` or `assert.InDelta`

utils/now_unix_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+		os.Setenv("TZ", fmt.Sprintf("UTC%d", 14-i%27))
```
lance6716, line 24: please add a comment for its purpose. I don't clearly understand

lance6716 (commented):
rest lgtm

dveeden (approved):

lance6716 (approved):

PR #962, opened 2024-12-29
Remove duplicate SIGINT in signal handling

soya111:
1. Remove duplicate SIGINT from signal.Notify call
2. Keep other signal handlers (SIGHUP, SIGTERM, SIGQUIT) unchanged


lance6716 (commented):
lgtm

PR #967, opened 2025-01-10
replication: Improve COM_REGISTER_SLAVE

### LINKED ISSUE #966: Why does binlog syncer need to register as a slave?
Why does it need to register as a slave? [Code ref](https://github.com/go-mysql-org/go-mysql/blob/dc262fec458419fc58a4dc0d4cc2785383c1038d/replication/binlogsyncer.go#L593)

By making this additional network call, we are running into a MySQL password limitation as outlined [here](https://dev.mysql.com/doc/refman/8.0/en/change-master-to.html#:~:text=The%20password%20for%20the%20replication%20user%20account%20to%20use%20for%20connecting%20to%20the%20replication%20source%20server.%20The%20maximum%20length%20of%20the%20string%20value%20is%2032%20characters.%20If%20you%20specify%20MASTER_PASSWORD%2C%20MASTER_USER%20is%20also%20required).

* This has been flagged by the community before in the past. [Example](https://github.com/github/gh-ost/issues/484#issuecomment-326771023)

* [mysql-binlog-connector-java](https://github.com/shyiko/mysql-binlog-connector-java) provides a similar capability and does not invoke `COM_REGISTER_SLAVE`

dveeden:
Closes #966


replication/binlogsyncer.go
```diff
@@ -616,10 +616,8 @@ func (b *BinlogSyncer) writeRegisterSlaveCommand() error {
 	n = copy(data[pos:], b.cfg.User)
 	pos += n
 
-	data[pos] = uint8(len(b.cfg.Password))
+	data[pos] = uint8(0)
 	pos++
-	n = copy(data[pos:], b.cfg.Password)
```
lance6716, line 621: 😂 Don't know why we bother to send password before. Does different version of server has different behaviour?

Will take a look soon

Tang8330: Oh interesting, we don't need to send the password?

dveeden: The password is *only* for use in the output of `SHOW REPLICAS` if `--show-replica-auth-info` is enabled. This makes it a bit of a security hazard and with the length limitation also quite useless.

dveeden:
@lance6716 not sure how to best test this, could you help with that?

1. Create a user with a password that is >32 chars long.
2. Run a binlog syncer
3. Run `SHOW REPLICAS`, expect one row with `Replica_UUID` to be a valid UUID

lance6716 (approved):

lance6716:
> Create a user with a password that is >32 chars long.
> Run a binlog syncer
> Run SHOW REPLICAS, expect one row with Replica_UUID to be a valid UUID

I have checked it. Merging

Issue #964, opened 2025-01-04
#BUG:  sql.Execute  get rows count will panic nil error


DoMySon:
```go

var result, _ = db.Execute(`insert into (x,y,z) values(1,2,3)`)

var columns = result.Columns()  // there result != nil
var rows  = result.Rows()       // there result == nil (panic)



```

dveeden:
Is this with `github.com/go-mysql-org/go-mysql/driver` or with `github.com/go-mysql-org/go-mysql/client`?

What version?

Could you give a full example that I can run to reproduce this?

Is there a stack ?

DoMySon:
it's `github.com/go-mysql-org/go-mysql/client`, version is ` v1.10.0`
i just execute an insert operation.
![image](https://github.com/user-attachments/assets/d20330f9-ea08-4a2d-b7f6-bda757e2f864)

DoMySon:
first `result` not nil, but next operation always nil then panic

dveeden:
I can reproduce this with this code:
```go
package main

import (
	"fmt"

	"github.com/go-mysql-org/go-mysql/client"
	_ "github.com/go-mysql-org/go-mysql/client"
)

func main() {
	fd, err := client.Connect("127.0.0.1:3306", "root", "", "test")
	if err != nil {
		panic(err)
	}
	fmt.Printf("fd = %#v\n", fd)

	fd.Execute(`drop table if exists t`)
	fd.Execute(`create table t(x int, y int, z int, primary key (x,y,z))`)

	result, err := fd.Execute(`insert into t(x,y,z) values(1,2,3)`)
	if err != nil {
		panic(err)
	}
	fmt.Printf("result = %#v\n", result)

	rows := result.RowNumber()
	fmt.Printf("rows = %#v\n", rows)

	columns := result.ColumnNumber()
	fmt.Printf("columns = %#v\n", columns)
}
```

Result:
```
$ go run main.go 
[...]

PR #968, opened 2025-01-10
mysql: Fix ColumnNumber() panic with nil Resultset

dveeden:


lance6716 (approved):
This is definitely correct, so I'm going to merge it. However I can't understand if #964 has the same root cause, so I remove the "Closes #964" in PR description

PR #969, opened 2025-01-13
*: don't use zero value of mysql.Result

### LINKED ISSUE #964: #BUG:  sql.Execute  get rows count will panic nil error

lance6716:
```
// Result should be created by NewResultWithoutRows or NewResult. The zero value
// of Result is invalid.
```

closes #964


lance6716:
@dveeden I have fixed CI, PTAL

dveeden (approved):

PR #979, opened 2025-02-04
Use 4-byte UTF-8 by default

dveeden:


lance6716 (approved):
also change or remove line 278 of client/auth.go

PR #981, opened 2025-02-07
README: add Platform Support section

lance6716:
preview https://github.com/go-mysql-org/go-mysql/blob/3406f1ec2f3bb90e87f49f5b645fabd949d8afaa/README.md


dveeden (approved):
Maybe add a link to https://go.dev/wiki/MinimumRequirements and say that support is limited to OS versions supported by Go. (Only supported Go versions, following the Release Policy on https://go.dev/doc/devel/release ).

Maybe say that this should be able to work on other operating systems, but that this hasn't been tested.

We could do some `GOOS=windows go build ...` etc in CI to test for incompatibilities. Same for `GOARCH=riscv64` etc.

lance6716:
@dveeden updated. please check again

dveeden (approved):

Issue #915, opened 2024-09-09
Crashed in bindStmtArgs(), when using the Server package to forward Execute statements from sysbench


rfyim:
I have implemented a proxy for forwarding queries using the Server package. The ServerHandler is implemented as follows:
```go
type ServerHandler struct {
	Conn  *client.Conn
}

func NewServerHandler(cfg Config) (*ServerHandler, error) {
	conn, err := client.Connect(fmt.Sprintf("%s:%d", cfg.Host, cfg.Port), cfg.User, cfg.Password, "")
	if err != nil {
		return nil, err
	}
	if err = conn.Ping(); err != nil {
		return nil, err
	}
	return &ServerHandler{
		Conn: conn,
	}, nil
}

func (h ServerHandler) UseDB(dbName string) error {
	return h.Conn.UseDB(dbName)
}

func (h ServerHandler) HandleQuery(query string) (*mysql.Result, error) {
	return h.Conn.Execute(query)
}

func (h ServerHandler) HandleFieldList(table string, fieldWildcard string) ([]*mysql.Field, error) {
	return h.Conn.FieldList(table, fieldWildcard)
}
func (h ServerHandler) HandleStmtPrepare(query string) (int, int, interface{}, error) {
	stmt, err := h.Conn.Prepare(query)
	if err != nil {
		return 0, 0, nil, err
	}
	return stmt.ParamNum(), stmt.ColumnNum(), stmt, nil
}
func (h ServerHandler) HandleStmtExecute(context interface{}, query string, args []interface{}) (*mysql.Result, error) {
	return context.(*client.Stmt).Execute(args...)
}

func (h ServerHandler) HandleStmtClose(context interface{}) error {
	return context.(*client.Stmt).Close()
}

func (h ServerHandler) HandleOtherCommand(cmd byte, data []byte) error {
	return mysql.NewError(
		mysql.ER_UNKNOWN_ERROR,
		fmt.Sprintf("command %d is not supported now", cmd),
	)
}
``` 

Unfortunately, it crashed in `(c *Conn) bindStmtArgs()` when I was testing with sysbench. It seems that when repeatedly executing the same statement with the same parameters, the EXECUTE message indicates that there is no need to re-bind the parameters, but the implementation of handleStmtExecute does not save the types and values of the parameters from the previous execution of the statement.

matttm:
@rfyim Did you ever get this to work?

[...]

PR #983, opened 2025-02-07
server: Correct the check for Resultset

dveeden:
This is likely a regression introduced by #969 

Found by running this code: https://github.com/go-mysql-org/go-mysql/issues/915#issuecomment-2640067648

Queries like `SELECT 1` were working correctly, but `DO 1` and `SET ...` were not as the code tried to write a resultset instead of an OK.

### DIFF
--- a/mysql/result.go
+++ b/mysql/result.go
@@ -34,3 +34,10 @@ func (r *Result) Close() {
 		r.Resultset = nil
 	}
 }
+
+func (r *Result) HasResultset() bool {
+	if r.Resultset != nil && len(r.Resultset.Fields) > 0 {
+		return true
+	}
+	return false
+}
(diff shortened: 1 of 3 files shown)


dveeden:
Without this PR:
```
mysql-8.0.11> SELECT 1;
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)

mysql-8.0.11> DO 1;
ERROR 2027 (HY000): Malformed packet
```

With this PR:
```
mysql-8.0.11> SELECT 1;
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)

mysql-8.0.11> DO 1;
Query OK, 0 rows affected (0.00 sec)
```

lance6716 (approved):

PR #987, opened 2025-02-09
client: Make example testable

dveeden:
This removes the client example from the README and puts it in its own file.

The benefits of this:
- This makes this a [testable example](https://go.dev/blog/examples)
- This shows up in the godoc documentation

The drawbacks of this:
- It isn't in the README anymore
- This might be a bit weird for server examples that need the MySQL client. But there we could just have no output.

Issues found with the example while doing this:
1.  Missing import for `github.com/go-mysql-org/go-mysql/mysql` (it uses `mysql.FieldValueTypeFloat`)
2. The example uses a table named `table`, which is confusing.
3. The example doesn't create the table
4. The `defer r.Close()` should probably be moved higher up
5. The `r.GetInt()` and `r.GetIntByName()` probably need more explanation
6. The test for the float type is never true
7. The values are never printed to standard out
8. The code doesn't run as it has an import, but the rest of the code isn't in a function as expected.
9. The `r` variable is created twice with `:=`
10. Not every result is closed
11. Error handling isn't done (Maybe that's ok or even desired for examples?)

Setting this as draft as I'm not sure if this is the right thing to do or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,54 +190,7 @@ You can see [go-mysql-elasticsearch](https://github.com/go-mysql-org/go-mysql-el
 
 Client package supports a simple MySQL connection driver which you can use it to communicate with MySQL server. 
 
-### Example
-
-```go
-import (
-	"github.com/go-mysql-org/go-mysql/client"
-)
-
-// Connect MySQL at 127.0.0.1:3306, with user root, an empty password and database test
-conn, _ := client.Connect("127.0.0.1:3306", "root", "", "test")
-
-// Or to use SSL/TLS connection if MySQL server supports TLS
-//conn, _ := client.Connect("127.0.0.1:3306", "root", "", "test", func(c *Conn) {c.UseSSL(true)})
-
-// Or to set your own client-side certificates for identity verification for security
-//tlsConfig := NewClientTLSConfig(caPem, certPem, keyPem, false, "your-server-name")
-//conn, _ := client.Connect("127.0.0.1:3306", "root", "", "test", func(c *Conn) {c.SetTLSConfig(tlsConfig)})
-
-conn.Ping()
-
-// Insert
-r, _ := conn.Execute(`insert into table (id, name) values (1, "abc")`)
-
-// Get last insert id
-println(r.InsertId)
-// Or affected rows count
-println(r.AffectedRows)
-
-// Select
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -190,54 +190,7 @@ You can see [go-mysql-elasticsearch](https://github.com/go-mysql-org/go-mysql-el
...
-	for _, val := range row {
-		_ = val.Value() // interface{}
-		// or
-		if val.Type == mysql.FieldValueTypeFloat {
-			_ = val.AsFloat64() // float64
-		}
-	}   
-}
-```
+For an example see [`example_client_test.go`](client/example_client_test.go)
```
lance6716, line 193: I think we can mention how to run it. golang testable examples are not common.

client/example_client_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+
+	"github.com/go-mysql-org/go-mysql/client"
+	"github.com/go-mysql-org/go-mysql/mysql"
+)
+
+func Example() {
+	// Connect MySQL at 127.0.0.1:3306, with user root, an empty password and database test
+	conn, _ := client.Connect("127.0.0.1:3306", "root", "", "test")
```
lance6716, line 12: I think when error is not nil, we can print something to guide the developer to setup a running MySQL for this test.

lance6716 (commented):
rest lgtm

dveeden:
This is how this looks with godoc:
image

Top part ↑
-------------------------------------------
Bottom part ↓

image


So this only renders the contents of the `Example` function, not the context.

dveeden:
This is after fixing the way I run godoc: (now the import is correct)
image

lance6716:
Oh you mean the the purpose of `Example` is to improve godoc of this package? They are also runnable by `go test -v ./client -run Example` so I think we should also consider the friendly instruction when running it.

lance6716:
Hi, do you need some help on this PR?

dveeden:
> Hi, do you need some help on this PR?

Still not 100% sure what the best way is. But I'm getting more sure that testable examples are the right choice.

lance6716:
I have added my commit, PTAL @dveeden

lance6716 (approved):

PR #990, opened 2025-02-14
replication: Add mysql::serialization based Gtid Log Event

### LINKED ISSUE #845: MySQL 8.3 will introduce new GTID format

dveeden:
Closes #845 

With `mysqlbinlog`:

```
# at 158
#250213 14:55:59 server id 1  end_log_pos 241 CRC32 0x52293f34 
# Position  Timestamp   Type   Source ID        Size      Source Pos    Flags 
# 0000009e ef f9 ad 67   2a   01 00 00 00   53 00 00 00   f1 00 00 00   00 00
# 000000b1 02 78 00 00 02 02 25 02  dc f0 09 02 30 f9 03 22 |.x..........0...|
# 000000c1 bd 03 ad 02 21 02 44 44  5a 68 51 03 22 04 d9 03 |......DDZhQ.....|
# 000000d1 06 0c 61 61 62 62 63 63  08 00 0a 04 0c 7f cf de |..aabbcc........|
# 000000e1 ef 6a 06 2e 06 10 45 03  12 c3 02 0b 34 3f 29 52 |.j....E.....4..R|
#       GTID    last_committed=0        sequence_number=1       rbr_only=no     original_committed_timestamp=1739454959050447   immediate_commit_timestamp=1739454959050447 transaction_length=209
# original_commit_timestamp=1739454959050447 (2025-02-13 14:55:59.050447 CET)
# immediate_commit_timestamp=1739454959050447 (2025-02-13 14:55:59.050447 CET)
/*!80001 SET @@session.original_commit_timestamp=1739454959050447*//*!*/;
/*!80014 SET @@session.original_server_version=90200*//*!*/;
/*!80014 SET @@session.immediate_server_version=90200*//*!*/;
SET @@SESSION.GTID_NEXT= '896e7882-18fe-11ef-ab88-22222d34d411:aabbcc:123'/*!*/;
```

With `go-mysqlbinlog` (decoding a similar, but not identical event):
```
=== Gtid_tagged_log_event ===
Date: 2025-02-17 21:14:49
Log position: 240
Event size: 82
Commit flag: 1
GTID_NEXT: 896e7882-18fe-11ef-ab88-22222d34d411:foobaz:1
LAST_COMMITTED: 0
SEQUENCE_NUMBER: 1
Immediate commmit timestamp: 1739823289369365 (2025-02-17T21:14:49.369365+01:00)
Orignal commmit timestamp: 0 (<n/a>)
Transaction length: 210
Immediate server version: 90200
Orignal server version: 0
```


serialization/serialization.go
```diff
@@ -0,0 +1,291 @@
...
+	Optional bool
+	Name     string
+	Skipped  bool
+}
+
+type FieldType interface {
+	fmt.Stringer
+}
+
+type FieldIntFixed struct {
```
lance6716, line 74: this type is called "...Fixed", I guess it means it's fixed size encoded? But it's implemented as a variable-length encoded.

serialization/serialization.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (f FieldIntFixed) String() string {
+	if f.Value == nil {
+		return ""
+	}
+	return fmt.Sprintf("0x%x", f.Value)
+}
+
+type FieldIntVar struct {
```
lance6716, line 86: And seems this is fixed-length encoded

serialization/serialization.go
```diff
@@ -0,0 +1,291 @@
...
+	return f.Value
+}
+
+func Unmarshal(data []byte, v interface{}) error {
+	r := bytes.NewReader(data)
+	switch m := v.(type) {
+	case *Message:
+		messageLen := 1
+		tmpVer := make([]byte, messageLen)
+		_, err := r.Read(tmpVer)
```
lance6716, line 109: the `Read` interface will copy from `data` to given buffer. It's better that we don't need this interface and directly use `data []byte`, like in `(f *Field) Parse` an extra `pos += n` is maintained and `p[pos:]` is used.

serialization/serialization.go
```diff
@@ -0,0 +1,291 @@
...
+	case 8:
+		tNum = binary.LittleEndian.Uint64(fieldBytes)
+	}
+	if unsigned {
+		return (tNum >> (tb + 2) * 2)+1, nil
+	}
+	return tNum >> (tb + 2), nil
+}
+
+func trailingOneBitCount(b byte) (count int) {
```
lance6716, line 278: we can revert the bits and use https://pkg.go.dev/math/bits#TrailingZeros8

dveeden: I assume you meant reverse instead of revert?

lance6716: Bitwise not, `^b`

dveeden:
@serprex if you're interested, maybe you could review this PR?

replication/event.go
```diff
@@ -543,6 +549,195 @@ func (e *GTIDEvent) OriginalCommitTime() time.Time {
...
+	if err != nil {
+		return err
+	}
+	if v, ok := f.Type.(serialization.FieldUintVar); ok {
+		e.TransactionLength = v.Value
+	} else {
+		return errors.New("failed to get transaction_length field")
+	}
+
+	// TODO: add and test commit_group_ticket
```
serprex, line 736: TODO

serialization/serialization.go
```diff
@@ -0,0 +1,283 @@
...
+
+		for i := 0; i < len(m.Fields); i++ {
+			if int(pos)+1 > len(data) || int(data[pos]>>1) != i {
+				// The field number we got doesn't match what we expect,
+				// so a field was skipped.
+				m.Fields[i].ID = i
+				m.Fields[i].Skipped = true
+				continue
+			}
+			m.Fields[i].ID = int(data[pos] >> 1)
```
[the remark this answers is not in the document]
serprex: is it plausible to use `data[pos]>>1` to fill in info about earlier `m.Fields`? ie is it ever allowed for protocol to send these fields out of order?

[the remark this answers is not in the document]
dveeden: Fields are in order, but it is allowed to skip one or more fields.

serialization/serialization.go
```diff
@@ -0,0 +1,283 @@
...
+			slices.Concat(data[pos:int(pos)+flen], []byte{0x0, 0x0, 0x0}))
+	case 6:
+		tNum = binary.LittleEndian.Uint64(
+			slices.Concat(data[pos:int(pos)+flen], []byte{0x0, 0x0}))
+	case 7:
+		tNum = binary.LittleEndian.Uint64(
+			slices.Concat(data[pos:int(pos)+flen], []byte{0x0}))
+	case 8:
+		tNum = binary.LittleEndian.Uint64(data[pos : int(pos)+flen])
+	}
```
serprex, line 270: simpler would be to have an `[8]byte bytes` which you `copy(bytes, data[pos:int(pos)+flen])` & then you can skip switch with `tNum := binary.LittleEndian.Uint64(bytes[:])`
```suggestion
	var bytes [8]byte
	copy(bytes[:], data[pos:int(pos)+flen])
	tNum := binary.LittleEndian.Uint64(bytes[:])
```

serialization/serialization.go
```diff
@@ -0,0 +1,283 @@
...
+	strLen := int(data[pos] >> 1)
+	pos++
+	if len(data) < int(pos)+strLen {
+		return "", pos, fmt.Errorf("string truncated, expected length: %d", strLen)
+	}
+	return string(data[pos : pos+uint64(strLen)]), pos + uint64(strLen), nil
+}
+
+func decodeFixed(data []byte, pos uint64, intlen int) ([]byte, uint64, error) {
+	var b bytes.Buffer
```
serprex, line 210:
```suggestion
	var b bytes.Buffer
	b.Grow(intlen)
```

serprex (approved):

serialization/serialization.go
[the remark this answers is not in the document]
serprex: in that case wouldn't `int(data[pos]>>1) > i`? So condition `(data[pos]>>1) != i` is too strict. Since if I understand you correctly the ID would skip ahead & we'd be expected to fill in the previous fields as skipped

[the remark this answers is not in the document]
serprex: Ah, I understand now. You effectively do that since `continue` advances `i` without changing `pos` so you'll do this until `i` in sync with data

serialization/serialization.go
dveeden: Hm. The fixed input might lead to variable output between `intlen` and `intlen*2`. So `intlen` is often not enough.

serprex: `b.Grow(intlen * 2)` would be suitable in that case

Ran a test on playground, cap is initialized to 8, then grows to 16

lance6716 (commented):
still reviewing, I hope I can find some spare time soon

serialization/serialization.go
```diff
@@ -0,0 +1,263 @@
...
+// FieldString is a 'string_format' field
+type FieldString struct {
+	Value string
+}
+
+func (f FieldString) String() string {
+	return f.Value
+}
+
+func Unmarshal(data []byte, v interface{}) error {
```
lance6716, line 120: it's better to split this function into two dedicated functions for Message and Format? I don't understand why they are merged. seems not shared many logic

serialization/serialization.go
```diff
@@ -0,0 +1,263 @@
...
+			return f, nil
+		}
+	}
+	return Field{}, fmt.Errorf("field not found: %s", name)
+}
+
+// Format is describing a `message_format`
+type Format struct {
+	Size                  uint64
+	LastNonIgnorableField int
```
lance6716, line 44: this is 1 byte as described in mysql source code document URL at the beginning of this file, so the type should be `byte` rather than `int`? also seems above Size field can be up to 9 byte 🤔 not sure if I understand it, will check the whole code file soon

serialization/serialization.go
```diff
@@ -0,0 +1,263 @@
...
+		m.Size = uint64(data[pos] >> 1)
+		pos++
+		m.LastNonIgnorableField = int(data[pos] >> 1)
+		pos++
+
+		for i := 0; i < len(m.Fields); i++ {
+			if int(pos)+1 > len(data) || int(data[pos]>>1) != i {
+				// The field number we got doesn't match what we expect,
+				// so a field was skipped.
+				m.Fields[i].ID = i
```
lance6716, line 139: seems also `ID` field can be expressed in `byte`

serialization/serialization.go
```diff
@@ -0,0 +1,263 @@
...
+				m.Fields[i].Skipped = true
+				continue
+			}
+			m.Fields[i].ID = int(data[pos] >> 1)
+			pos++
+			var n uint64
+			var err error
+			switch f := m.Fields[i].Type.(type) {
```
lance6716, line 147: It's more like a golang-style to let every type (`FieldIntFixed`, `FieldUintVar`, ...) implement an interface with `decode` method. So this type-switch can be replaced by

```
n, err := m.Fields[i].Type.decode(data, pos)
```

replication/event.go
```diff
@@ -543,6 +549,195 @@ func (e *GTIDEvent) OriginalCommitTime() time.Time {
 	return microSecTimestampToTime(e.OriginalCommitTimestamp)
 }
 
+// GtidTaggedLogEvent is for a GTID event with a tag.
+// This is similar to GTIDEvent, but it has a tag and uses a different serialization format.
+type GtidTaggedLogEvent struct {
+	GTIDEvent
+	msg serialization.Message
```
lance6716, line 556: we don't need to store `msg` field in the structure? Seems it's only used in `func (e *GtidTaggedLogEvent) Decode`. A function-local `msg` should be enough

lance6716 (commented):
rest lgtm

lance6716:
Hi @dveeden would you like me to directly modify your PR based on my comments?

dveeden:
> Hi @dveeden would you like me to directly modify your PR based on my comments?

I plan to apply those changes today. If more changes is needed after today you can either modify this PR directly or send a PR for my branch.

serialization/serialization.go
dveeden: This almost works now:

```patch
diff --git a/serialization/serialization.go b/serialization/serialization.go
index 2b71d76..29ff5ed 100644
--- a/serialization/serialization.go
+++ b/serialization/serialization.go
@@ -69,6 +69,7 @@ type Field struct {
 // FieldType represents a `type_field`
 type FieldType interface {
        fmt.Stringer
+       decode(data []byte, pos uint64) (uint64, error)
 }
 
 // FieldIntFixed is for values with a fixed length.
@@ -221,35 +222,9 @@ func Unmarshal(data []byte, v interface{}) error {
                        }
                        m.Fields[i].ID = int(data[pos] >> 1)
                        pos++
-                       var n uint64
-                       var err error
-                       switch f := m.Fields[i].Type.(type) {
-                       case FieldIntFixed:
-                               n, err = f.decode(data, pos)
-                               if err != nil {
-                                       return err
-                               }
-                               m.Fields[i].Type = f
-                       case FieldUintVar:
-                               n, err = f.decode(data, pos)
-                               if err != nil {
-                                       return err
-                               }
-                               m.Fields[i].Type = f
-                       case FieldIntVar:
-                               n, err = f.decode(data, pos)
-                               if err != nil {
-                                       return err
-                               }
-                               m.Fields[i].Type = f
-                       case FieldString:
-                               n, err = f.decode(data, pos)
-                               if err != nil {
-                                       return err
-                               }
-                               m.Fields[i].Type = f
-                       default:
-                               return fmt.Errorf("unsupported field type: %T", m.Fields[i].Type)
+                       n, err := m.Fields[i].Type.decode(data, pos)
+                       if err != nil {
+                               return err
                        }
                        pos = n
                }
```

Gives me this:
```
/home/dvaneeden/dev/go-mysql/serialization/serialization_test.go:238:12: cannot use FieldIntFixed{…} (value of type FieldIntFixed) as FieldType value in struct literal: FieldIntFixed does not implement FieldType (method decode has pointer receiver)
```
```

File: serialization/serialization.go (hunk 9 above)
Comment (dveeden) [reply serialization/serialization.go:193]: If you compare it to `json.Unmarshal()` that also doesn't have different functions depending on the type (array, string, object, etc).
I think I like to keep it like this.

File: replication/event.go (hunk 5 above)
Comment (dveeden) [reply replication/event.go:735]: I haven't yet been able to generate any events with an actual `commit_group_ticket`.

I first tried with `binlog_group_commit_sync_delay=...` but that didn't make any difference.

Then looking at the code I noticed that this has to do with Group Replication (part of InnoDB Cluster), So I [setup](https://dev.mysql.com/doc/mysql-shell/8.0/en/setting-up-innodb-cluster-and-mysql-router.html) a cluster with 3 sandbox instances... but that also didn't give me any events for testing.

Looks like this might also be known as [BGC](https://dev.mysql.com/doc/dev/mysql-server/latest/classbinlog_1_1BgcTicket.html#details) (Binlog Group Commit) tickets.

File: serialization/serialization.go (hunk 12 above)
Comment (lance6716) [reply serialization/serialization.go:231]: Oh I see that's because `decode` will change itself so it should has a pointer receiver. In other words `*FieldIntFixed` implements `FieldType` interface. So we should assign `*FieldIntFixed` like

```diff
diff --git i/replication/event.go w/replication/event.go
index f2056e7..500587f 100644
--- i/replication/event.go
+++ w/replication/event.go
@@ -561,7 +561,7 @@ func (e *GtidTaggedLogEvent) Decode(data []byte) error {
                        Fields: []serialization.Field{
                                {
                                        Name: "gtid_flags",
-                                       Type: serialization.FieldIntFixed{
+                                       Type: &serialization.FieldIntFixed{
                                                Length: 1,
                                        },
                                },
```

Do you it's acceptable? In this way we also don't need to assign back to `Type` field like line 237

lance6716 (commented):
rest lgtm

serialization/serialization.go
dveeden: This doesn't seem to work. I thried this: https://github.com/dveeden/go-mysql/commit/7a582362cdab42ebb41726bf285d058e5f453e05

It flags `f.Value = ...` as unused write. Should we return it instead?

The switch also checks for valid field types, without the switch that's not done anymore.

lance6716: I'll try to provide a valid diff today

PR #992, opened 2025-02-19
replace time.After with Ticker

serprex:
addresses #423 however as of 1.23 the issue is moot & this was never really leaking since the 1 second would generally pass

this does slightly reduce allocations, but this is on an error path, so not valuably

Implemented as proof of concept, #423 can be closed


dveeden:
Thanks. Interesting. If I understand things correctly this is mostly a FYI and should be closed?

serprex:
@dveeden yes

Issue #929, opened 2024-10-26
Update Go version


dveeden:
In `go.mod` we have the version set to 1.18

The Go release policy is "Each major Go release is supported until there are two newer major releases.". This means only 1.23 and 1.22 are in support.

https://go.dev/doc/devel/release#:~:text=Each%20major%20Go%20release%20is%20supported%20until%20there%20are%20two%20newer%20major%20releases.

Not sure how this would impact projects depending on go-mysql.

Not sure if there are any real benefits to bumping the version either.

If we bump the version, should we go to 1.21, 1.22? or 1.23?

```patch
diff --git a/go.mod b/go.mod
index 612dad7..d7e79cf 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/go-mysql-org/go-mysql
 
-go 1.18
+go 1.21
 
 require (
        github.com/BurntSushi/toml v1.3.2
```

@lance6716 @atercattus what do you think?

serprex:
go 1.21 introduced `log/slog`, I could write a PR switching out logger. That would be a breaking change, unless some compatibility layer is kept to support both logging interfaces (which I currently do to wrap slog logger in https://github.com/PeerDB-io/peerdb/blob/main/flow/connectors/mysql/logger.go)

dveeden:
My personal preference is to use the standard library where possible, and slog would match that. However I don't know what @atercattus and @lance6716 think of that.

lance6716:
I also prefer the standard library. For compatibility problem, I think the developers won't be bothered much if the new logger can be configurated easily, or we provide a converter function, or we release a major version which means API break.

serprex:
go.mod bumped to 1.22 by #993, this issue can be closed

PR #993, opened 2025-02-20
replace github.com/siddontang/go-log with log/slog

serprex:
as discussed in #929

There's room for further improvement. Ideally `slog.Info` functions would never be called, as these force usage of the global slog default. Instead all instances should somehow receive a logger. Or, if that isn't feasible, a global variable for the package should allow overriding the package's logger independently from the application *(tho should still end up being `slog.Default()` in this case)*

This is a breaking change, & since current interface includes setting handler in go-log to control logging, there's no way to make this not breaking


mysql/mariadb_gtid.go
```diff
@@ -93,7 +93,7 @@ func (gtid *MariadbGTID) forward(newer *MariadbGTID) error {
 		| mysqld-bin.000001 | 2215 | Gtid              |       111 |        2257 | BEGIN GTID 0-111-6  |
 	*/
 	if newer.SequenceNumber <= gtid.SequenceNumber {
-		log.Warnf("out of order binlog appears with gtid %s vs current position gtid %s", newer, gtid)
+		slog.Warn("out of order binlog", slog.Any("new", newer), slog.Any("current", gtid))
```
serprex, line 96: Seems like it may be best to remove this check, as it's difficult to have a non-global logger here

lance6716: I haven't find a better way for this problem. May we acn add a package level logger in future. For now I prefer we keep your code.

lance6716 (approved):

README.md
```diff
@@ -537,26 +536,7 @@ We pass all tests in https://github.com/bradfitz/go-sql-test using go-mysql driv
...
-To write logging to any [`io.Writer`](https://pkg.go.dev/io#Writer):
-```go
-import "github.com/siddontang/go-log/log"
-...
-        w := ...
-        streamHandler, _ := log.NewStreamHandler(w)
-        cfg.Logger = log.NewDefault(streamHandler)
-```
-
-Or you can implement your own [`log.Handler`](https://pkg.go.dev/github.com/siddontang/go-log/log#Handler).
```
lance6716, line 559: The slog is more complex to be implemented at the user side. Please leave the guidance in readme https://github.com/golang/example/blob/master/slog-handler-guide/README.md

lance6716 (commented):
Oh I just found it's not trivial to convert github.com/siddontang/go-log to log/slog 😂 I think it's better to merge it when we have a converting helper function or provide clear guidance.

README.md
serprex: I generally expect people will use an existing adapter like https://github.com/samber/slog-zerolog or https://github.com/samber/slog-zap

serprex: something like https://github.com/serprex/slog-siddontang

lance6716: Thanks! I'll take a look soon.

lance6716:
Hi @dveeden I want to merge this PR, and in readme or release note we remind the developers about this change. If the developers want the old logger he/she can still use https://github.com/serprex/slog-siddontang to fit the interface. What do you think?

dveeden:
> Hi @dveeden I want to merge this PR, and in readme or release note we remind the developers about this change. If the developers want the old logger he/she can still use https://github.com/serprex/slog-siddontang to fit the interface. What do you think?

This sounds good to me. I'll double check this PR today and then you can merge it later today or tomorrow.

dveeden:
@lance6716 looks like there is a merge conflict? And this is in draft status?

README.md
```diff
@@ -537,26 +536,7 @@ We pass all tests in https://github.com/bradfitz/go-sql-test using go-mysql driv
...
+Logging uses log/slog.
```
dveeden, line 539:
```suggestion
Logging uses [log/slog](https://pkg.go.dev/log/slog) and by default is sent to standard out.
```

canal/dump.go
```diff
@@ -49,7 +50,7 @@ func (h *dumpParseHandler) Data(db string, table string, values []string) error
 			e == schema.ErrMissingTableMeta {
 			return nil
 		}
-		h.c.cfg.Logger.Errorf("get %s.%s information err: %v", db, table, err)
+		h.c.cfg.Logger.Error("error getting table information", slog.String("databse", db), slog.String("table", table), slog.Any("error", err))
```
dveeden, line 53:
```suggestion
		h.c.cfg.Logger.Error("error getting table information", slog.String("database", db), slog.String("table", table), slog.Any("error", err))
```

canal/dump.go
```diff
@@ -163,7 +164,7 @@ func (c *Canal) dump() error {
 		if err != nil {
 			return errors.Trace(err)
 		}
-		c.cfg.Logger.Infof("skip master data, get current binlog position %v", pos)
+		c.cfg.Logger.Info("skip master data, get current binlog position", slog.Any("pos", pos))
```
dveeden, line 167:
```suggestion
		c.cfg.Logger.Info("skip master data, get current binlog position", slog.Any("position", pos))
```

We should aim to have meaningful and easy to understand names that are consistent. Might be better to use "position" instead of "pos" for the key of the value here. Feel free to ignore if you don't agree.

canal/dump.go
```diff
@@ -185,8 +186,8 @@ func (c *Canal) dump() error {
 		c.master.UpdateGTIDSet(h.gset)
 		startPos = h.gset
 	}
-	c.cfg.Logger.Infof("dump MySQL and parse OK, use %0.2f seconds, start binlog replication at %s",
-		time.Since(start).Seconds(), startPos)
+	c.cfg.Logger.Info(fmt.Sprintf("dump MySQL and parse OK, use %0.2f seconds, start binlog replication at %s",
+		time.Since(start).Seconds(), startPos))
```
dveeden, line 190: Shouldn't we put the time and position in separate fields?

Something like this, maybe?
```suggestion
	c.cfg.Logger.Info("dump MySQL and parse OK", "use", slog.Time(time.Since(start)), "position", startPos)
```

Or keep the fmt.Sprintf, but add the fields as well?

canal/dump.go
```diff
@@ -196,7 +197,7 @@ func (c *Canal) tryDump() error {
 	if (len(pos.Name) > 0 && pos.Pos > 0) ||
 		(gset != nil && gset.String() != "") {
 		// we will sync with binlog name and position
-		c.cfg.Logger.Infof("skip dump, use last binlog replication pos %s or GTID set %v", pos, gset)
+		c.cfg.Logger.Info(fmt.Sprintf("skip dump, use last binlog replication pos %s or GTID set %v", pos, gset))
```
dveeden, line 200: Shouldn't we put the set and pos into fields?

dveeden (approved):
I don't really like the log+sprintf that's used in many places, however I think this is fine as a transition and can even help to reduce the changes in the default user visible output.

canal/dump.go
serprex: `time.Since` returns a `time.Duration` while `slog.Time` expects `time.Time`. Will update to pass fields

canal/dump.go
serprex: Yes

serprex:
> @lance6716 looks like there is a merge conflict? And this is in draft status?

PR was started as proof of concept, so was waiting on feedback on whether the design was acceptable or not. Will clean it up & remove draft status

dveeden:
To build it needs something like this:

```diff
diff --git a/replication/binlogsyncer.go b/replication/binlogsyncer.go
index e95fbfd..cc50849 100644
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -862,7 +862,7 @@ func (b *BinlogSyncer) handleEventAndACK(s *BinlogStreamer, e *BinlogEvent, need
        case *RotateEvent:
                b.nextPos.Name = string(event.NextLogName)
                b.nextPos.Pos = uint32(event.Position)
-               b.cfg.Logger.Info("rotate to next binlog", slog.String("file", b.nextPos.Name), slog.Int("position", b.nextPos))
+               b.cfg.Logger.Info("rotate to next binlog", slog.String("file", b.nextPos.Name), slog.Uint64("position", uint64(b.nextPos.Pos)))
 
        case *GTIDEvent:
                if b.prevGset == nil {
```

Weirdly enough it doesn't give a merge conflict for me locally...

dveeden:
We could do something like this as example:
```
diff --git a/cmd/go-mysqlbinlog/main.go b/cmd/go-mysqlbinlog/main.go
index 12f3ced..bb0bd12 100644
--- a/cmd/go-mysqlbinlog/main.go
+++ b/cmd/go-mysqlbinlog/main.go
@@ -7,6 +7,7 @@ import (
        "context"
        "flag"
        "fmt"
+       "log/slog"
        "os"
 
        "github.com/pingcap/errors"
@@ -31,6 +32,7 @@ var (
        backupPath = flag.String("backup_path", "", "backup path to store binlog files")
 
        rawMode = flag.Bool("raw", false, "Use raw mode")
+       format  = flag.String("format", "plain", "log format")
 )
 
 func main() {
@@ -50,6 +52,15 @@ func main() {
                MaxReconnectAttempts: 10,
        }
 
+       switch *format {
+       case "json":
+               cfg.Logger = slog.New(slog.NewJSONHandler(os.Stderr, nil))
+       case "plain":
+               // This is the default format
+       default:
+               panic("unsupported log format")
+       }
+
        err := mysql.ValidateFlavor(*flavor)
        if err != nil {
                fmt.Printf("Flavor error: %v\n", errors.ErrorStack(err))
```

```
dvaneeden@dve-carbon:~/dev/go-mysql$ ./bin/go-mysqlbinlog -format json > /dev/null
{"time":"2025-03-04T15:56:05.619590715+01:00","level":"INFO","msg":"begin to sync binlog from position","position":{"Name":"","Pos":4}}
{"time":"2025-03-04T15:56:05.625397936+01:00","level":"INFO","msg":"Connected to server","flavor":"mysql","version":"9.2.0"}
{"time":"2025-03-04T15:56:05.625582082+01:00","level":"INFO","msg":"rotate to next binlog","file":"binlog.000001","position":4}
^C
```

dveeden:
Or even like this:
```
dvaneeden@dve-carbon:~/dev/go-mysql$ ./bin/go-mysqlbinlog -format json > /dev/null
{"time":"2025-03-04T16:04:40.707995695+01:00","level":"INFO","msg":"begin to sync binlog from position","position":{"Name":"","Pos":4}}
{"time":"2025-03-04T16:04:40.711097722+01:00","level":"INFO","msg":"Connected to server","flavor":"mysql","version":"9.2.0"}
{"time":"2025-03-04T16:04:40.711274055+01:00","level":"INFO","msg":"rotate to next binlog","file":"binlog.000001","position":4}
^C
dvaneeden@dve-carbon:~/dev/go-mysql$ ./bin/go-mysqlbinlog -format json -verbose > /dev/null
{"time":"2025-03-04T16:04:46.85523127+01:00","level":"INFO","source":{"function":"github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).StartSync","file":"/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go","line":440},"msg":"begin to sync binlog from position","position":{"Name":"","Pos":4}}
{"time":"2025-03-04T16:04:46.860961252+01:00","level":"INFO","source":{"function":"github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).prepare","file":"/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go","line":407},"msg":"Connected to server","flavor":"mysql","version":"9.2.0"}
{"time":"2025-03-04T16:04:46.861151704+01:00","level":"INFO","source":{"function":"github.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).handleEventAndACK","file":"/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go","line":865},"msg":"rotate to next binlog","file":"binlog.000001","position":4}
^C
dvaneeden@dve-carbon:~/dev/go-mysql$ ./bin/go-mysqlbinlog -verbose > /dev/null
time=2025-03-04T16:04:58.933+01:00 level=INFO source=/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go:440 msg="begin to sync binlog from position" position="(, 4)"
time=2025-03-04T16:04:58.938+01:00 level=INFO source=/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go:407 msg="Connected to server" flavor=mysql version=9.2.0
time=2025-03-04T16:04:58.939+01:00 level=INFO source=/home/dvaneeden/dev/go-mysql/replication/binlogsyncer.go:865 msg="rotate to next binlog" file=binlog.000001 position=4
^C
```

```diff
diff --git a/cmd/go-mysqlbinlog/main.go b/cmd/go-mysqlbinlog/main.go
index 12f3ced..512be58 100644
--- a/cmd/go-mysqlbinlog/main.go
+++ b/cmd/go-mysqlbinlog/main.go
@@ -7,6 +7,7 @@ import (
        "context"
        "flag"
        "fmt"
+       "log/slog"
        "os"
 
        "github.com/pingcap/errors"
@@ -31,6 +32,8 @@ var (
        backupPath = flag.String("backup_path", "", "backup path to store binlog files")
 
        rawMode = flag.Bool("raw", false, "Use raw mode")
+       format  = flag.String("format", "plain", "log format")
+       verbose = flag.Bool("verbose", false, "verbose logging")
 )
 
 func main() {
@@ -50,6 +53,19 @@ func main() {
                MaxReconnectAttempts: 10,
        }
 
+       logOpts := &slog.HandlerOptions{
+               AddSource: *verbose,
+       }
+
+       switch *format {
+       case "json":
+               cfg.Logger = slog.New(slog.NewJSONHandler(os.Stderr, logOpts))
+       case "plain":
+               cfg.Logger = slog.New(slog.NewTextHandler(os.Stderr, logOpts))
+       default:
+               panic("unsupported log format")
+       }
+
        err := mysql.ValidateFlavor(*flavor)
        if err != nil {
                fmt.Printf("Flavor error: %v\n", errors.ErrorStack(err))
```

serprex:
I think readme is best place, for configuring logging people should either setup default slog logger (for consistent logging across app), or pass in logger (which tends to happen when code paths have enriched logger with extra fields)

dveeden (approved):

README.md
```diff
@@ -537,26 +536,7 @@ We pass all tests in https://github.com/bradfitz/go-sql-test using go-mysql driv
...
-```go
-import "github.com/siddontang/go-log/log"
-...
-        w := ...
-        streamHandler, _ := log.NewStreamHandler(w)
-        cfg.Logger = log.NewDefault(streamHandler)
-```
-
-Or you can implement your own [`log.Handler`](https://pkg.go.dev/github.com/siddontang/go-log/log#Handler).
+Logging uses [log/slog](https://pkg.go.dev/log/slog) and by default is sent to standard out.
```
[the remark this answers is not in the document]
serprex: * stdout/stderr is configured when creating handler, will change to stdout, could be flag
* config is my mistake, changed `Infof` to `Debug`, will change back to Info
* filenames should be included if you pass `--verbose`

dveeden:
To me it looks like this is now ready to be merged. @lance6716 , do you agree?

replication/binlogsyncer.go
```diff
@@ -169,11 +168,10 @@ type BinlogSyncer struct {
 // NewBinlogSyncer creates the BinlogSyncer with the given configuration.
 func NewBinlogSyncer(cfg BinlogSyncerConfig) *BinlogSyncer {
 	if cfg.Logger == nil {
-		streamHandler, _ := log.NewStreamHandler(os.Stdout)
-		cfg.Logger = log.NewDefault(streamHandler)
+		cfg.Logger = slog.Default()
 	}
 	if cfg.ServerID == 0 {
-		cfg.Logger.Fatal("can't use 0 as the server ID")
+		panic("can't use 0 as the server ID")
```
lance6716, line 174: panic may send output to a different place than `cfg.Logger`, I prefer

```suggestion
		cfg.Logger.Error("can't use 0 as the server ID, will panic")
		panic("can't use 0 as the server ID")
```

README.md
```diff
@@ -491,26 +490,7 @@ We pass all tests in https://github.com/bradfitz/go-sql-test using go-mysql driv
...
```
lance6716, line 494:
```suggestion

For the old logging package `github.com/siddontang/go-log/log`, a converting package 
`https://github.com/serprex/slog-siddontang` is available.
```

lance6716 (approved):
thanks @serprex , please check if my comments are acceptable.

PR #995, opened 2025-02-20
Use math/bits OnesCount to preallocate slices

serprex:


replication/row_event.go
```diff
@@ -1146,15 +1147,19 @@ func (e *RowsEvent) decodeImage(data []byte, bitmap []byte, rowImageType EnumRow
...
 
 	// refer: https://github.com/alibaba/canal/blob/c3e38e50e269adafdd38a48c63a1740cde304c67/dbsync/src/main/java/com/taobao/tddl/dbsync/binlog/event/RowsLogBuffer.java#L63
 	count := 0
-	for i := 0; i < int(e.ColumnCount); i++ {
-		if isBitSet(bitmap, i) {
+	col := 0
+	for ; col+8 <= int(e.ColumnCount); col += 8 {
+		count += bits.OnesCount8(bitmap[col>>3])
+	}
+	for ; col < int(e.ColumnCount); col++ {
```
lance6716, line 1157: for this leftover, if the out-of-range bits are zero, we still can use `bits.OnesCount8`

lance6716 (approved):
waiting to check the optimization in comment

replication/row_event.go
serprex: Yes, that'd be ideal. Not sure about mysql server itself, but current unit tests have a 2 column count followed by 0xff byte

lance6716: oh I think we can mask it, then use bits.OnesCount8

lance6716 (approved):

PR #997, opened 2025-02-21
driver: context support

serprex:
`database/sql` `QueryContext` falls back to checking if context is done before calling `Query`, so query does not get cancelled when context is cancelled once execution starts

[driver.Queryer is deprecated](https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/database/sql/driver/driver.go;l=203)

Unfortunately supporting this on top of net.Conn requires a goroutine. But that goroutine can be shared across lifetime of connection since connection doesn't need to be threadsafe

To cancel ongoing Read/Write the connection is closed when context finishes

At the end of the function active context is switched to `context.Background()`. Some error handling could be implemented around receiving non-background context when active context is a non-background context, as that implies thread safety has been violated

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -35,6 +35,8 @@ type Conn struct {
 	// Connection read and write timeouts to set on the connection
 	ReadTimeout  time.Duration
 	WriteTimeout time.Duration
+	contexts     chan context.Context
+	closed       bool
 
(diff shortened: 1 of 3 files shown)


lance6716 (approved):
Thanks!

PR #998, opened 2025-02-21
add support for MariaDB ED25519 plugin

heavycrystal:
MariaDB doesn't support `caching_sha2_password` or `sha256_password` authentication plugins, and instead has a completely different `client_ed25519` plugin, which uses a modified version of the ED25519 algorithm. MariaDB recommends using this over `mysql_native_password`.

This PR implements support for `client_ed25519`. The bulk of this PR is taken from https://github.com/go-sql-driver/mysql/pull/1518 with the author's permission. 

Co authored with: @Gusted

### DIFF
--- a/mysql/util.go
+++ b/mysql/util.go
@@ -7,6 +7,7 @@ import (
 	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/binary"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


lance6716 (approved):

mysql/util.go
```diff
@@ -83,6 +85,44 @@ func CalcCachingSha2Password(scramble []byte, password string) []byte {
...
+
+	s, err := edwards25519.NewScalar().SetBytesWithClamping(h[:32])
+	if err != nil {
+		return nil, err
+	}
+	A := (&edwards25519.Point{}).ScalarBaseMult(s)
+
+	mh := sha512.New()
+	mh.Write(h[32:])
+	mh.Write(scramble)
```
dveeden, line 102: Is the length of the scramble expected to be the same? should we add a check for the length?

heavycrystal: the length of scramble is checked (to be 32) before this function is called (L176-178 in client/auth.go)

PR #1000, opened 2025-02-25
CI: Test on 32-bit arm

dveeden:
See also:
- https://github.com/pingcap/tidb/issues/59751
- https://github.com/pingcap/tidb/pull/59753

This should probably be merged once the above PR is merged and we've updated the parser.


lance6716:
we can also update the "Platform Support" section in readme

dveeden:
@lance6716 @atercattus this is ready for review now

lance6716 (approved):

PR #1001, opened 2025-02-25
Add BinlogEvent.ByteLen

dt8269:
Add a `.ByteLen()` method to `BinlogEvent` so we can match a `GTIDEvent`’s `TransactionLength` with the actual byte size of events.

### DIFF
--- a/replication/event.go
+++ b/replication/event.go
@@ -38,6 +38,11 @@ func (e *BinlogEvent) Dump(w io.Writer) {
 	e.Event.Dump(w)
 }
 
+// ByteLen returns the total size of the raw event data in bytes.
+func (e *BinlogEvent) ByteLen() int {
+	return len(e.RawData)
+}
+
 type Event interface {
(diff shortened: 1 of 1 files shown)


dveeden (commented):
Could you add a comment? That helps the godoc docs

dt8269:
> Could you add a comment? That helps the godoc docs

added

dveeden (approved):

replication/event.go
```diff
@@ -38,6 +38,11 @@ func (e *BinlogEvent) Dump(w io.Writer) {
 	e.Event.Dump(w)
 }
 
+// ByteLen returns the total size of the raw event data in bytes.
+func (e *BinlogEvent) ByteLen() int {
+	return len(e.RawData)
```
lance6716, line 43: lgtm. Just to confirm that without this change you can still use `len(e.RawData)` to get the same result?

lance6716: ping @dt8269

PR #1006, opened 2025-03-05
Improve unit tests around `Hasresultset` and `WriteValue`

### LINKED ISSUE #1005: zero-value of `mysql.Result` was expected by `server.WriteValue`, breaks in `v1.11.0`
https://github.com/go-mysql-org/go-mysql/commit/d02e79aea163459997784240e0e8bcf0deb3535b introduced new constructors for `mysql.Result` that always initializes `Result.Resultset` to be non-nil, but this breaks the logic in `server.WriteValue` that differentiates between a resultset response and an OK response. This in turn breaks responses to things like `INSERT` statements.

This is missed in tests because `server/server_test.go` creates `Result`s without the constructors, and with nil `Resultset`s

I have a fix https://github.com/userclouds/go-mysql/tree/stg/server_missed_ok that I'm happy to PR if it makes sense and I'm not mis-understanding the intent here.

Thanks!

stgarrity:
Followup to #969 and #983

resolves #1005 (comment re unit tests, bug is already fixed in #983)


mysql/result_test.go
```diff
@@ -17,3 +17,17 @@ func TestHasResultset_true(t *testing.T) {
 	b := r.HasResultset()
 	require.Equal(t, true, b)
 }
+
+// this shouldn't happen after d02e79a, but test just in case
+func TestHasResultset_nilset(t *testing.T) {
+	r := NewResultReserveResultset(0)
+	r.Resultset = nil
+	b := r.HasResultset()
+	require.Equal(t, false, b)
```
lance6716, line 26: can use `require.False`

mysql/result_test.go
```diff
@@ -17,3 +17,17 @@ func TestHasResultset_true(t *testing.T) {
...
+}
+
+func TestHasResultset_nil(t *testing.T) {
+	var r *Result
+	b := r.HasResultset()
+	require.Equal(t, false, b)
```
lance6716, line 32: ditto

lance6716 (commented):
rest lgtm

stgarrity:
good call, fixed those and a couple of existing places in the same file. thanks!

dveeden (approved):

PR #1012, opened 2025-03-10
CI: Test with MySQL 5.7

### LINKED ISSUE #775: Run tests for mysql 5.x without a github ubuntu-18.04 image
ubuntu-18.04 image [is deprecated already](https://github.com/actions/runner-images/issues/6002). We use this image for mysql 5.7 image.

image

We should use another way to run old MySQL versions.

dveeden:
Closes #775

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,8 +56,9 @@ jobs:
     strategy:
       matrix:
         mysql_version:
-          - 8.0.40
-          - 8.4.3
+          - '5.7'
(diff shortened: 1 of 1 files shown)


dveeden:
- It seems to be difficult/impossible to specify container arguments (not options) with service containers
- Specifying a `my.cnf` etc. for a service container also seems to be harder that it should be.
- MySQL 5.7 needs config: `server-id=100`+`log_bin` and maybe also TLS config.

PR #1016, opened 2025-03-20
Give `RowsEvent` a method to inspect the underlying event type.

proton-lisandro-pin:
It's currently impossible to figure out what kind of operation led to a `RowsEvent` 😞
Multiple results indicate an `UPDATE`, but there's no way to differentiate an `INSERT` from a `DELETE`.

This PR implements a new method to determine the event type, regardless of version and/or engine.


replication/row_event.go
```diff
@@ -1120,6 +1120,18 @@ func (e *RowsEvent) Decode(data []byte) error {
 	return e.DecodeData(pos, data)
 }
 
+func (e *RowsEvent) IsInsert() bool {
```
lance6716, line 1123: Yes currently `RowsEvent` has no public type information. You can know the type from `EventHeader`

lance6716 (commented):
nit: It's better to add a separate type `RowsEventType`? Because I think the new methods of this PR can be used like

```
if e.IsInsert() {
  ...
} else if e.IsUpdate {
 ...
}
...
```

`e.eventType` will be checked for many times.

dveeden:
Does this work correctly with compressed binlog events in MySQL?

proton-lisandro-pin:
> nit: It's better to add a separate type `RowsEventType`? Because I think the new methods of this PR can be used like

Sure :smile: I was mostly following the existing style for `Is*Column()` methods for `TableMapEvent`s, but i do like introducing a dedicated type better. Fixed.

proton-lisandro-pin:
> Does this work correctly with compressed binlog events in MySQL?

Not sure tbh. The change covers all binlog versions _currently handled_ by the module, so if compressed binlogs are already supported, then yes.

dveeden:
> > Does this work correctly with compressed binlog events in MySQL?
> 
> Not sure tbh. The change covers all binlog versions _currently handled_ by the module, so if compressed binlogs are already supported, then yes.

These are in `TransactionPayloadEvent`, Not sure if and how that's handled. Might be good to test.

lance6716:
For TransactionPayloadEvent, I see that it still holds sub-events like normal BinlogEvent
https://github.com/go-mysql-org/go-mysql/blob/567aa5991d5a7762e4419e4a1fd816efa4256635/replication/transaction_payload_event.go#L29-L35
And the events are decoded like it's in a normal code path
https://github.com/go-mysql-org/go-mysql/blob/567aa5991d5a7762e4419e4a1fd816efa4256635/replication/transaction_payload_event.go#L144-L148

The problem is we didn't have a unit test to verify it, like capturing the wire data and decoding the TransactionPayloadEvent. I'll consider add such unit test later.

proton-lisandro-pin:
> For TransactionPayloadEvent, I see that it still holds sub-events like normal BinlogEvent

Hmm, would it make sense to move (and rename) `EnumRowsEventType` to `replication/const.go` then?

I imagine access to the operation type would be useful for `TransactionPayloadEvent` as well down the line.

lance6716:
> Hmm, would it make sense to move (and rename) EnumRowsEventType to replication/const.go then?

Do you mean because it's a TransactionPayloadEvent rather than RowsEvent, changing the name will better reflect its usage? I think we don't need to do that. Because we can see TransactionPayloadEvent holds BinlogEvent, and when we cast the types of BinlogEvent we still get RowsEvent, it's OK to have a dedicated type name for RowsEvent.

proton-lisandro-pin:
> I have created this PR #1018 , can you take a look? After merging it, I hope you can modify the test to verify your changes still work for TransactionPayloadEvent

Can't approve :smile: but the test looks a-ok to me.

proton-lisandro-pin:
> I have merged that PR. Please merge / rebase master branch and make use that test

Done, works perfectly :+1:

lance6716 (approved):

PR #1019, opened 2025-03-28
replication: add IsEmpty method to GTIDSet interface & implementations

wesleywright:
Small quality of life improvement. For some use cases (such as writing tests), it is common to want to check if a given GTID set is empty or not. Currently this requires using `Equal()` with an empty GTID set (which takes a decent amount of boilerplate) or else requires inspecting the `Set` field (which is specific to a single concrete implementation of the `GTIDSet` interface).

This commits adds a simple `IsEmpty() bool` method to the `GTIDSet` interface for this use case, as well as implementations for both `MysqlGTIDSet` and `MariadbGTIDSet`.

### DIFF
--- a/mysql/gtid.go
+++ b/mysql/gtid.go
@@ -17,6 +17,9 @@ type GTIDSet interface {
 	Update(GTIDStr string) error
 
 	Clone() GTIDSet
+
+	// IsEmpty returns true if the given set is empty and false otherwise.
+	IsEmpty() bool
 }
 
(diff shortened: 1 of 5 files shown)


wesleywright:
This is my first contribution to this repo; it seemed like a pretty small and straightforward change so I thought it was simplest to directly submit it as a PR. I'm happy to take any feedback on it :-)

lance6716 (approved):

PR #1021, opened 2025-04-01
server: Rework NewConn/defaultServer

### LINKED ISSUE #1020: Expensive default variable creation
The RSA keys generated for the defaultServer global variable have a dramatically negative impact on any consumers of the `server` package.

https://github.com/go-mysql-org/go-mysql/blob/639eccdeb52729e521e1164c44fad1580c91a151/server/server_conf.go#L11

This can be measured using `GODEBUG='inittrace=1'` on any programs that imports `github.com/go-mysql-org/go-mysql/server`. 

From a M2:

```212 ms github.com/go-mysql-org/go-mysql/server```

From a t3.xlarge:

```537 ms github.com/go-mysql-org/go-mysql/server```

### PR DESCRIPTION
**Pull request by dveeden**
Closes #1020 

- De-duplicate `NewConn()` and `NewCustomizedConn()` code
- Only call `NewDefaultServer()` when needed

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -36,8 +36,10 @@ type Conn struct {
 	closed atomic.Bool
 }
 
-var baseConnID uint32 = 10000
-var defaultServer *Server
+var (
+	baseConnID    uint32 = 10000
+	defaultServer *Server
+)
 
 // NewConn: create connection with default server settings
 //
--- a/server/conn.go
+++ b/server/conn.go
@@ -41,7 +41,6 @@ var defaultServer *Server
  // NewConn: create connection with default server settings
 //
 // Deprecated: Use [Server.NewConn] instead.
-// Deprecated: Use [Server.NewConn] instead.
 func NewConn(conn net.Conn, user string, password string, h Handler) (*Conn, error) {
 	if defaultServer == nil {
 		defaultServer = NewDefaultServer()
--- a/server/conn.go
+++ b/server/conn.go
@@ -3,6 +3,7 @@ package server
 import (
 	"errors"
 	"net"
+	"sync"
 	"sync/atomic"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
@@ -38,17 +39,16 @@ type Conn struct {
 
 var (
 	baseConnID    uint32 = 10000
-	defaultServer *Server
+	defaultServer        = sync.OnceValue(func() *Server {
+		return NewDefaultServer()
+	})
 )
 
 // NewConn: create connection with default server settings
 //
 // Deprecated: Use [Server.NewConn] instead.
 func NewConn(conn net.Conn, user string, password string, h Handler) (*Conn, error) {
-	if defaultServer == nil {
-		defaultServer = NewDefaultServer()
-	}
-	return defaultServer.NewConn(conn, user, password, h)
+	return defaultServer().NewConn(conn, user, password, h)
 }
 
 // NewCustomizedConn: create connection with customized server settings
--- a/server/conn.go
+++ b/server/conn.go
@@ -48,7 +48,6 @@ func NewConn(conn net.Conn, user string, password string, h Handler) (*Conn, err
 // NewCustomizedConn: create connection with customized server settings
 //
 // Deprecated: Use [Server.NewConn] instead.
-
 // NewCustomizedConn: create connection with customized server settings
 // Deprecated: Use Server.NewConn instead.
 func NewCustomizedConn(conn net.Conn, serverConf *Server, p CredentialProvider, h Handler) (*Conn, error) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: server/conn.go [hunk 1]
```diff
@@ -42,31 +42,9 @@ var baseConnID uint32 = 10000
 func NewConn(conn net.Conn, user string, password string, h Handler) (*Conn, error) {
 	p := NewInMemoryProvider()
 	p.AddUser(user, password)
+	server := NewDefaultServer()
```
lance6716, line 45: multiple `NewConn` will call multiple times of `NewDefaultServer`. We still need to cache the result of `NewConn`

dveeden: Maybe `NewConn` should be a method of `type Server`?

server/conn.go
```diff
@@ -37,50 +37,43 @@ type Conn struct {
 }
 
 var baseConnID uint32 = 10000
+var defaultServer *Server
 
 // NewConn: create connection with default server settings
```
ccoVeille, line 42: 🎨 nitpicking

 You would break the godoc comment if you do like that

This is the expected format

```suggestion
 // NewConn: create connection with default server settings
//
// Deprecated: Use Server.NewConn instead.
```

You can see the difference by using [pkgsite](https://github.com/golang/pkgsite) locally

ccoVeille, line 42: 🎨 You should use godoc links, it would help people to navigate to the new resource.

```suggestion
// Deprecated: Use [Server.NewConn] instead.
```

Here also you can use pkgsite to check the difference

server/conn.go
```diff
@@ -37,50 +37,43 @@ type Conn struct {
...
-	var packetConn *packet.Conn
-	if defaultServer.tlsConfig != nil {
-		packetConn = packet.NewTLSConn(conn)
-	} else {
-		packetConn = packet.NewConn(conn)
+	if defaultServer == nil {
+		defaultServer = NewDefaultServer()
 	}
+	return defaultServer.NewConn(conn, user, password, h)
+}
```
ccoVeille, line 49: Same remarks as above


```suggestion
// NewCustomizedConn: create connection with customized server settings
//
// Deprecated: Use [Server.NewConn] instead.
```

ccoVeille (commented):
Some remarks, otherwise LGTM 👍

server/conn.go
```diff
@@ -37,50 +37,45 @@ type Conn struct {
 }
 
 var baseConnID uint32 = 10000
-
-// NewConn: create connection with default server settings
+var defaultServer *Server
+ // NewConn: create connection with default server settings
+//
+// Deprecated: Use [Server.NewConn] instead.
+// Deprecated: Use [Server.NewConn] instead.
```
ccoVeille, line 44: I'm unsure why GitHub messed up my suggestion.

There is no reason for this line to be duplicated

```suggestion
```

server/conn.go
```diff
@@ -37,50 +37,45 @@ type Conn struct {
...
-		serverConf:         defaultServer,
-		credentialProvider: p,
-		h:                  h,
-		connectionID:       atomic.AddUint32(&baseConnID, 1),
-		stmts:              make(map[uint32]*Stmt),
-		salt:               mysql.RandomBuf(20),
-	}
-	c.closed.Store(false)
+// NewCustomizedConn: create connection with customized server settings
+// Deprecated: Use Server.NewConn instead.
```
ccoVeille, line 54: Here again, I'm unsure why my suggestion lead to this.

The two block are present, this one should be deleted

server/conn.go
```diff
@@ -37,50 +37,45 @@ type Conn struct {
...
-	if defaultServer.tlsConfig != nil {
-		packetConn = packet.NewTLSConn(conn)
-	} else {
-		packetConn = packet.NewConn(conn)
+	if defaultServer == nil {
+		defaultServer = NewDefaultServer()
 	}
+// NewCustomizedConn: create connection with customized server settings
+//
+// Deprecated: Use [Server.NewConn] instead.
```
ccoVeille, line 52: This line should be deleted

```suggestion
```

ccoVeille (changes requested):

server/conn.go
```diff
@@ -37,50 +37,44 @@ type Conn struct {
...
-	p := NewInMemoryProvider()
-	p.AddUser(user, password)
-
-	var packetConn *packet.Conn
-	if defaultServer.tlsConfig != nil {
-		packetConn = packet.NewTLSConn(conn)
-	} else {
-		packetConn = packet.NewConn(conn)
+	if defaultServer == nil {
+		defaultServer = NewDefaultServer()
```
lance6716, line 47: Accessing `defaultServer` will cause data race if `NewConn` is called concurrently. I also prefer use a global variable of type `sync.OnceValue` to achieve a) lazily creating the server only when calling it b) caching the computed result for later usage.

ccoVeille (approved):

lance6716 (approved):

PR #1023, opened 2025-04-04
server: Fix formatting of credential provider note

dveeden:
Fixes this:

image

An alternative:
```patch
diff --git a/server/credential_provider.go b/server/credential_provider.go
index 3d44eb0..5333f89 100644
--- a/server/credential_provider.go
+++ b/server/credential_provider.go
@@ -4,7 +4,9 @@ import "sync"
 
 // interface for user credential provider
 // hint: can be extended for more functionality
+//
 // =================================IMPORTANT NOTE===============================
+//
 // if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
 // for 'caching_sha2_password' by calling 'func (s *Server)InvalidateCache(string, string)'.
 type CredentialProvider interface {
```

### DIFF
--- a/server/credential_provider.go
+++ b/server/credential_provider.go
@@ -5,7 +5,9 @@ import "sync"
 // interface for user credential provider
 // hint: can be extended for more functionality
 //
-// Note: if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
+// # Important Note
+//
+// if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
 // for 'caching_sha2_password' by calling 'func (s *Server)InvalidateCache(string, string)'.
 type CredentialProvider interface {
 	// check if the user exists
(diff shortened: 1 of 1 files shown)


server/credential_provider.go
```diff
@@ -4,8 +4,8 @@ import "sync"
 
 // interface for user credential provider
 // hint: can be extended for more functionality
-// =================================IMPORTANT NOTE===============================
-// if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
+//
+// Note: if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
```
ccoVeille, line 8: I would like you to suggest using this

```suggestion
//
// # Important Note
//
// if the password in a third-party credential provider could be updated at runtime, we have to invalidate the caching
```

Source https://tip.golang.org/doc/comment#headings

lance6716 (approved):

PR #1027, opened 2025-04-27
feat: Charset parameter for Mysql connection

sigalikanevsky:


README.md
```diff
@@ -132,7 +132,7 @@ type MyEventHandler struct {
 }
 
 func (h *MyEventHandler) OnRow(e *canal.RowsEvent) error {
-	log.Infof("%s %v\n", e.Action, e.Rows)
+	log.Debugf("%s %v\n", e.Action, e.Rows)
```
lance6716, line 135: the log level should be kept, other developers may rely on this behaviour.
Also for other files.

canal/canal.go
```diff
@@ -65,6 +65,10 @@ func NewCanal(cfg *Config) (*Canal, error) {
 
 	c.ctx, c.cancel = context.WithCancel(context.Background())
 
+	if cfg.WaitTimeBetweenConnectionSeconds > 0 {
```
lance6716, line 68: it should be `<= 0` to set the default value?

canal/sync.go
```diff
@@ -270,38 +279,10 @@ func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
 		return errors.Errorf("%s not supported now", e.Header.EventType)
 	}
 	events := newRowsEvent(t, action, ev.Rows, e.Header)
+	events.Header.Gtid = c.SyncedGTIDSet()
```
lance6716, line 282: Can you explain why this is changed?

replication/binlogsyncer.go
```diff
@@ -697,7 +711,8 @@ func (b *BinlogSyncer) onStream(s *BinlogStreamer) {
 							return
 						}
 
-						log.Errorf("retry sync err: %v, wait 1s and retry again", err)
+						log.Errorf("retry sync err: %v, wait %s s and retry again", err, b.cfg.WaitTimeBetweenConnectionSeconds)
+						time.Sleep(b.cfg.WaitTimeBetweenConnectionSeconds)
```
lance6716, line 715: the sleep is at line 705. Don't need to add extra `time.Sleep`.

replication/binlogsyncer.go
```diff
@@ -719,7 +734,11 @@ func (b *BinlogSyncer) onStream(s *BinlogStreamer) {
 
 		switch data[0] {
 		case OK_HEADER:
-			if err = b.parseEvent(s, data); err != nil {
+			if err = b.parseEvent(b.nextPos.Name, s, data); err != nil {
+				// if the error is errMissingTableMapEvent skip
```
lance6716, line 738: data may be lost if we don't return error

replication/row_event.go
```diff
@@ -1141,17 +1145,34 @@ func (e *RowsEvent) decodeValue(data []byte, tp byte, meta uint16) (v interface{
...
+		fmt.Printf("islatin is: %v", e.Table.isLatin)
+		if !utf8.Valid(data) && e.Table.isLatin {
+			v, n = decodeStringLatin1(data, length)
+		} else {
+			v, n = decodeString(data, length)
+		}
 	case MYSQL_TYPE_STRING:
-		v, n = decodeString(data, length)
+		if !utf8.Valid(data) && e.Table.isLatin {
+			v, n = decodeStringLatin1(data, length)
```
lance6716, line 1163: The decoding and converting should be done at upper layer, not here. `replication` package aims to provide a golang representation of binlog.

lance6716 (changes requested):
Please fix git conflict

dveeden (changes requested):
Please reference an issue or describe the changes in the description of the PR

replication/row_event.go
```diff
@@ -1182,6 +1217,29 @@ func decodeString(data []byte, length int) (v string, n int) {
 	return
 }
 
+func decodeStringLatin1(data []byte, length int) (v string, n int) {
+	// Define the Latin1 decoder
+	decoder := charmap.ISO8859_1.NewDecoder()
```
dveeden, line 1222: Note that `latin1` in MySQL is `cp1252` and not `ISO 8859-1`:

"MySQL's latin1 is the same as the Windows cp1252 character set. This means it is the same as the official ISO 8859-1 or IANA (Internet Assigned Numbers Authority) latin1, except that IANA latin1 treats the code points between 0x80 and 0x9f as “undefined,” whereas cp1252, and therefore MySQL's latin1, assign characters for those positions. For example, 0x80 is the Euro sign. For the “undefined” entries in cp1252, MySQL translates 0x81 to Unicode 0x0081, 0x8d to 0x008d, 0x8f to 0x008f, 0x90 to 0x0090, and 0x9d to 0x009d. "

source: https://dev.mysql.com/doc/refman/8.4/en/charset-we-sets.html


```suggestion
	decoder := charmap.Windows1252.NewDecoder()
```

replication/row_event.go
```diff
@@ -1182,6 +1217,29 @@ func decodeString(data []byte, length int) (v string, n int) {
 	return
 }
 
+func decodeStringLatin1(data []byte, length int) (v string, n int) {
```
dveeden, line 1220: Might be good to make things a bit more generic, so it could work with multiple charsets and not just `latin1`.

However I think a vast majority of MySQL installations use either `latin1` or `utf8`/`utf8mb4`. Maybe some `ascii`, but that's a subset of both `latin1` and `utf8`/`utf8mb4`...

dveeden:
@sigalikanevsky Could you fix the conflicts by merging the master branch into your branch?

canal/sync.go
```diff
@@ -270,38 +279,10 @@ func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
...
-				return nil
-			} else {
-				log.Debugf("master pos is %v, wait catching %v", curPos, pos)
-				time.Sleep(100 * time.Millisecond)
-			}
-		}
-	}
-
-	return nil
-}
```
ccoVeille, line 303: Removing this break the API of the package

I don't see why they were moved as functions accepting Canal

canal/sync.go
```diff
@@ -142,7 +143,15 @@ func (c *Canal) runSyncBinlog() error {
 		case *replication.QueryEvent:
 			stmts, _, err := c.parser.Parse(string(e.Query), "", "")
 			if err != nil {
-				log.Errorf("parse query(%s) err %v, will skip this event", e.Query, err)
+				msg := err.Error()
+				if strings.Contains(strings.ToLower(msg), strings.ToLower("procedure")) {
```
ccoVeille, line 147: Parsing error message like this is a bad practice.

Don't you have an error code should could catch?

canal/sync.go
```diff
@@ -22,15 +23,15 @@ func (c *Canal) startSyncer() (*replication.BinlogStreamer, error) {
 		if err != nil {
 			return nil, errors.Errorf("start sync replication at binlog %v error %v", pos, err)
 		}
-		log.Infof("start sync binlog at binlog file %v", pos)
+		log.Debugf("start sync binlog at binlog file %v", pos)
```
ccoVeille, line 26: You might think there is a need to change the log lever for these information messages.

But it should be done in another PR, not this one.

It brings confusion, especially because there are no links between these changes and the topic of the PR

canal/sync.go
ccoVeille: +1

replication/binlogsyncer.go
```diff
@@ -28,6 +30,9 @@ type BinlogSyncerConfig struct {
 	// Flavor is "mysql" or "mariadb", if not set, use "mysql" default.
 	Flavor string
 
+	// MaxWaitTimeBetween Check connection
+	WaitTimeBetweenConnectionSeconds time.Duration
```
ccoVeille, line 34:
```suggestion
	// WaitTimeBetweenConnectionSeconds is the delay between two connections when an connection fails
	WaitTimeBetweenConnectionSeconds time.Duration
```

Also, it's a time.Duration, so for naming you should avoid mentioning seconds


```suggestion
	// DelayBetweenConnection is the delay between two connections when an connection fails
	DelayBetweenConnection time.Duration
```

replication/binlogsyncer.go
```diff
@@ -719,7 +730,11 @@ func (b *BinlogSyncer) onStream(s *BinlogStreamer) {
 
 		switch data[0] {
 		case OK_HEADER:
-			if err = b.parseEvent(s, data); err != nil {
+			if err = b.parseEvent(b.nextPos.Name, s, data); err != nil {
+				// if the error is errMissingTableMapEvent skip
+				if strings.Contains(strings.ToLower(errors.Cause(err).Error()), errMissingTableMapEvent.Error()) {
```
ccoVeille, line 735: This way to test the error is absolutely discouraged.

Using errors.Is maybe?

dveeden:
@sigalikanevsky It looks like you made a few small changes to your fork of go-mysql and now use the same branch for this PR. However that will also include all the changes that you had on your branch before you started doing this small improvement. And probably your branch was significantly behind our master branch.

I would suggest that you start with a fresh version of our master branch and then start cherry picking the commits that you need and update/fix then so they work with our master branch. 

Also I would to reduce the differences between our branch and your branch where possible. This makes it easier for you to contribute things and also makes it easier for you to update to a new version without having to rebase all your changes.

PR #1028, opened 2025-04-28
update semver to semver/v3

serprex:


dveeden:
Why is TestCompareServerVersions failing?

dveeden:
Looks like it won't parse `8.0.32-0ubuntu0.20.04.2`. Maybe we should truncate the part after the `-`?

Side note: MySQL versions are not really semantic: 8.0.11 has 8.0 as major version and not 8 for example. However that shouldn't really matter much.

MySQL versions for MariaDB, TiDB, Vitess, etc can be a bit weird: e.g. `8.0.11-TiDB-v9.0.0-alpha-658-gc619031356` which means it is TiDB v9.0.0 with git hash gc619031356 that claims to be MySQL 8.0.11.

And Percona versions and MySQL Enterprise and Cloud versions can also have all kinds of suffixes. (e.g. `8.0.36-u5-cloud`)

dveeden:
Maybe we could/should do version parsing in a way that would be useful to users of the library as well. For example to extract the implementation (MariaDB, TiDB, etc) and the version

serprex:
Offering convention checks is unfortunately inconsistent. Noticed semver package being unsupported while investigating what eventually resulted in https://github.com/PeerDB-io/peerdb/pull/2897 since RDS MariaDB returns something like `10.2.0-MariaDB` from `version()`, but serverVersion from handshake is more like `5.5.0-10.2.0-MariaDB`

I think keeping scope to the small existing scope is good. Users who want tag info can use `_, tag, hasTag := strings.Cut(version, "-")`

dveeden (approved):

serprex:
This change can be an issue if consumers specify only 8.0 instead of 8.0.0 to compare function

But hard to define behavior. Likely it should behave as 8.0.x not 8.0.0, skipping patch check. Would only require checking third return value of Cut & returning 0 whenever number missing

mysql/util.go
```diff
@@ -452,21 +453,43 @@ func ErrorEqual(err1, err2 error) bool {
...
+		return 0, fmt.Errorf("cannot parse %s version %s of %s", typ, a, aFull)
+	}
+	if bNum, err = strconv.Atoi(b); err != nil {
+		return 0, fmt.Errorf("cannot parse %s version %s of %s", typ, b, bFull)
+	}
+
+	return cmp.Compare(aNum, bNum), nil
+}
+
 func CompareServerVersions(a, b string) (int, error) {
```
lance6716, line 470: Please add some comments to explain the behaviour of this function. Like it will ignore bytes after `-`, it assumes input string must in format `{major}.{minor}.{patch}`

lance6716 (approved):

mysql/util.go
serprex: Commented (with description of extended wildcard behavior) & added more tests

mysql/util.go
```diff
@@ -452,21 +453,50 @@ func ErrorEqual(err1, err2 error) bool {
...
 
-	if aVer, err = semver.NewVersion(a); err != nil {
-		return 0, fmt.Errorf("cannot parse %q as semver: %w", a, err)
+	if minorCompare, err := compareSubVersion("minor", aMinor, bMinor, a, b); err != nil || minorCompare != 0 {
+		return minorCompare, err
 	}
 
-	if bVer, err = semver.NewVersion(b); err != nil {
-		return 0, fmt.Errorf("cannot parse %q as semver: %w", b, err)
+	if patchCompare, err := compareSubVersion("patch", aPatch, bPatch, a, b); err != nil || patchCompare != 0 {
```
lance6716, line 495: maybe `return compareSubVersion("patch", aPatch, bPatch, a, b)`

lance6716 (commented):
rest lgtm

PR #1030, opened 2025-05-02
fix canal.GetMasterPos() for mariadb

### LINKED ISSUE #1029: `func (c *canal.Canal) GetMasterPos()` always errors when using the mariadb flavor on latest mariadb v11
When running `func (c *canal.Canal) GetMasterPos()` with a `cfg.Flavor = "mariadb"`, the method always fails with the following error:

```
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'LOG STATUS' at line 1
```

It always fails because [these lines](https://github.com/go-mysql-org/go-mysql/blob/34b6b0998dde44e51dff0bbcc1ac88339f57f830/canal/sync.go#L328-L331) assume only mysql versions, excluding mariadb.

The proper way to get the master position is either using `SHOW MASTER STATUS` for mariadb version previous than `10.5.2` or the more recent `SHOW BINLOG STATUS`. [Here's the source](https://mariadb.com/kb/en/show-binlog-status/#:~:text=SHOW%20MASTER%20STATUS%20was%20renamed,an%20alias%20for%20compatibility%20purposes.).

alarbada:
Adds support for mariadb for the method `canal.GetMasterPos()`.

Fixes #1029


dveeden (commented):
Might be good to add some references to release notes and/or documentation.

This could be made into a function that takes a flavor and version and returns the query string. That would make it easier to add a unit test for this.

canal/sync.go
```diff
@@ -325,12 +325,21 @@ func (c *Canal) WaitUntilPos(pos mysql.Position, timeout time.Duration) error {
 }
 
 func (c *Canal) GetMasterPos() (mysql.Position, error) {
-	showBinlogStatus := "SHOW BINARY LOG STATUS"
-	if eq, err := c.conn.CompareServerVersion("8.4.0"); (err == nil) && (eq < 0) {
-		showBinlogStatus = "SHOW MASTER STATUS"
+	query := "SHOW BINARY LOG STATUS"
+	switch c.cfg.Flavor {
+	case mysql.MariaDBFlavor:
+		query = "SHOW BINLOG STATUS"
```
lance6716, line 331: I think we should use `SHOW MASTER STATUS` as the default query for MariaDB? as https://mariadb.com/kb/en/show-binlog-status/#:~:text=SHOW%20MASTER%20STATUS%20was%20renamed,an%20alias%20for%20compatibility%20purposes. said, it has better compatibility. Just in case that `CompareServerVersion` has some problems.

Also please reply to dveeden's comments

alarbada: Done

lance6716 (approved):

alarbada:
@lance6716 thanks for accepting my pull request :)
I do need this fix to add mariadb support on the mysql conduit connector I'm working on. When do you think you could release this little fix, in a `v1.12.1` or on the upcoming `v1.13.0`?

dveeden:
@alarbada I assume you've tested with `go mod edit -replace ...` to make sure this is the only issue that's needed to make this work with recent MariaDB releases?

alarbada:
> @alarbada I assume you've tested with `go mod edit -replace ...` to make sure this is the only issue that's needed to make this work with recent MariaDB releases?

Not really, after backporting the fix to my repo I got further problems with the `replication.RowsEvent`. For some reason the `LogPos` from EventHeader is 0, I'm debugging why.

I will add an issue (hopefully a PR) once I figure out exactly why is this happneing.

dveeden:
@alarbada OK, Please let me know once you figured this out. I'll wait with releasing a new version.

serprex:
@dveeden as a point of comparison, we're working with MariaDB on PeerDB, & had this fix in our logic already: https://github.com/PeerDB-io/peerdb/blob/84fe9ce286a8c8cc03a6369e4b7ce4b2f7505d1c/flow/connectors/mysql/mysql.go#L305

Link also shows references for versions where language shifted

PR #1036, opened 2025-05-13
Fix: charset latin1

sigalikanevsky:


dveeden:
@sigalikanevsky There are merge conflicts that need to be fixed. This is also missing either a link to an issue or a good description about what this is fixing exactly.

canal/canal.go
```diff
@@ -65,6 +65,10 @@ func NewCanal(cfg *Config) (*Canal, error) {
 
 	c.ctx, c.cancel = context.WithCancel(context.Background())
 
+	if cfg.WaitTimeBetweenConnectionSeconds > 0 {
```
serprex, line 68: this seems odd, why replace all configs with 5 seconds?

replication/row_event.go
```diff
@@ -1182,6 +1220,33 @@ func decodeString(data []byte, length int) (v string, n int) {
 	return
 }
 
+// Replaces smart quotes with ASCII equivalents
+func replaceUnsupportedLatin1Characters(s string) string {
+	s = string(bytes.ReplaceAll([]byte(s), []byte("‘"), []byte("'")))
```
serprex, line 1225: Should avoid repeatedly converting to/from string

replication/row_event.go
```diff
@@ -1182,6 +1220,33 @@ func decodeString(data []byte, length int) (v string, n int) {
 	return
 }
 
+// Replaces smart quotes with ASCII equivalents
+func replaceUnsupportedLatin1Characters(s string) string {
```
serprex, line 1224:
```suggestion
func replaceUnsupportedLatin1Characters(s []byte) string {
```

PR #1041, opened 2025-05-31
client: make CLIENT_QUERY_ATTRIBUTES opt-out to restore proxy compatibility

masahide:
### Background — why proxies built with this SDK break

https://github.com/masahide/mysql8-audit-proxy/issues/8

A common architecture for a MySQL **audit / filter / router** built on *go-mysql* looks like this:

```
[application] -- go-mysql/server  -->  [proxy]  -- go-mysql/client -->  [mysqld]
```

* **Downstream side** (`go-mysql/server`)
  *Implements the old 5.7/8.0 protocol* and **does not recognise
  `CLIENT_QUERY_ATTRIBUTES`**.
  When an application connects, the handshake therefore has the bit **cleared**.

* **Upstream side** (`go-mysql/client` ≥ v1.12.0)
  In `writeAuthHandshake()` the client **automatically copies
  `CLIENT_QUERY_ATTRIBUTES` from the server greeting**, claiming the proxy
  will send query attributes.

* **Mismatch**
  The proxy simply forwards the raw `COM_QUERY` it received from the
  application.
  Because the downstream never added the attributes section, the packet sent
  upstream is missing mandatory fields → MySQL returns
  **`ERROR 1835 (HY000) Malformed communication packet.`**

### What the proxy author needs

A way to say **“don’t advertise `CLIENT_QUERY_ATTRIBUTES` on the upstream
connection even if the server supports it, because my downstream side can’t
handle that mode.”**

Currently that is impossible: `UnsetCapability()` removes the flag from
`ccaps`, but `writeAuthHandshake()` puts it back from the server’s flag set.

### Patch overview

1. **Add `dcaps` (“deny caps”)** to `client.Conn` to track flags explicitly
   disabled by the caller.

2. Modify `SetCapability` / `UnsetCapability`:

   ```go
   SetCapability(x): ccaps |= x ; dcaps &^= x
   UnsetCapability(x): ccaps &^= x ; dcaps |= x
   ```

3. In `writeAuthHandshake()`:

   ```go
   inherit := c.capability &^ c.dcaps          // server flags minus explicit denies
   // ...
   // Advertise QUERY_ATTRIBUTES only if:
   //   a) caller opted-in  OR
   //   b) server supports it AND caller did NOT opt-out
   ```

4. Store the final negotiated set back into `c.capability` for later use.

### Resulting behaviour

| Scenario                | Downstream (server pkg) | Proxy call                                 | Upstream handshake |
| ----------------------- | ----------------------- | ------------------------------------------ | ------------------ |
| **Default**             | no support              | none                                       | bit **set** (auto) |
| Proxy wants legacy mode | no support              | `UnsetCapability(CLIENT_QUERY_ATTRIBUTES)` | bit **cleared**    |
| Force new mode anyway   | anything                | `SetCapability(CLIENT_QUERY_ATTRIBUTES)`   | bit **set**        |

### Example — disabling the flag in a proxy

```go
backend, _ := client.Connect(
    upstreamAddr, user, pass, db,
    func(c *client.Conn) error {
        c.UnsetCapability(mysql.CLIENT_QUERY_ATTRIBUTES)
        return nil
    },
)
```

With this one-liner the proxy once again works with both modern MySQL servers
and the existing `go-mysql/server` front end.

### Safety & compatibility

* **No API break** – existing code compiles; only behaviour changes when the
  caller explicitly opts out.
* Maintains the convenient “auto-enable” path for direct clients.
* Restores plug-and-play compatibility for every proxy built on
  *go-mysql/server*.

### DIFF
--- a/client/auth.go
+++ b/client/auth.go
@@ -211,8 +211,15 @@ func (c *Conn) writeAuthHandshake() error {
 	capability := mysql.CLIENT_PROTOCOL_41 | mysql.CLIENT_SECURE_CONNECTION |
 		mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_PLUGIN_AUTH
 	// Adjust client capability flags based on server support
-	capability |= c.capability & mysql.CLIENT_LONG_FLAG
-	capability |= c.capability & mysql.CLIENT_QUERY_ATTRIBUTES
+	// ---- Inherit capabilities that the server has and the user has NOT explicitly denied ----
+	inherit := c.capability & ^c.dcaps             // Server-side capabilities minus explicit denies
+	capability |= inherit & mysql.CLIENT_LONG_FLAG // Existing
+	// ---- Handling of CLIENT_QUERY_ATTRIBUTES ----
+	if c.ccaps&mysql.CLIENT_QUERY_ATTRIBUTES != 0 { // Explicitly ON
+		capability |= mysql.CLIENT_QUERY_ATTRIBUTES
+	} else if inherit&mysql.CLIENT_QUERY_ATTRIBUTES != 0 { // Server has it and no denial
+		capability |= mysql.CLIENT_QUERY_ATTRIBUTES
+	}
 	// Adjust client capability flags on specific client requests
 	// Only flags that would make any sense setting and aren't handled elsewhere
 	// in the library are supported here
@@ -360,6 +367,8 @@ func (c *Conn) writeAuthHandshake() error {
 		// zstd_compression_level
 		data[pos] = 0x03
 	}
+
+	c.capability = capability // update capability to the one we sent
 
 	return c.WritePacket(data)
 }
--- a/client/auth.go
+++ b/client/auth.go
@@ -211,15 +211,8 @@ func (c *Conn) writeAuthHandshake() error {
 	capability := mysql.CLIENT_PROTOCOL_41 | mysql.CLIENT_SECURE_CONNECTION |
 		mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_PLUGIN_AUTH
 	// Adjust client capability flags based on server support
-	// ---- Inherit capabilities that the server has and the user has NOT explicitly denied ----
-	inherit := c.capability & ^c.dcaps             // Server-side capabilities minus explicit denies
-	capability |= inherit & mysql.CLIENT_LONG_FLAG // Existing
-	// ---- Handling of CLIENT_QUERY_ATTRIBUTES ----
-	if c.ccaps&mysql.CLIENT_QUERY_ATTRIBUTES != 0 { // Explicitly ON
-		capability |= mysql.CLIENT_QUERY_ATTRIBUTES
-	} else if inherit&mysql.CLIENT_QUERY_ATTRIBUTES != 0 { // Server has it and no denial
-		capability |= mysql.CLIENT_QUERY_ATTRIBUTES
-	}
+	capability |= c.capability & mysql.CLIENT_LONG_FLAG
+	capability |= c.capability & mysql.CLIENT_QUERY_ATTRIBUTES
 	// Adjust client capability flags on specific client requests
 	// Only flags that would make any sense setting and aren't handled elsewhere
 	// in the library are supported here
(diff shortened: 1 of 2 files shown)


client/auth.go
```diff
@@ -211,8 +211,15 @@ func (c *Conn) writeAuthHandshake() error {
 	capability := mysql.CLIENT_PROTOCOL_41 | mysql.CLIENT_SECURE_CONNECTION |
 		mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_PLUGIN_AUTH
 	// Adjust client capability flags based on server support
-	capability |= c.capability & mysql.CLIENT_LONG_FLAG
-	capability |= c.capability & mysql.CLIENT_QUERY_ATTRIBUTES
+	// ---- Inherit capabilities that the server has and the user has NOT explicitly denied ----
+	inherit := c.capability & ^c.dcaps             // Server-side capabilities minus explicit denies
```
dveeden: maybe name this `inheritedCaps` or `inherited`?

client/conn.go
```diff
@@ -45,6 +45,7 @@ type Conn struct {
 	capability uint32
 	// client-set capabilities only
 	ccaps uint32
+	dcaps uint32 // disabled capabilities
```
lance6716: I see you followed the naming style of `ccaps`. Just a weak comment: `capability`, `ccaps` and `dcaps` are hard to understand at their usage functions. Maybe rename them to `serverCaps`, `clientExplicitOnCaps`, `clientExplicitOffCaps` or something.

client/auth.go
```diff
@@ -211,8 +211,15 @@ func (c *Conn) writeAuthHandshake() error {
 	capability := mysql.CLIENT_PROTOCOL_41 | mysql.CLIENT_SECURE_CONNECTION |
 		mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_PLUGIN_AUTH
 	// Adjust client capability flags based on server support
-	capability |= c.capability & mysql.CLIENT_LONG_FLAG
-	capability |= c.capability & mysql.CLIENT_QUERY_ATTRIBUTES
```
lance6716: I think we don't need to specially process CLIENT_QUERY_ATTRIBUTES. There are 3 states of capability:

1. developer calls `SetCapability`, which means the `ccaps` bit is 1 and the `dcaps` bit is 0
2. developer calls `UnsetCapability`, which means `ccaps` bit is 0 and the `dcaps` bit is 1
3. No calls, which means both bits are 0

So here we turn on the bits by `ccaps` at around line 219, and have another round to turn off bits by `dcaps`. I suggest

```
capability |= c.ccaps&mysql.CLIENT_FOUND_ROWS | ...
capability &= c.dcaps... // ⬅️ here we add CLIENT_QUERY_ATTRIBUTES and other flags from above line to turn off

masahide:
<html>
<body>
<html><head></head><body><h3>2nd round update — why this patch? 🚀</h3>
<p>Thanks for the quick feedback!<br>
I pushed a follow-up commit that folds the earlier design into the existing <code inline="">ccaps</code> world, so virtually <strong>no downstream call-sites have to change</strong>.</p>
<hr>
<h4>✨  What changed</h4>

Area | Before | Now
-- | -- | --
Explicit OFF mask | dcaps | clientExplicitOffCaps (clearer name)
Negotiated caps field | ccaps contained only “explicit ON” bits | ccaps is now the final negotiated set, so HasCapability() etc. keep working untouched
Handshake build | Special‐case logic around inherit + CLIENT_QUERY_ATTRIBUTES | Generic recipe:1. start with library defaults2. OR server caps we want (CLIENT_LONG_FLAG, CLIENT_QUERY_ATTRIBUTES, …)3. OR whatever the caller explicitly enabled (ccaps)4. AND-NOT whatever they explicitly disabled (clientExplicitOffCaps)
Setter helpers | Set/UnsetCapability twiddled both on/off masks | Same intent, but write to the new field names
Post-send bookkeeping | Overwrote c.capability = capability | Leave server greeting in serverCaps, negotiated value lives in ccaps (safer, fewer surprises)


<p>All other capability flags inherit the same three-state behaviour automatically.</p>
<hr>
<h4>📝  Why this shape?</h4>
<ul>
<li>
<p><strong>Minimal surface-change</strong> — Keeping <code inline="">ccaps</code> as “truth” means every existing <code inline="">HasCapability</code>, compression switch, etc. keeps compiling &amp; behaving.</p>
</li>
<li>
<p><strong>Explicit naming</strong> — “<code inline="">clientExplicitOffCaps</code>” reads self-explanatory next to <code inline="">ccaps</code>.</p>
</li>
<li>
<p><strong>No more special cases</strong> — <code inline="">CLIENT_QUERY_ATTRIBUTES</code> is now just another bit; easier to reason about.</p>
</li>
</ul>
<p>Let me know if you spot anything I’ve missed!</p></body></html>
</body>
</html>

lance6716 (approved):

PR #1043, opened 2025-06-05
Add visible field to schema Index

chenhuiyeh:
This pull request introduces a new feature targeting MySQL version 8 and above, where the concept of invisible indexes has been introduced. Specifically, this PR adds a Visible field to the schema.Index struct, allowing users to easily filter out invisible indexes.

We encountered an issue with [ghostferry](https://github.com/Shopify/ghostferry) where an invisible index was mistakenly selected as the [shardingKeyIndexName](https://github.com/Shopify/ghostferry/blob/674ba761598a0eaa1f9b5cd0baa06d86570b643d/sharding/filter.go#L162-L194). To address this, we initially applied a patch directly to the ghostferry project, as documented in [ghostferry PR #383](https://github.com/Shopify/ghostferry/pull/383). This current PR seeks to integrate those changes into the upstream codebase for broader applicability. Part of https://github.com/Shopify/ghostferry/issues/384.


schema/schema.go
```diff
@@ -320,6 +321,56 @@ func (ta *Table) fetchColumnsViaSqlDB(conn *sql.DB) error {
...
 
+// isMySQL8OrAbove checks if the given version string represents MySQL 8.0 or above.
+// It handles various MySQL version formats including:
+// - "8.0.25"
+// - "8.4.0"
+// - "8.0.25-log"
+// - "8.0.25-0ubuntu0.20.04.1"
+func isMySQL8OrAbove(version string) bool {
+
+	// Extract the version number before any additional info (like "-log", "-ubuntu", etc.)
```
lance6716: `CompareServerVersions` has been updated, the comment says "ignoring anything past `-` in version.". I think you can directly use it.

schema/schema.go
```diff
@@ -338,6 +391,14 @@ func (ta *Table) fetchIndexes(conn mysql.Executer) error {
 		colName, _ := r.GetString(i, 4)
 		currentIndex.AddColumn(colName, cardinality)
 		currentIndex.NoneUnique, _ = r.GetUint(i, 1)
+
+		// Only set to false if explicitly marked as invisible
+		if hasInvisibleIndex {
```
lance6716, line 368: nit: considering MariaDB use case, we can check the columns returned by `show index` to know if there are `Visible` column.

lance6716 (commented):
rest lgtm

schema/schema.go
```diff
@@ -320,6 +321,56 @@ func (ta *Table) fetchColumnsViaSqlDB(conn *sql.DB) error {
...
+	// Use semver comparison for accurate version checking
+	if eq, err := mysql.CompareServerVersions(versionStr, "8.0.0"); err == nil && eq >= 0 {
+		return true
+	}
+
+	return false
+}
+
+func hasInvisibleIndexSupport(conn mysql.Executer) bool {
+	versionQuery := "SELECT VERSION()"
```
dveeden: Maybe we can use the version from the handshake and skip the extra roundtrip?

schema/schema.go
chenhuiyeh: I updated to use `show index` and check for the `Visible` column to determine if an index is visible/invisible and if it has invisible index support

schema/schema.go
chenhuiyeh: Instead of comparing version, I think using `show index` to check for the `Visible` column is enough to see if it has invisible index support. I removed the logic for comparing versions. What do you think?

schema/schema.go
chenhuiyeh: Similar comment as above ^ Instead of comparing version, I think using show index to check for the Visible column is enough to see if it has invisible index support. I removed the logic for comparing versions. What do you think?

schema/schema.go
```diff
@@ -338,6 +365,13 @@ func (ta *Table) fetchIndexes(conn mysql.Executer) error {
 		colName, _ := r.GetString(i, 4)
 		currentIndex.AddColumn(colName, cardinality)
 		currentIndex.NoneUnique, _ = r.GetUint(i, 1)
+		// Only set to false if explicitly marked as invisible
+		if hasInvisibleIndex {
+			visible, _ := r.GetString(i, 13)
+			if isIndexInvisible(visible) {
```
lance6716: maybe 

```
currentIndex.Visible = !isIndexInvisible(visible)
```

schema/schema.go
dveeden: I think this is much better. This would make it easier to support the same for MySQL, TiDB, MariaDB, etc. without too many version checks.

schema/schema.go
chenhuiyeh: I updated the two functions `fetchIndex` and `fetchIndexesViaSqlDB` to use `currentIndex.Visible = !isIndexInvisible(visible)`

lance6716 (approved):

PR #1048, opened 2025-07-04
prefer atomic values to atomic operations

serprex:
not sure why there's no atomic.NewUint32,
so leave baseConnID alone

### DIFF
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -51,7 +51,7 @@ type Canal struct {
 	includeTableRegex []*regexp.Regexp
 	excludeTableRegex []*regexp.Regexp
 
-	delay *uint32
+	delay atomic.Uint32
 
 	ctx    context.Context
 	cancel context.CancelFunc
@@ -84,8 +84,6 @@ func NewCanal(cfg *Config) (*Canal, error) {
 		c.errorTablesGetTime = make(map[string]time.Time)
 	}
 	c.master = &masterInfo{logger: c.cfg.Logger}
-
-	c.delay = new(uint32)
 
 	var err error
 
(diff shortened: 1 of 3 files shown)


mysql/util.go
```diff
@@ -171,13 +170,10 @@ func AppendLengthEncodedInteger(b []byte, n uint64) []byte {
 
 func RandomBuf(size int) []byte {
 	buf := make([]byte, size)
-	// When this project supports golang 1.20 as a minimum, then this mrand.New(...)
-	// line can be eliminated and the random number can be generated by simply
-	// calling mrand.Intn()
-	random := mrand.New(mrand.NewSource(time.Now().UTC().UnixNano()))
-	min, max := 30, 127
+	random := mrand.New(mrand.NewSource(int64(mrand.Uint64())))
```
lance6716: the seed is still read from global generator. It's better to use current time?

mysql/util.go
```diff
@@ -171,13 +170,10 @@ func AppendLengthEncodedInteger(b []byte, n uint64) []byte {
...
+	min, max := int32(30), int32(127)
 	for i := 0; i < size; i++ {
-		buf[i] = byte(min + random.Intn(max-min))
```
lance6716: by checking the source code of `Intn` I think there's not big different with `Int31n`. We don't need to update this file?

mysql/util.go
serprex: It's probably too soon, go is moving towards rand.Seed being a nop, & without calls to Seed in recent versions you get RNG seeded by system's entropy

probably best to switch to `math/rand/v2` here, but this would be better split out, so will revert the RNG change for now

lance6716 (approved):

PR #1052, opened 2025-07-21
fix(replication): calculate LogPos for non-artificial events in MariaDB 11.4+

### LINKED ISSUE #1035: canal.Canal emits an empty e.Header.LogPos event in mariadb 11.4
# BUG

I originally mentioned this behaviour [here](https://github.com/go-mysql-org/go-mysql/pull/1030#issuecomment-2853379039).

The LogPos header value is empty, as you will be able to see in the test logs. I've looked into the output of `show binlog events in 'mysql-bin.000002'` and it shows me this:

Image

I can see that for the Write_rows_v1 event the end_log_pos is 0. The Pos is still non zero though.

I've checked the code [here](https://github.com/alarbada/go-mysql/blob/76417ef684eb4e1493051a1e6168cbafc5724c92/replication/event.go#L72-L102), which I believe is the responsible for parsing the event header, but I think it is fine.

I'm a bit lost on how to continue debugging this. From the docker commands `canal.Canal` should in theory be able to fill the LogPos, I cannot see any mariadb docs specifying that LogPos is empty:

https://mariadb.com/kb/en/2-binlog-event-header

# how to reproduce

I've created a small reproduction example here:

https://github.com/alarbada/go-mysql/tree/mariadb-logpos-bug

Basically, git clone the repo and [run this test](https://github.com/alarbada/go-mysql/blob/76417ef684eb4e1493051a1e6168cbafc5724c92/bug_test.go#L51-L124). There's also docker commands to setup the mysql and mariadb instances.

mefcorvi:
In MariaDB 11.4+, some legitimate binlog events may have `LogPos=0` in their header, particularly inter-transaction events. The current code in `BinlogSyncer.handleEventAndACK()` unconditionally ignores all events with `LogPos=0`:

```go
if e.Header.LogPos > 0 {
    // Some events like FormatDescriptionEvent return 0, ignore.
    b.nextPos.Pos = e.Header.LogPos
}
```

This can cause issues when replicating from MariaDB 11.4+ because valid events are being ignored for position tracking.

## Proposed solution

Add a configuration parameter in binlogsyncer that enables dynamic calculation of LogPos for those events where we received 0. For this to work, the [BINLOG_SEND_ANNOTATE_ROWS_EVENT flag must be set when connecting to the database](https://mariadb.com/docs/server/reference/clientserver-protocol/replication-protocol/com_binlog_dump#note). This flag will cause the database to start sending ANNOTATE_ROWS_EVENT to the replica. This is necessary to correctly calculate the position of events because if ANNOTATE_ROWS_EVENTS are present in the binlog of the primary, they affect the position of other events. The flag itself will increase the traffic between the replica and the database, but otherwise, there should be no difference; MariadbAnnotateRowsEvent should be handled normally in go-mysql.

## References

- https://jira.mariadb.org/browse/MDEV-31273
- [MariaDB source: `sql/log_event_server.cc` - `Log_event::net_send()` method](https://github.com/MariaDB/server/blob/46d7450f752b8f9d782aa661a26d8c003849ea6b/sql/log_event_server.cc#L638)
- [MariaDB source: `sql/sql_repl.cc` - `mysql_show_binlog_events()` function](https://github.com/MariaDB/server/blob/46d7450f752b8f9d782aa661a26d8c003849ea6b/sql/sql_repl.cc#L4422)
- [MariaDB documentation on artificial events and the 0x20 flag](https://mariadb.com/docs/server/reference/clientserver-protocol/replication-protocol/2-binlog-event-header#event-flag)
- https://mariadb.com/docs/server/clients-and-utilities/logging-tools/mariadb-binlog/annotate_rows_log_event
- https://mariadb.com/docs/server/ha-and-performance/standard-replication/replication-and-binary-log-system-variables#binlog_annotate_row_events

## Fixes
#1035

## Related
#893


replication/binlogsyncer.go
```diff
@@ -861,6 +861,16 @@ func (b *BinlogSyncer) handleEventAndACK(s *BinlogStreamer, e *BinlogEvent, need
 	if e.Header.LogPos > 0 {
 		// Some events like FormatDescriptionEvent return 0, ignore.
 		b.nextPos.Pos = e.Header.LogPos
+	} else if b.cfg.Flavor == mysql.MariaDBFlavor && (e.Header.Flags&LOG_EVENT_ARTIFICIAL_F) == 0 && e.Header.LogPos == 0 {
+		// For MariaDB 11.4+, some events may have LogPos=0 but are still valid position updates
+		// if they are not artificial events (marked with LOG_EVENT_ARTIFICIAL_F flag).
```
dveeden: MySQL also has this flag: https://dev.mysql.com/doc/dev/mysql-server/9.3.0/group__group__cs__binglog__event__header__flags.html

So maybe this isn't MariaDB only?

For MariaDB this is documented on https://mariadb.com/docs/server/clients-and-utilities/server-client-software/client-libraries/clientserver-protocol/replication-protocol/2-binlog-event-header

mefcorvi:
> What do you think of using flags vs specific whitelisted events?

In my opinion, flags are a more reliable solution. New event types can be added to the database engine, and everything should remain stable as long as the primary database continues to send these events to the replicas.

mefcorvi:
@lance6716 @dveeden What do you think? I can add a new configuration parameter and update the documentation if you think it's a good idea. Also, I'd like to add some functional tests using snapshots of real network communications with the database.

replication/binlogsyncer.go
mefcorvi: I'd like to keep the scope of changes minimal and focus solely on MariaDB, as the original issue is caused by optimizations in MariaDB's binlog format. Additionally, as we've discussed below, I believe it would be better to introduce a new configuration parameter for those who wish to have LogPos even when the end_log_pos is 0 in the original binlog event.

dveeden:
How do the flags that are send to the MySQL/MariaDB server affect applications that use go-mysql? 

I think events that were previously ignored or not send to the client or had less data should be fine as either the application is already for specific events or should be made to accept many events (e.g. by logging it or throwing a warning).

For compatibility I'm more concerned by flags that are not accepted by certain server versions or server vendors.

I hope to have time to review this in more detail soon-ish, but I can't guarantee anything.

mefcorvi:
@dveeden Currently, MariaDB 11.4+ returns 0 as the end_log_pos (i.e., LogPos) for a number of events from the binlog, which may seem non-obvious to library users. I propose to add a configuration parameter in binlogsyncer that enables dynamic calculation of LogPos for those events where we received 0. For this to work, the BINLOG_SEND_ANNOTATE_ROWS_EVENT flag must be set when connecting to the database. This flag will cause the database to start sending ANNOTATE_ROWS_EVENT to the replica. The flag itself will increase the traffic between the replica and the database, but otherwise, there should be no difference; MariadbAnnotateRowsEvent should be handled normally in go-mysql.

Considering that all this logic will be hidden under a configuration flag, it should not affect those library users who do not need it.

mefcorvi:
> What happens if you have turned off annotations via the server variables? Nothing?

As far as I can see, there are several server variables in MariaDB related to ANNOTATE_ROWS_EVENTS:
• binlog_annotate_row_events, indicates to the primary whether to record ANNOTATE_ROWS_EVENTS in the binlog. If disabled, there will be no issues, events will not be in the primary, and they should not be sent to the replica.
• replicate_annotate_row_events, indicates that the replica should copy ANNOTATE_ROWS_EVENTS to its binlog. If I understand correctly, this flag tells the replica whether it should send the BINLOG_SEND_ANNOTATE_ROWS_EVENT flag when connecting to the primary. In the case of go-mysql, this setting should not affect anything.

However, this all needs to be tested thoroughly, and I currently do not have the opportunity to do so.

lance6716:
Thanks for contributing, I'll start review about 1~2 days later.

replication/binlogsyncer.go
```diff
@@ -128,6 +128,12 @@ type BinlogSyncerConfig struct {
 
 	EventCacheCount int
 
+	// MariaDBDynamicLogPos enables dynamic LogPos calculation for MariaDB.
+	// When enabled, automatically adds BINLOG_SEND_ANNOTATE_ROWS_EVENT flag
+	// to ensure correct position calculation in MariaDB 11.4+.
+	// Only works with MariaDB flavor.
+	MariaDBDynamicLogPos bool
```
lance6716, line 135: in `NewBinlogSyncer` we can check the flavor and `MariaDBDynamicLogPos` match, and print warning log if mismatch.

replication/binlogsyncer.go
```diff
@@ -509,7 +515,14 @@ func (b *BinlogSyncer) writeBinlogDumpCommand(p mysql.Position) error {
 	binary.LittleEndian.PutUint32(data[pos:], p.Pos)
 	pos += 4
 
-	binary.LittleEndian.PutUint16(data[pos:], b.cfg.DumpCommandFlag)
+	dumpCommandFlag := b.cfg.DumpCommandFlag
+	if b.cfg.MariaDBDynamicLogPos && b.cfg.Flavor == mysql.MariaDBFlavor {
+		// Add BINLOG_SEND_ANNOTATE_ROWS_EVENT flag when MariaDBDynamicLogPos is enabled.
+		// This ensures the server sends ANNOTATE_ROWS_EVENT events which are needed
+		// for correct LogPos calculation in MariaDB 11.4+, where some events have LogPos=0.
```
lance6716, line 522: Do you have some manual test result to make sure only this flag is enough? Please add some explanation in PR description.

lance6716 (approved):

replication/binlogsyncer.go
lance6716: How about `MariaDBFillZeroLogPos`? `Dynamic` makes me think more LogPos will be computed on-the-fly other than zero positions. And maybe the value is changeable, not the same one as upstream binlog position.

replication/binlogsyncer.go
nickchomey: This has been my ongoing concern and feedback - that this might not be the only necessary flag, and likewise that perhaps using event types explicitly would be a better approach. It is too "hand wavey" to just use this flag without a more thorough rationale. I don't have much to contribute to resolving this, however.

lance6716: Don't worry. I think because this fix indeed fixed your problem, we can merge it and leave more enhancement to future PR.

And because it's possible (not likely, though) we will change to another implementation and break properties from this PR, can you add some unit tests to prevent it breaking unintentionally?

nickchomey: I havent yet tested the current iteration. In fact, I only had it working when I used my own event type implementation rather than flags. 

I'll test it tomorrow and report back

lance6716:
ping @mefcorvi @nickchomey 

Just following up – is there an update on this? No worries at all if you're busy, I'm happy to push the merge forward myself.

PR #1053, opened 2025-08-15
change FillZeroLog config name and add to canal

nickchomey:


lance6716 (commented):
README should also be changed. You can search for `MariaDBDynamicLogPos` in the project

lance6716 (approved):

PR #1057, opened 2025-10-14
fix parsing prepare response

serprex:
https://dev.mysql.com/doc/dev/mysql-server/9.3.0/page_protocol_com_stmt_prepare.html

warnings is optional from server, & is preceded by a reserved byte

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -269,6 +269,7 @@ func (s *clientTestSuite) testStmt_DropTable() {
 
 	stmt, err := s.c.Prepare(str)
 	require.NoError(s.T(), err)
+	require.Zero(s.T(), stmt.WarningsNum())
 
 	defer stmt.Close()
 


client/stmt.go
```diff
@@ -265,9 +265,14 @@ func (c *Conn) Prepare(query string) (*Stmt, error) {
 	s.params = int(binary.LittleEndian.Uint16(data[pos:]))
 	pos += 2
 
-	// warnings
-	s.warnings = int(binary.LittleEndian.Uint16(data[pos:]))
-	// pos += 2
+	// reserved
+	pos += 1
```
lance6716, line 269: I just wonder why the old code behaves wrong but user didn't report it 😢 Maybe they didn't check the warning counts? 

There are already some unit tests using PREPARE, can you modify one case to check `func (s *Stmt) WarningsNum() int` is correct?

lance6716 (commented):
lgtm

client/stmt.go
serprex: Will do tomorrow

serprex: reserved byte is 0 so would usually be 0 warnings

Issue #1056, opened 2025-10-14
querying with parameters hangs on mysql 8.0.25


serprex:
Reproducible with: https://github.com/danielrmz/peerdb-mysql-test (run `docker-compose up mysql25` & `go run main.go`)

Debugging, gets stuck waiting on reading results, after having received column info

dveeden:
Any packet traces, logging or other details that might help?

serprex:
Seems this is a mysql bug: 8.0.25 advertises a bunch of bits it doesn't actually support, which is troublesome with `CLIENT_QUERY_ATTRIBUTES`

also, https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-26.html

> Including query attributes for a prepared statement could cause a statement execution failure. (Bug #32676878)

Issue #980, opened 2025-02-06
Set CLIENT_DEPRECATE_EOF by default


dveeden:
The `CLIENT_DEPRECATE_EOF`  is set by many client and server implementations. We should do this as well.

MySQL has changed this in 5.7.5

dveeden:
@serprex we can close this now, right?

PR #1059, opened 2025-10-14
CLIENT_DEPRECATE_EOF

serprex:
Taking a stab at #980

Part of investigating #1056 

Something I noticed was that server capabilities aren't limited by what's sent as client capabilities. Reading [docs](https://dev.mysql.com/doc/dev/mysql-server/9.4.0/group__group__cs__capabilities__flags.html) it seems like both ends are expected to look at the intersection of each others flags. So I decided to tweaks things so that `c.capabilities` is that intersection, rather than whatever server sends

Seems to be working with latest mysql, but on 8.0.25 instead of a hang I get an empty result set back (while `go-sql-driver/mysql` still pulls the non-empty result)

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -466,7 +466,7 @@ func (c *Conn) FieldList(table string, wildcard string) ([]*mysql.Field, error) 
 		}
 
 		// EOF Packet
-		if data[0] == mysql.EOF_HEADER && len(data) <= 0xffffff {
+		if c.isEOFPacket(data) {
 			return fs, nil
 		}
 
--- a/client/resp.go
+++ b/client/resp.go
@@ -14,9 +14,11 @@ import (
 	"github.com/go-mysql-org/go-mysql/utils"
 )
 
-// this should only be called when CLIENT_DEPRECATE_EOF not enabled
 func (c *Conn) isEOFPacket(data []byte) bool {
-	return data[0] == mysql.EOF_HEADER && len(data) <= 5
+	// 0xffffff due to https://dev.mysql.com/worklog/task/?id=7766
+	// "Server will never send OK packet longer than 16777216 bytes thus limiting
+	// size of OK packet to be 16777215 bytes"
+	return data[0] == mysql.EOF_HEADER && len(data) <= 0xffffff
 }
 
 func (c *Conn) handleOKPacket(data []byte) (*mysql.Result, error) {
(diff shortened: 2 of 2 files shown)


client/conn.go
```diff
@@ -466,7 +466,7 @@ func (c *Conn) FieldList(table string, wildcard string) ([]*mysql.Field, error)
 		}
 
 		// EOF Packet
-		if c.isEOFPacket(data) {
+		if data[0] == mysql.EOF_HEADER && len(data) <= 0xffffff {
```
lance6716: Can you explain why we change it to `len(data) <= 0xffffff`?

client/resp.go
```diff
@@ -336,33 +321,16 @@ func (c *Conn) readResultsetStreaming(data []byte, binary bool, result *mysql.Re
 }
 
 func (c *Conn) readResultColumns(result *mysql.Result) (err error) {
-	i := 0
 	var data []byte
 
-	for {
+	for i := range len(result.Fields) {
```
lance6716:
```suggestion
	for i := range result.Fields {
```

client/conn.go
serprex: Will add comment later, bit unavailable this week

https://dev.mysql.com/worklog/task/?id=7766

> In case of huge data packet with length greater than 16777216L client will treat it as a data packet and process accordingly.

serprex:
https://bugs.mysql.com/bug.php?id=91533

> This seems to contradict mysql documentation which states that client should not enable flags not supported by server.

Maybe masking added here should also apply to what we send back to server

client/resp.go
```diff
@@ -14,24 +14,11 @@ import (
...
-		// EOF Packet
-		if c.isEOFPacket(data) {
-			return err
-		}
-	}
-}
-
+// 0xffffff due to https://dev.mysql.com/worklog/task/?id=7766
+// "Server will never send OK packet longer than 16777216 bytes thus limiting
+// size of OK packet to be 16777215 bytes"
```
lance6716: move these comments above line 21? Because `0xffffff` is only used inside `isEOFPacket`, we can limit the comment scope.

client/stmt.go
```diff
@@ -275,14 +275,33 @@ func (c *Conn) Prepare(query string) (*Stmt, error) {
...
+			} else if !c.isEOFPacket(packet) {
+				return nil, mysql.ErrMalformPacket
+			}
 		}
 	}
 
 	if s.columns > 0 {
-		if err := s.conn.readUntilEOF(); err != nil {
-			return nil, errors.Trace(err)
+		// TODO process when CLIENT_CACHE_METADATA enabled
```
lance6716, line 293: Do you mean `MARIADB_CLIENT_CACHE_METADATA`? https://mariadb.com/docs/server/reference/clientserver-protocol/mariadb-protocol-differences-with-mysql

I want to understand what's the effect after this PR. Hope it will not break existing program with that capabilities flag.

serprex: We don't currently support those flags, so it won't affect. We have code that reads until eof to skip info that certain options would omit from protocol

PR #1061, opened 2025-10-18
Checks if there is an error in the streamer.

oleksiivelychko:
Hi there. I have a scenario when I'd like to stop my app after this error appears in the logs:
```
[error] binlogstreamer.go:78 close sync with err: ERROR 1236 (HY000): Could not find first log file name in binary log index file
```
I see **BinlogStreamer** has all the necessary private fields, but I can't access them from the outside:
```go
type BinlogStreamer struct {
	ch  chan *BinlogEvent
	ech chan error
	err error
}
```
So, as a possible solution I implemented a new one public method (getter) to fetch the actual state of streamer:
```go
func (s *BinlogStreamer) IsErrorInStreamer() bool {
	select {
	case err := <-s.ech:
		return err != nil
	default:
		return false
	}
}
```
There maybe a more appropriate approach to achieve this goal, but this one seems good to me. What do you think?
Thanks.

### DIFF
--- a/replication/binlogstreamer.go
+++ b/replication/binlogstreamer.go
@@ -118,3 +118,13 @@ func (s *BinlogStreamer) AddErrorToStreamer(err error) bool {
 		return false
 	}
 }
+
+// IsErrorInStreamer checks if there is an error in the streamer.
+func (s *BinlogStreamer) IsErrorInStreamer() bool {
+	select {
+	case err := <-s.ech:
+		return err != nil
+	default:
+		return false
+	}
+}


replication/binlogstreamer.go
```diff
@@ -118,3 +118,13 @@ func (s *BinlogStreamer) AddErrorToStreamer(err error) bool {
 		return false
 	}
 }
+
+// IsErrorInStreamer checks if there is an error in the streamer.
+func (s *BinlogStreamer) IsErrorInStreamer() bool {
+	select {
+	case err := <-s.ech:
```
lance6716, line 125: You can use 

https://github.com/go-mysql-org/go-mysql/blob/dec4081107ac185374622ef80da6c062a333117b/replication/binlogstreamer.go#L22-L33

And this new method will interfere with `GetEvent` because err maybe taken from `.ech`

oleksiivelychko: Yes, I checked it, and it works too. I've seen this method before I created PR, but one detail makes me wary: `GetEvent gets the binlog event one by one,` where my goal is to check for a non-existent binlog and stop my app for that, but in this case if my binlog file exists it will start the binlog events processing, isn't it?
Also, you mentioned _**And this new method will interfere with GetEvent because err maybe taken from .ech**_, if this a new method it will not affect the previous logic in the package, right? I assumed I would get such an error before starting my app logic, and would stop it if it happened.

lance6716: > where my goal is to check for a non-existent binlog and stop my app for that, but in this case if my binlog file exists it will start the binlog events processing, isn't it?

Oh yes, in your case `GetEvent` has some side effects and is not optimal. But I didn't find other good way to achieve your goal. I think set `EventCacheCount` to 1

https://github.com/go-mysql-org/go-mysql/blob/dec4081107ac185374622ef80da6c062a333117b/replication/binlogsyncer.go#L423

and try to read one binlog event / error is a good solution. Because binlog streamer is started at background at anytime afterwards, it may even not start when you check `IsErrorInStreamer`.


> it will not affect the previous logic in the package

Reading from `.ech` is exported to all developers, so maybe someone will call it while another goroutine expects all errors should be returned from `GetEvent`.

oleksiivelychko: I tried changing `EventCacheCount` but I think it's not a suitable option here, since changing the capacity from 10240 to 1 can affect the performance. In general, as a workaround I could use the following method:
```go
func (s *MyBinlogStreamer) AddEventToStreamer(ev *replication.BinlogEvent) error {
	select {
	case s.ch <- ev:
		return nil
	case err := <-s.ech:
		return err
	}
}
```
... where I send nil value of `BinlogEvent` and get the expected error.
Also, as an option I could create a new `s := NewBinlogStreamerWithChanSize(1)` before running my event handling logic and check for this error, and if it exists then stop my app, otherwise close this new streamer and continue executing the main flow.
Also, as an alternative I could completely override `BinlogStreamer` and `BinlogSyncer` adding this new method.

lance6716: Okay, these workarounds seems not good. I'm thinking that you can modify your PR to ...

```golang
func (s *BinlogStreamer) IsErrorInStreamer() bool {
  return len(s.ech) > 0
}
```

... with function comment to explain that `IsErrorInStreamer` returns the internal `s.ech` status and that channel is affected by delayed goroutine running or other functions that consume the channel, so caller is not guaranteed to see the expected result.

And can you provide code snippet of your usage? I think `IsErrorInStreamer` is really not helpful in the case of delayed goroutine running. Just want to understand your code.

oleksiivelychko: Thanks for your help, I decided to implement it through a new one separated `NewBinlogSyncer` and check that error with `GetEvent` method.

PR #1062, opened 2025-10-20
Add custom authentication

desdic:
I found a lot of references to `mysql_clear_password` but the handling was missing

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -134,6 +134,10 @@ type BinlogSyncerConfig struct {
 	// Only works with MariaDB flavor.
 	FillZeroLogPos bool
 
+	// PayloadDecoderConcurrency is used to control concurrency for decoding TransactionPayloadEvent.
+	// Default 0,  this will be set to GOMAXPROCS.
+	PayloadDecoderConcurrency int
+
 	// SynchronousEventHandler is used for synchronous event handling.
 	// This should not be used together with StartBackupWithHandler.
 	// If this is not nil, GetEvent does not need to be called.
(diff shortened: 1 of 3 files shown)


server/auth.go
```diff
@@ -102,6 +106,16 @@ func scrambleValidation(cached, nonce, scramble []byte) bool {
 	return subtle.ConstantTimeCompare(m, cached) == 1
 }
 
+func (c *Conn) compareClearPasswordAuthData(clientAuthData []byte, credential Credential) error {
+	clearText := bytes.Trim(clientAuthData, "\x00")
```
dveeden: Maybe this should be `bytes.TrimRight()`? What if there are multiple `\x00`, like in the middle of the `clientAuthData`?

dveeden:
Test failures are due to a Docker outage: https://www.dockerstatus.com/

server/auth.go
desdic: ah yes fixed

dveeden:
1. Yes making it easy to do custom auth would be good.
2. Note that with `caching_sha2_password` over TLS without a cached entry you also get a plaintext password. This might not be usable for you and is probably a bit too much of a hack. But this may have better/improved compatibility with clients. Just so you know.

dveeden:
> making customer auth would also be the only way for a user of this library to get clientAuthData in raw form right ?

Yes, that's probably correct.

desdic:
Oki hopefully last commit. I've added the authentication as an interface with a default one so behavior is the same as before but its possible to implement you own authentication and plugin validation

desdic:
is my latest patch better ?

lance6716 (approved):

Issue #1071, opened 2025-12-05
Authentication hooks for post-authentication logic before OK packet


ramnes:
Hey again!

I'd like to be able to execute custom logic after authentication succeeds or fails but before the packet is sent to the client: rejecting connections based on authorization rules, logging authentication failure, etc.

Right now, by the time `NewCustomizedConn` returns, the client has already received the authentication / handshake OK packet, so I can only reject on the first command rather than during the handshake.

What would you think of optional authentication hooks, e.g. `OnAuthSuccess` and `OnAuthFailure`? 

We could either pass them to `NewCustomizedConn`, or replace `CredentialProvider` with a new interface that has a larger scope, e.g. something like `AuthenticationHandler`.

What do you think? Happy to discuss alternative approaches. :)

ramnes:
A larger `AuthHandler` interface could even bundle other changes for the issues discussed in #1069 and #1070:

```go
// AuthHandler provides user credentials and optional lifecycle hooks.
// This interface replaces CredentialProvider with additional hooks for
// executing logic after authentication completes but before the OK/error
// packet is sent to the client.
type AuthHandler interface {
	// CheckUsername checks if the user exists.
	CheckUsername(username string) (found bool, err error)

	// GetCredentials returns the user's credentials.
	// Supports multiple valid credentials per user (e.g., password rotation).
	GetCredentials(username string) (credentials []Credential, found bool, err error)  // <-- Addresses #1069 

	// OnAuthSuccess is called after successful authentication, before the OK packet.
	// Return an error to reject the connection (error will be sent to client instead of OK).
	// Return nil to proceed with sending the OK packet.
	OnAuthSuccess(conn *Conn) error

	// OnAuthFailure is called when local authentication fails.
	// Return a backend connection to retry authentication against the backend
	// server using an auth switch request.
	// Return nil to reject the connection with the original error.
	OnAuthFailure(conn *Conn, err error) (*client.Conn, error)  // <-- Addresses #1070 (maybe?)
}
```

lance6716:
Can the new methods be put inside `Handler` (I didn't implement a proxy using this package so have no taste)? I'm thinking of putting hooks together or putting authentication functions together. Will think about it later.

Issue #1069, opened 2025-12-03
Multiple valid passwords per user in `CredentialProvider`


ramnes:
Hey there, thanks for great library!

Currently, `CredentialProvider.GetCredential` returns a single password per username. In some use cases, a user may have multiple valid credentials (e.g., API tokens, rotating secrets) that should all be accepted during authentication.

Would you be open to replace `GetCredential` by a new `GetCredentials` method that returns a slice of valid credentials?

Another approach if we don't want to break existing implementations of the interface would be to add a new `MultiCredentialProvider` interface that embeds `CredentialProvider` and adds that new method.

Then in the authentication flow, we could check if the provider implements MultiCredentialProvider via type assertion:

```go
  if mcp, ok := provider.(MultiCredentialProvider); ok {
      passwords, found, err := mcp.GetCredentials(username)
      // ...
  } else {
      password, found, err := provider.GetCredential(username)
      // ...
  }
```

Happy to submit a PR if any of the two approaches sounds reasonable.

ramnes:
If we're ready to break API compatibility by updating `Credential`, I would just replace `GetCredential` with `GetCredentials`, what do you think? That's probably the simplest approach here, and I already have a PR ready for it actually. :)

On the `PasswordVerifier` idea, I'm not sure to understand what you have in mind. What would be an example implementation and why would we keep `AuthPluginName`?

PS - `MultiCredentialProvider` was meant for "multiple credentials of one user". Maybe something like `CredentialsProvider` (note the `s` at the end of `Credentials`) would be easier to read if we want to go that route.

PR #1072, opened 2025-12-06
Implement `AuthenticationHandler` for custom auth mechanisms

ramnes:
Hey team,

Here is some work around #1069 and #1071 to allow a discussion based on an actual implementation.

I think the changes here are pretty obvious:
* Renames;
* A `Credential` can have multiple passwords now;
* Added `OnAuthSuccess` and `OnAuthFailure` to the new `AuthenticationHandler`.

One side effect of the multiple passwords per user is that we can't have `Xor` update inplace anymore, because otherwise we would XOR the password we received every time we compare it to a new hash in the list of hashes to try.

About hashes – I moved the `Credential` passwords hash at comparison-time to avoid paying the full price for every connection. Let's say you have 1000 passwords to match against but match after 5 comparisons: it's better to hash only 5 passwords, not 1000.

### DIFF
--- a/server/server_conf.go
+++ b/server/server_conf.go
@@ -103,7 +103,7 @@ func NewServerWithAuth(serverVersion string, collationId uint8, defaultAuthMetho
 }
 
 func isAuthMethodSupported(authMethod string) bool {
-	return authMethod == mysql.AUTH_NATIVE_PASSWORD || authMethod == mysql.AUTH_CACHING_SHA2_PASSWORD || authMethod == mysql.AUTH_SHA256_PASSWORD
+	return authMethod == mysql.AUTH_NATIVE_PASSWORD || authMethod == mysql.AUTH_CACHING_SHA2_PASSWORD || authMethod == mysql.AUTH_SHA256_PASSWORD || authMethod == mysql.AUTH_CLEAR_PASSWORD
 }
 
 func (s *Server) InvalidateCache(username string, host string) {


ramnes:
Yeah, the PR is a bit too large for my taste too. :)

I didn’t mean for it to be merged as-is, more as a starting point for discussion around the issues I opened. Happy to split it up once we have a clearer idea of what we want to do for each one!

server/authentication_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+type Credential struct {
+	Passwords      []string // raw passwords, hashed on demand during comparison
+	AuthPluginName string
+}
+
+// HashPassword computes the password hash for a given password using the credential's auth plugin.
+func (c Credential) HashPassword(password string) (string, error) {
```
lance6716: seems it can be unexported

server/authentication_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	case mysql.AUTH_CLEAR_PASSWORD:
+		return password, nil
+
+	default:
+		return "", errors.Errorf("unknown authentication plugin name '%s'", c.AuthPluginName)
+	}
+}
+
+// HasEmptyPassword returns true if any password in the credential is empty.
+func (c Credential) HasEmptyPassword() bool {
```
lance6716: ditto

server/authentication_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, p := range c.Passwords {
```
lance6716: Can use https://pkg.go.dev/slices#Contains now

server/authentication_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	defaultAuthMethod string
+}
+
+func (h *InMemoryAuthenticationHandler) CheckUsername(username string) (found bool, err error) {
+	_, ok := h.userPool.Load(username)
+	return ok, nil
+}
+
+func (h *InMemoryAuthenticationHandler) GetCredential(username string) (credential Credential, found bool, err error) {
+	v, ok := h.userPool.Load(username)
```
lance6716, line 94: please add comment for AuthenticationHandler.GetCredential that it should be concurrent safe

server/authentication_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	return c, true, nil
+}
+
+func (h *InMemoryAuthenticationHandler) AddUser(username, password string, optionalAuthPluginName ...string) error {
+	authPluginName := h.defaultAuthMethod
+	if len(optionalAuthPluginName) > 0 {
+		authPluginName = optionalAuthPluginName[0]
+	}
+
+	if !isAuthMethodSupported(authPluginName) {
```
lance6716, line 111: seems in the old code we also support AUTH_CLEAR_PASSWORD. @dveeden Should we change `isAuthMethodSupported` to add that?

lance6716 (commented):
6 / 11 viewed, will review later

server/authentication_handler.go
ramnes: indeed :)

server/authentication_handler.go
ramnes: done

server/authentication_handler.go
ramnes: done!

server/authentication_handler.go
ramnes: I can also remove this check entirely if it doesn't make sense in that context, as there was no check before. Feel free to tell me.

server/auth.go
```diff
@@ -30,15 +30,15 @@ func (c *Conn) compareAuthData(authPluginName string, clientAuthData []byte) err
...
+	if len(c.credential.Passwords) > 0 {
 		return nil
 	}
-	credential, found, err := c.credentialProvider.GetCredential(c.user)
+	credential, found, err := c.authHandler.GetCredential(c.user)
 	if err != nil {
 		return err
 	}
-	if !found {
+	if !found || len(credential.Passwords) == 0 {
```
lance6716, line 41: We should add comments to `Credential` type that, if a user has no password, empty password, or one authentication method is empty password, `Credential.Passwords` should be `[]string{"", ...}` rather than zero length slice

server/auth.go
```diff
@@ -30,15 +30,15 @@ func (c *Conn) compareAuthData(authPluginName string, clientAuthData []byte) err
 	return c.serverConf.authProvider.Authenticate(c, authPluginName, clientAuthData)
 }
 
-func (c *Conn) acquirePassword() error {
-	if c.credential.Password != "" {
+func (c *Conn) acquireCredential() error {
+	if len(c.credential.Passwords) > 0 {
```
lance6716, line 34: Seems we can also use hasEmptyPassword here?

server/authentication_handler.go
lance6716: I think it's better to remove the check. And let's wait @dveeden until all comments are resolved to understand if `isAuthMethodSupported` should also be updated

dveeden: I think we probably should support `AUTH_CLEAR_PASSWORD`.

Note that this is a [client side plugin](https://dev.mysql.com/doc/refman/8.4/en/cleartext-pluggable-authentication.html). 

The usecase for this is to have the client send the cleartext password to the server which then allows the server to use this to authenticate against LDAP (via binding, a hash stored in ldap doesn't need this) or via PAM or anything else that is custom.

Note that on the client one needs to use `--enable-cleartext-plugin` or use set it via an option in their apps.

The risk with cleartext is obvious. It should be used only over TLS connections.

Note that `caching_sha2_password` has two options:
- RSA keypair
- TLS connection.

With a RSA keypair the public key has to be specified or you have to enable an option to fetch it from the server (which is insecure). I don't think this should be used.

With `caching_sha2_password` and an empty cache the password is transmitted in cleartext over a TLS connection. We should make sure we don't accept it over a non-TLS connection (except for UNIX domain socket)

dveeden (commented):
Seems ok on first glance.

dveeden:
Could we mix authentication methods? E.g. a user having two authentication methods:

1. Caching_sha2_password with hash abcdefg
2. Auth_socket linked to username jdoe

This could be useful for migrating from one to another. Or when a authmethod has limitations (like auth_socket). Or it could be used for MFA

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_multi_factor_authentication_methods.html

ramnes:
I’m not entirely sure, but I don’t think this is feasible. As discussed in #1069:

> Playing a little more with this, I think the only way to allow for multiple passwords is actually to keep GetCredential as-is, but instead modify Credential so that it holds multiple passwords with a Passwords []string field.
> 
> My understanding of the MySQL protocol is that it's impossible to handle authentication against multiple auth plugins for a single user on the wire, even if the database implementation would support it.

For additional context: I previously tried to implement a server that stored credentials for both `mysql_native_password` and `caching_sha2_password`, with the goal of allowing *any* client to connect: older clients that only support `mysql_native_password` (and may lack newer auth capabilities), as well as newer clients that only support `caching_sha2_password`. I couldn’t make this work reliably.

From what I remember of these experiments, the only potentially viable strategy was for the server to announce `mysql_native_password` in the initial handshake and then request an auth switch to `caching_sha2_password` when needed, since the server speaks first in the MySQL handshake / authentication protocol. However, IIRC, when a client literally couldn't perform native password authentication (e.g. because it doesn’t have the plugin at all, such as the client installed with `brew install mysql`), the connection was failing before an `AuthSwitchRequest` could even be processed. The client couldn't be "rescued" by switching.

Given that, I’d prefer not to allow multiple authentication plugin names per user if we can’t guarantee that this will work end-to-end. Exposing this would likely be misleading for developers playing with `go-mysql`.

MFA is out-of-scope for me here.

server/authentication_handler.go
ramnes: @dveeden Do you mean we should add `AUTH_CLEAR_PASSWORD` in `isAuthMethodSupported`, or outside of it? Do you think we should keep this check here?

server/auth.go
ramnes: Hm, I'm not sure to understand why you would want to use `hasEmptyPassword` here.

We can't replace `len(c.credential.Passwords) > 0` with `c.credential.hasEmptyPassword`.

The only scenario I could imagine would be to make the check more explicit with something like

```go
if len(c.credential.Passwords) > 0 || c.credential.hasEmptyPassword() {
	return nil
}
```

but then it would be redundant, because if `hasEmptyPassword()` returns true, then `len(Passwords) >)` is true as well.

What do you have in mind?

server/auth.go
ramnes: added!

server/authentication_handler.go
lance6716: seems dveeden is busy these days. I think you can add AUTH_CLEAR_PASSWORD in isAuthMethodSupported

ramnes: done

PR #1073, opened 2025-12-06
Store prepared statements field definitions

ramnes:
Hey,

When a MySQL client uses prepared statements (i.e. `COM_STMT_PREPARE`/`COM_STMT_EXECUTE`), the server sends column and parameter field definitions in the response, and clients rely on this metadata.

Currently, go-mysql's client discards these field definitions after parsing, and only stores the counts. When used in a proxy scenario, the server would use the client and send the response back, meaning that the downstream client would end up with the current dummy field definitions, which cause issues (e.g., deserialization issue due to a varchar that's interpreted as a decimal.)

This PR enables proxy implementations to transparently pass through prepared statement metadata easily.

### DIFF
--- a/server/stmt_test.go
+++ b/server/stmt_test.go
@@ -90,6 +90,10 @@ func TestStmtPrepareWithPreparedStmt(t *testing.T) {
 	st := result.(*Stmt)
 	require.NotNil(t, st.RawParamFields)
 	require.NotNil(t, st.RawColumnFields)
-	require.Equal(t, mysql.MYSQL_TYPE_LONG, st.GetParamFields()[0].Type)
-	require.Equal(t, mysql.MYSQL_TYPE_LONGLONG, st.GetColumnFields()[0].Type)
+	paramFields, err := st.GetParamFields()
+	require.NoError(t, err)
+	require.Equal(t, mysql.MYSQL_TYPE_LONG, paramFields[0].Type)
+	columnFields, err := st.GetColumnFields()
+	require.NoError(t, err)
+	require.Equal(t, mysql.MYSQL_TYPE_LONGLONG, columnFields[0].Type)
 }
--- a/stmt/stmt.go
+++ b/stmt/stmt.go
@@ -14,30 +14,38 @@ type PreparedStmt struct {
 	columnFields []*mysql.Field
 }
 
-func (s *PreparedStmt) GetParamFields() []*mysql.Field {
+func (s *PreparedStmt) GetParamFields() ([]*mysql.Field, error) {
 	if s.RawParamFields == nil {
-		return nil
+		return nil, nil
 	}
 	if s.paramFields == nil {
-		s.paramFields = make([]*mysql.Field, len(s.RawParamFields))
+		fields := make([]*mysql.Field, len(s.RawParamFields))
 		for i, raw := range s.RawParamFields {
-			s.paramFields[i] = &mysql.Field{}
-			_ = s.paramFields[i].Parse(raw)
+			field := &mysql.Field{}
+			if err := field.Parse(raw); err != nil {
+				return nil, err
+			}
+			fields[i] = field
 		}
+		s.paramFields = fields
 	}
-	return s.paramFields
+	return s.paramFields, nil
 }
 
-func (s *PreparedStmt) GetColumnFields() []*mysql.Field {
+func (s *PreparedStmt) GetColumnFields() ([]*mysql.Field, error) {
 	if s.RawColumnFields == nil {
-		return nil
+		return nil, nil
 	}
 	if s.columnFields == nil {
-		s.columnFields = make([]*mysql.Field, len(s.RawColumnFields))
+		fields := make([]*mysql.Field, len(s.RawColumnFields))
 		for i, raw := range s.RawColumnFields {
-			s.columnFields[i] = &mysql.Field{}
-			_ = s.columnFields[i].Parse(raw)
+			field := &mysql.Field{}
+			if err := field.Parse(raw); err != nil {
+				return nil, err
+			}
+			fields[i] = field
 		}
+		s.columnFields = fields
 	}
-	return s.columnFields
+	return s.columnFields, nil
 }


client/stmt.go
```diff
@@ -19,6 +19,10 @@ type Stmt struct {
 	params   int
 	columns  int
 	warnings int
+
+	// Field definitions from the PREPARE response (for proxy passthrough)
+	ParamFields  []*mysql.Field
```
lance6716: no need to be public? (change it to `paramFields` and `columnFields`)

client/stmt.go
```diff
@@ -33,6 +37,18 @@ func (s *Stmt) WarningsNum() int {
 	return s.warnings
 }
 
+// GetParamFields returns the parameter field definitions from the PREPARE response.
+// Implements server.StmtFieldsProvider for proxy passthrough.
+func (s *Stmt) GetParamFields() []*mysql.Field {
```
lance6716: please add comment that should not modify the returned values.

server/stmt.go
```diff
@@ -61,7 +65,11 @@ func (c *Conn) writePrepare(s *Stmt) error {
 	if s.Params > 0 {
 		for i := 0; i < s.Params; i++ {
 			data = data[0:4]
-			data = append(data, paramFieldData...)
+			if s.ParamFields != nil && i < len(s.ParamFields) {
+				data = append(data, s.ParamFields[i].Dump()...)
```
lance6716: Seems no need to parse it? We can only forward the raw value.

server/command.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/go-mysql-org/go-mysql/utils"
 )
 
+// StmtFieldsProvider is an optional interface that prepared statement contexts can implement
+// to provide field definitions for proxy passthrough scenarios.
+type StmtFieldsProvider interface {
```
lance6716: Just to verify why we need this interface? If the only reason is to cast `st.Context`, I think maybe we should move the common struct of client and server package to a common package, like utils or a new package, to avoid a runtime cast.

client/stmt.go
ramnes: Makes sense. Do you want to return a copy instead?

client/stmt.go
ramnes: Yeah, we can make these private! Done.

server/stmt.go
ramnes: The rationale was that for a high-level library that users can use to inspect fields metadata, it makes sense that they don't have to parse the bytes themselves. But for a pure proxy use-case, yeah, the raw value would be better. Both work for me!

server/command.go
ramnes: Are you suggesting that we create something like

```
type StmtFields struct {
    ParamFields []*Field
    ClumnFields []*Field
}
```

and then use it in `client.Stmt`?

client/stmt.go
ramnes: Added the comments, feel free to tell me if you prefer a copy instead. :)

ramnes:
@lance6716 Do you like it better that way? :)

ramnes:
@lance6716 After playing with it a bit more, one problem with our approach here is that a proxy that wants to leverage the transmitted raw fields has to do something like this:

```go
type Handler struct {
    backendConn *client.Conn
    stmts       map[uint32]*client.Stmt
}

func (h *Handler) HandleStmtPrepare(query string) (int, int, any, error) {
    s, err := h.backendConn.Prepare(query)
    if err != nil {
        return 0, 0, nil, err
    }
    if h.stmts == nil {
        h.stmts = make(map[uint32]*client.Stmt)
    }
    
    // store the client.Stmt so that we can reuse it for Execute / Close later
    h.stmts[s.ID] = s
    
    // return a stmt.PreparedStmt so that its raw fields are copied
    // (this is not obvious at all, unless you check go-mysql source code)
    return s.ParamNum(), s.ColumnNum(), &s.PreparedStmt, nil
}

func (h *Handler) HandleStmtExecute(ctx any, query string, args []any) (*mysql.Result, error) {
    ps := ctx.(*stmt.PreparedStmt)
    s := h.stmts[ps.ID]
    return s.Execute(args...)
}

func (h *Handler) HandleStmtClose(ctx any) error {
    if ctx != nil {
        ps := ctx.(*stmt.PreparedStmt)
        if s, ok := h.stmts[ps.ID]; ok {
            delete(h.stmts, ps.ID)
            return s.Close()
        }
    }
    return nil
}
```

If we checked against `client.Stmt` or used an interface instead, we could make this code much simpler, because we wouldn't need `HandleStmtPrepare` to return a `stmt.PreparedStatement` specifically and would still forward the raw data.

```diff
type Handler struct {
    backendConn *client.Conn
-   stmts       map[uint32]*client.Stmt
}

func (h *Handler) HandleStmtPrepare(query string) (int, int, any, error) {
    s, err := h.backendConn.Prepare(query)
    if err != nil {
        return 0, 0, nil, err
    }
-   if h.stmts == nil {
-       h.stmts = make(map[uint32]*client.Stmt)
-   }
-   
-   // store the client.Stmt so that we can reuse it for Execute / Close later
-   h.stmts[s.ID] = s
-    
-   // return a stmt.PreparedStmt so that its raw fields are copied
-   // (this is not obvious at all, unless you check go-mysql source code)
-   return s.ParamNum(), s.ColumnNum(), &s.PreparedStmt, nil
+   return s.ParamNum(), s.ColumnNum(), s, nil
}

func (h *Handler) HandleStmtExecute(ctx any, query string, args []any) (*mysql.Result, error) {
-   ps := ctx.(*stmt.PreparedStmt)
-   s := h.stmts[ps.ID]
+   s := ctx.(*client.Stmt)
    return s.Execute(args...)
}

func (h *Handler) HandleStmtClose(ctx any) error {
    if ctx != nil {
-       ps := ctx.(*stmt.PreparedStmt)
-       if s, ok := h.stmts[ps.ID]; ok {
-           delete(h.stmts, ps.ID)
-           return s.Close()
-       }
+       s := ctx.(*client.Stmt)
+       return s.Close()
    }
    return nil
}
```

I feel like importing `client.Stmt` and adding a check against it is the simplest thing to do here, but we might end up with a circular import at some point down the line; which I think is why I ended up with an interface in the first place!

PR #1074, opened 2025-12-06
mysql,client: add log for caps we can't set

dveeden:
Related to: https://github.com/go-mysql-org/go-mysql/pull/1067#issuecomment-3603104244

```go
	h.conn, err = client.Connect("127.0.0.1:3306", "root", "", "", func(conn *client.Conn) error {
		conn.SetCapability(mysql.CLIENT_SESSION_TRACK)
		return nil
	})
```

As we don't support `mysql.CLIENT_SESSION_TRACK` the `SetCapability()` call above didn't result in the flag being set in the network packet. This should make that a bit more obvious.

Other things that we could do:
- Have `SetCapability()` return an error
- Have `SetCapability()` emit a log message
- Have `writeAuthHandshake()` return an error instead of emitting a log message.


lance6716:
I prefer we choose "Have SetCapability() return something", because it's earlier than the logging in this PR and can save some time for developers to locate the root cause. Maybe return an `error` (so some linters can tell user to check this error) or `bool` (like `ok` value from map access or type cast)? Will take a look later.

dveeden:
Some of this code was re-used for #1079

PR #1078, opened 2025-12-10
Send `using password: NO` when no password is used

ramnes:
Without this PR, the `using password: NO` appears when the user password on the server is empty. It should be the opposite: the `YES` or `NO` should depend on the data sent by the client, not the password stored for the user.


server/auth_switch_response.go
```diff
@@ -75,7 +75,7 @@ func (c *Conn) checkSha2CacheCredentials(clientAuthData []byte, credential Crede
 	if match && err == nil {
 		return nil
 	}
-	return errAccessDenied(credential)
+	return ErrAccessDenied
```
lance6716, line 85: in this call path, we didn't check no password case 🤔 I'm not familiar with it. Can you give more details?

ramnes: Good catch! Not sure why I skipped this one. Fixed it and added a test as well.

server/auth_switch_response_test.go
```diff
@@ -0,0 +1,43 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckSha2CacheCredentials_EmptyPassword(t *testing.T) {
```
lance6716, line 9: Can this test be put into auth_test.go, below TestCompareSha256PasswordAuthData_EmptyPassword?

ramnes: do you want to move `checkSha2CacheCredentials` in `auth.go` as well?

lance6716: Oh, your reason is to use auth_switch_response_test.go to check function in auth_switch_response.go. I think it's reasonable from this view.

PR #1084, opened 2025-12-19
Handle empty password from broader set of MySQL clients

ramnes:
Some MySQL clients (e.g. libmysql) send a single null byte to indicate an empty password, while others (e.g. mariadb) send an empty packet. This matches MySQL server's own handling:

```c
if (!pkt_len || (pkt_len == 1 && *pkt == 0))
```

(Source: https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sha2_password.cc)

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -18,6 +18,16 @@ var (
 	ErrAccessDeniedNoPassword = fmt.Errorf("%w without password", ErrAccessDenied)
 )
 
+// isEmptyPassword returns true if the auth data represents an empty password.
+// Some clients send an empty packet (len == 0), while others (e.g. MySQL's libmysql)
+// send a single null byte. This matches MySQL server's own handling:
+// if (!pkt_len || (pkt_len == 1 && *pkt == 0))
+// https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sha2_password.cc
+// https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sql_authentication.cc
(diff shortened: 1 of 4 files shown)


server/auth.go
```diff
@@ -66,7 +75,7 @@ func scrambleValidation(cached, nonce, scramble []byte) bool {
 }
 
 func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential Credential) error {
-	if len(clientAuthData) == 0 {
+	if isEmptyPassword(clientAuthData) {
```
lance6716, line 79: I see your MySQL reference is only for caching_sha2_password_authenticate. Will it also apply for compareNativePasswordAuthData and compareSha256PasswordAuthData?

ramnes: SHA256 authentication does have the exact same check here: https://github.com/mysql/mysql-server/blob/056a391cdc1af9b17b5415aee243483d1bac532d/sql/auth/sql_authentication.cc#L4763

For `mysql_native_password`, I'm less sure but my understanding is that both cases were supported as well by setting a default `\0`, and then trimming with `get_length_encoded_string`: https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sql_authentication.cc#L3003-L3012

lance6716: Please also add the second link to function comment of `isEmptyPassword`. Rest lgtm

ramnes: Done!

Issue #1054, opened 2025-08-20
[Question] How can I stream the results?


fansehep:
In my code:
```go
package main

import (
	"log"
	"net"
	"time"

	"github.com/go-mysql-org/go-mysql/mysql"
	"github.com/go-mysql-org/go-mysql/server"
)

type MySQLStreamResultTestHandler struct {
	clientAddr string
	*server.Conn
}

func (h *MySQLStreamResultTestHandler) UseDB(dbName string) error {
	return nil
}

func (h *MySQLStreamResultTestHandler) HandleQuery(query string) (*mysql.Result, error) {
	log.Printf("Received query: %s from %s", query, h.clientAddr)
	if query == "show binary logs" {
		names := []string{"Log_name", "File_size"}
		values := [][]any{{"mysql-bin.001000", "123"}}
		r1, _ := mysql.BuildSimpleResultset(names, values, false)
		// r1.Streaming = mys
		r1.StreamingDone = false
		r1.Streaming = mysql.StreamingSelect

		r1result := mysql.NewResult(r1)
		h.WriteValue(r1result)

		r2, _ := mysql.BuildSimpleResultset(names, values, false)
		r2.Streaming = mysql.StreamingNone
		r2.StreamingDone = true
		r2result := mysql.NewResult(r2)
		// h.WriteValue(r2result)

		return r2result, nil
	}

	return nil, nil
}

func (h *MySQLStreamResultTestHandler) HandleFieldList(table string, fieldWildcard string) ([]*mysql.Field, error) {
	return nil, nil
}

func (h *MySQLStreamResultTestHandler) HandleStmtPrepare(query string) (params int, columns int, context interface{}, err error) {
	return 0, 0, nil, nil
}

func (h *MySQLStreamResultTestHandler) HandleStmtExecute(context interface{}, query string, args []interface{}) (*mysql.Result, error) {
	return nil, nil
}

func (h *MySQLStreamResultTestHandler) HandleStmtClose(context interface{}) error {
	return nil
}

func (h *MySQLStreamResultTestHandler) HandleOtherCommand(cmd byte, data []byte) error {
	return nil
}

func main() {

	listener, err := net.Listen("tcp", "127.0.0.1:4000")
	if err != nil {
		log.Fatalf("Failed to listen on TCP address: %v", err)
	}

	go func() {
		for {
			conn, err := listener.Accept()
			if err != nil {
				log.Fatalf("Failed to accept TCP connection: %v", err)
				return
[...]

PR #1089, opened 2026-01-18
feat: support return stream result

fansehep:
This PR introduces StreamResult, a new streaming interface for sending MySQL query results row by row. This is useful for scenarios where:
- The result set is too large to fit in memory
- Results need to be sent incrementally as they are generated

### DIFF
--- a/examples/stream_result/README.md
+++ b/examples/stream_result/README.md
@@ -18,7 +18,7 @@ fields := []*mysql.Field{
     {Name: []byte("id"), Type: mysql.MYSQL_TYPE_LONG},
     {Name: []byte("name"), Type: mysql.MYSQL_TYPE_VAR_STRING},
 }
-sr := mysql.NewStreamResult(fields, 10)  // buffer size = 10
+sr := mysql.NewStreamResult(fields, 10, false)  // buffer=10, text protocol
 
 // 2. Write rows in goroutine
 go func() {
--- a/examples/stream_result/main.go
+++ b/examples/stream_result/main.go
@@ -16,9 +16,7 @@ import (
 
 // StreamResultHandler implements the server.Handler interface
 // and demonstrates streaming query results.
-type StreamResultHandler struct {
-	conn *server.Conn
-}
+type StreamResultHandler struct{}
 
 func (h *StreamResultHandler) UseDB(dbName string) error {
 	log.Printf("USE %s", dbName)
@@ -134,8 +132,6 @@ func handleConnection(srv *server.Server, conn net.Conn) {
 		log.Printf("Failed to create MySQL connection: %v", err)
 		return
 	}
-	handler.conn = mysqlConn
-
 	log.Printf("Client connected: %s", conn.RemoteAddr())
 
 	// Handle commands in a loop


fansehep:
about: https://github.com/go-mysql-org/go-mysql/issues/1054

can someone take a look?

mysql/stream_result.go
```diff
@@ -0,0 +1,116 @@
...
+// channel returned by RowsChan().
+type StreamResult struct {
+	// Fields contains the column metadata for the result set.
+	Fields []*Field
+	// rowsChan is the internal channel for streaming rows.
+	rowsChan chan []any
+
+	// closed indicates whether the stream has been closed.
+	closed bool
+	// mu protects closed and err fields.
```
lance6716: better to put `mu` above `closed` and `err`

mysql/stream_result.go
```diff
@@ -0,0 +1,116 @@
...
+	bufSize := 100
+	if len(bufferSize) > 0 && bufferSize[0] > 0 {
+		bufSize = bufferSize[0]
+	}
+
+	fields := make([]*Field, len(columnNames))
+	for i, name := range columnNames {
+		fields[i] = &Field{
+			Name: []byte(name),
+			Type: MYSQL_TYPE_VAR_STRING,
```
lance6716: should we let caller to decide the type of fields/columns?

mysql/stream_result.go
```diff
@@ -0,0 +1,116 @@
...
+// RowsChan returns a read-only channel for consuming rows.
+// The channel is closed when Close() is called on the StreamResult.
+// Consumers should range over this channel to receive all rows.
+func (sr *StreamResult) RowsChan() <-chan []any {
+	return sr.rowsChan
+}
+
+// WriteRow sends a row to the stream. It returns true if the row was successfully
+// written, or false if the stream has been closed. This method is safe for
+// concurrent use and will block if the channel buffer is full.
```
lance6716: because it will block, I think it's more proper to use a `ctx` to allow cancelation sending to channel.

fansehep: good suggestion, done

mysql/stream_result.go
fansehep: done

mysql/stream_result.go
fansehep: done

fansehep:
cc @lance6716 

I had added an example about the stream result usage example, and fix the reviews.

examples/stream_result/main.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func handleConnection(conn net.Conn) {
+	defer conn.Close()
+
+	handler := &StreamResultHandler{}
+
+	// Create a new MySQL server connection
+	mysqlConn, err := server.NewConn(conn, "root", "", handler)
```
lance6716: it says better to create a server by `NewDefaultServer` and use its method

examples/stream_result/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/server"
+)
+
+// StreamResultHandler implements the server.Handler interface
+// and demonstrates streaming query results.
+type StreamResultHandler struct {
+	server.EmptyHandler
+	conn *server.Conn
```
lance6716: no need to add this field?

mysql/stream_result.go
```diff
@@ -0,0 +1,119 @@
...
+// Consumers should range over this channel to receive all rows.
+func (sr *StreamResult) RowsChan() <-chan []any {
+	return sr.rowsChan
+}
+
+// WriteRow sends a row to the stream. It returns true if the row was successfully
+// written, or false if the stream has been closed or context is canceled.
+// This method is safe for concurrent use and will block if the channel buffer is full.
+func (sr *StreamResult) WriteRow(ctx context.Context, row []any) (ok bool) {
+	defer func() {
```
lance6716: I think the `recover` should be put at the level where `go func() { ... }` is called. This function seems has no point to panic

mysql/stream_result.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Close closes the stream and signals to consumers that no more rows will be sent.
+// This method is idempotent and safe to call multiple times.
+// After Close() is called, WriteRow() will return false.
+func (sr *StreamResult) Close() {
+	sr.mu.Lock()
+	defer sr.mu.Unlock()
+
+	// Check if already closed by testing the done channel
```
lance6716: nit: we can reuse below `IsClosed()`

server/resp.go
```diff
@@ -163,6 +163,48 @@ func (c *Conn) writeResultset(r *mysql.Resultset) error {
...
+		return err
+	}
+
+	for row := range sr.RowsChan() {
+		data = data[0:4]
+		for _, v := range row {
+			if v == nil {
+				data = append(data, 0xfb) // NULL
+			} else {
+				tv, err := mysql.FormatTextValue(v)
```
lance6716, line 198: (I have forgotten some protocol details) Seems you are assuming it's text protocol, not binary protocol. You can refer to `BuildSimpleResultset` to also support both one, or add limitation comments on public functions of `StreamResult`

fansehep: done, both support text protocol and binary protocol.

lance6716:
```
--- FAIL: TestStreamResultClose (0.00s)
panic: send on closed channel [recovered]
	panic: send on closed channel

goroutine 105 [running]:
testing.tRunner.func1.2({0xf608c0, 0x11a59c0})
	/opt/hostedtoolcache/go/1.24.11/x64/src/testing/testing.go:1734 +0x3eb
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.24.11/x64/src/testing/testing.go:1737 +0x696
panic({0xf608c0?, 0x11a59c0?})
	/opt/hostedtoolcache/go/1.24.11/x64/src/runtime/panic.go:792 +0x132
github.com/go-mysql-org/go-mysql/mysql.(*StreamResult).WriteRow(0xc0005ec3c0, {0x11af9e0, 0x1747100}, {0xc00003c170, 0x1, 0x1})
	/home/runner/work/go-mysql/go-mysql/mysql/stream_result.go:54 +0x171
github.com/go-mysql-org/go-mysql/server.TestStreamResultClose(0xc000226a80)
	/home/runner/work/go-mysql/go-mysql/server/resp_test.go:340 +0x47e
testing.tRunner(0xc000226a80, 0x10468f0)
	/opt/hostedtoolcache/go/1.24.11/x64/src/testing/testing.go:1792 +0x226
created by testing.(*T).Run in goroutine 1
	/opt/hostedtoolcache/go/1.24.11/x64/src/testing/testing.go:1851 +0x8f3
FAIL	github.com/go-mysql-org/go-mysql/server	1.776s
```

server/resp_test.go
```diff
@@ -261,3 +262,304 @@ func TestWriteValue(t *testing.T) {
...
+	// close should be idempotent
+	require.False(t, sr.IsClosed())
+	sr.Close()
+	require.True(t, sr.IsClosed())
+
+	// calling Close again should not panic
+	sr.Close()
+	require.True(t, sr.IsClosed())
+
+	// WriteRow should return false after close
```
lance6716: Is it really needed to support concurrent-safe write vs close?

And I think even concurrent-safe write vs write is not common, database result set may contains ORDER BY, they should not be re-ordered.

fansehep: agree with you. I remove this test.

mysql/stream_result.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"context"
+	"sync"
+)
+
+// StreamResult provides a streaming interface for sending query results row by row.
+// It is designed for scenarios where the result set is too large to fit in memory
+// or when results need to be sent incrementally as they are generated.
+//
+// StreamResult is safe for concurrent use. The producer writes rows via WriteRow()
```
lance6716: After deleting the unit tests, please also update the comment (here and function comments) to explain the expected usage of `WriteRow` and `Close`. I'll review according to the description.

lance6716 (commented):
Also, pleae check unresolved comments

mysql/stream_result.go
fansehep: done

examples/stream_result/main.go
fansehep: done

examples/stream_result/main.go
fansehep: done

mysql/stream_result.go
fansehep: done, good suggestion!

examples/stream_result/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-mysql-org/go-mysql/server"
+)
+
+// StreamResultHandler implements the server.Handler interface
+// and demonstrates streaming query results.
+type StreamResultHandler struct {
+	conn *server.Conn
+}
```
lance6716:
```suggestion
type StreamResultHandler struct {}
```

mysql/stream_result.go
lance6716: this one, or explain that you don't want to reuse `IsClosed()`

mysql/stream_result.go
```diff
@@ -0,0 +1,144 @@
...
+// The consumer reads rows from the channel returned by RowsChan().
+//
+// Usage pattern:
+//
+//	sr := mysql.NewStreamResult(fields, bufSize, binary)
+//	go func() {
+//	    defer sr.Close()  // Always close when done
+//	    for row := range rows {
+//	        if !sr.WriteRow(ctx, row) {
+//	            return  // Context canceled or stream closed by consumer
```
lance6716: If we allow consumer to close data channel, panic is still possible now:

1. consumer close the `rowsChan` channel.
2. producer goroutine enter `WriteRow`, try to send to `rowsChan`

Maybe you can learn how https://pkg.go.dev/io#Pipe works, or directly use it

fansehep: I believe allowing producers to close the channel is a poor design for users. Our design is for producers to close the channel after sending data.

fansehep: The `io.Pipe` interface in Go sends `[]byte`, which requires users to manually serialize data for transmission. I believe channels are sufficient for this case.

lance6716: > I believe allowing consumers to close the channel is a poor design for users. Our design is for consumers to close the channel after sending data.

Do you mean producer (sender, writer) or consumer (receiver, reader)? Because your comment says "stream closed by consumer" so consumer can close channel. And you also just said "allowing consumers to close the channel is a poor design"

lance6716: > sends []byte, which requires users to manually serialize data for transmission

I mean learn `io.Pipe` on how to do a reader-side close. It also uses channels internally.

fansehep: > > I believe allowing consumers to close the channel is a poor design for users. Our design is for consumers to close the channel after sending data.
> 
> Do you mean producer (sender, writer) or consumer (receiver, reader)? Because your comment says "stream closed by consumer" so consumer can close channel. And you also just said "allowing consumers to close the channel is a poor design"

This annotation contains a typographical error. Consumer can't close this channal.

examples/stream_result/main.go
lance6716: This field is not needed?

examples/stream_result/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Usage
+
+```go
+// 1. Create StreamResult
+fields := []*mysql.Field{
+    {Name: []byte("id"), Type: mysql.MYSQL_TYPE_LONG},
+    {Name: []byte("name"), Type: mysql.MYSQL_TYPE_VAR_STRING},
+}
+sr := mysql.NewStreamResult(fields, 10)  // buffer size = 10
```
lance6716: this file should also be updated. Like there's the 3rd parameter.

fansehep: done，fix this.

PR #1091, opened 2026-01-27
Adds OnQueryEvent() and OnUserVar() handlers

dbnski:
The PR adds handlers that allow tracking the original clients' session vars (e.g. charsets for context) and at least some user vars that may be referenced in occasional SBR queries.


replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+			size = int(e.StatusVars[pos])
+			pos += 1
+
+			vars.InvokerHostname = make([]byte, size)
+			copy(vars.InvokerHostname, e.StatusVars[pos:])
+			pos += size
+		case 0x0c:
+			pos += 1
+
+			c := uint8(e.StatusVars[pos])
```
lance6716, line 575: hi @dbnski please fix this linter error

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+	UpdatedDBNames               [][]byte
+	Microseconds                 uint32
+	ExplicitDefaultsForTimestamp uint8
+	XID                          uint64
+	DefaultCollationForUtf8MB4   uint16
+	SQLRequirePrimaryKey         uint8
+	DefaultTableEncryption       uint8
+}
+
+func (e *QueryEvent) GetStatusVars() *QueryStatusVars {
```
lance6716, line 480: Can you port this test here? https://github.com/pingcap/tiflow/blob/master/dm/pkg/binlog/event/util_test.go#L141 (I have added these code years ago in our projects. And I think this project is a better place for processing all things related to binlog events)

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
 	fmt.Fprintln(w)
 }
 
+type QueryStatusVars struct {
+	Options                      uint32
```
dveeden, line 457: I think it would be good to put the names from the docs in a comment like this:

```suggestion
	Options                      uint32 // Q_FLAGS2_CODE
```

When working on #1101 I first thought that some options might appear multiple times, but that isn't the case. So a type like this is ok. In that PR I used this:
```go
type QueryEventStatusVar struct {
	VarType  int
	VarValue any
}
```
There is one possible advantage of that, which is that it is more explicit about what options were used for the event (but possibly with a default value) and which ones were not in the binlog.

My PR also decodes the values of FLAGS2:
```go
type QFlags uint32

func (f QFlags) String() string {
...
}
```

While it is 4 bytes, FLAGS2 only has 4 bits that are actually used. This is because it uses the same bitmask as a bitmask that is used in MySQL Server for other purposes. From these 4 bits they also stopped to use the bit that set the `autocommit` (see comment in the server code). So in practice only 3 bits out of the 4 bytes are actually used. Yes this wastes 29 out of 32 bits...

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+	SQLMode                      uint64
```
dveeden, line 458: In #1101 I'm decoding the SQLMode like this:

```go
type SQLMode uint64

func (m SQLMode) String() string {
	var modes []string

	if uint64(m)&MODE_REAL_AS_FLOAT != 0 {
		modes = append(modes, "REAL_AS_FLOAT")
	}
```

Note that `mysqlbinlog` doesn't decode this and MySQL allows you to set a SQL mode with an integer value.

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+	AutoIncrementIncrement       uint16
+	AutoIncrementOffset          uint16
+	CharacterSetClient           uint16
+	CollationConnection          uint16
+	CollationServer              uint16
+	TimezoneCode                 []byte
+	CatalogNZCode                []byte
```
dveeden, line 465: I think we can use this for both `Q_CATALOG_NZ_CODE` and `Q_CATALOG_CODE` (the latter is almost never used)
```suggestion
	Catalog                []byte
```

The in MySQL is not very consistent. Newer additions don't end in `_CODE`. I think we can remove the `Code` suffix in this type.

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+	InvokerUsername              []byte
+	InvokerHostname              []byte
+	UpdatedDBNames               [][]byte
+	Microseconds                 uint32
+	ExplicitDefaultsForTimestamp uint8
+	XID                          uint64
+	DefaultCollationForUtf8MB4   uint16
+	SQLRequirePrimaryKey         uint8
+	DefaultTableEncryption       uint8
+}
```
dveeden, line 478: MariaDB has two MariaDB only types: `Q_HRNOW` and `Q_XID`. However MySQL has almost identical options: `Q_MICROSECONDS` and `Q_DDL_LOGGED_WITH_XID`. Are we storing those in the same members of the type?

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
 	fmt.Fprintln(w)
 }
 
+type QueryStatusVars struct {
```
dveeden, line 456:
```suggestion
// QueryStatusVars is storing the status variables of a query event
type QueryStatusVars struct {
```

replication/event.go
dveeden, line 480:
```suggestion
// GetStatusVars is decoding the status variables of the event that are stored in e.StatusVars
func (e *QueryEvent) GetStatusVars() *QueryStatusVars {
```

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+
+			vars.Options = binary.LittleEndian.Uint32(e.StatusVars[pos:])
+			pos += 4
+		case 0x01:
+			pos += 1
+
+			vars.SQLMode = binary.LittleEndian.Uint64(e.StatusVars[pos:])
+			pos += 8
+		case 0x02:
+			// Not used
```
dveeden, line 501:
```suggestion
			// Q_CATALOG_CODE, Not used. This was replaced by Q_CATALOG_NZ_CODE
```

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+	if len(e.StatusVars) == 0 {
+		return nil
+	}
+
+	vars := &QueryStatusVars{}
+	pos := 0
+
+	for pos < len(e.StatusVars) {
+		switch e.StatusVars[pos] {
+		case 0x00:
```
dveeden, line 490:
```suggestion
		case Q_FLAGS2_CODE:
```

And similar for all other entries

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+
+			vars.CharsetDatabaseCode = binary.LittleEndian.Uint16(e.StatusVars[pos:])
+			pos += 2
+		case 0x09:
+			pos += 1
+
+			vars.TableMapForUpdateCode = binary.LittleEndian.Uint64(e.StatusVars[pos:])
+			pos += 8
+		case 0x0a:
+			// Not used
```
dveeden, line 555:
```suggestion
			// Not used: Q_MASTER_DATA_WRITTEN_CODE
```

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+
+			low := binary.LittleEndian.Uint16(e.StatusVars[pos:])
+			pos += 2
+
+			high := uint8(e.StatusVars[pos])
+			pos += 1
+
+			vars.Microseconds = uint32(low) + (uint32(high) << 16)
+		case 0x0e:
+			// Not used
```
dveeden, line 602:
```suggestion
			// Not used: Q_COMMIT_TS
```

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+		case 0x12:
+			pos += 1
+
+			vars.DefaultCollationForUtf8MB4 = binary.LittleEndian.Uint16(e.StatusVars[pos:])
+			pos += 2
+		case 0x13:
+			pos += 1
+
+			vars.SQLRequirePrimaryKey = e.StatusVars[pos]
+			pos += 1
```
dveeden:
```suggestion
			pos += 2
```

I think this is wrong, the docs say this is two bytes. I think the first byte is 0 or 1, and the second byte is storing the type. I think for `Q_SQL_REQUIRE_PRIMARY_KEY` and `Q_DEFAULT_TABLE_ENCRYPTION` the type is static and can be ignored, but it should still be skipped over.

This can be a bool instead of a uint8.

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+		case 0x14:
+			pos += 1
+
+			vars.DefaultTableEncryption = e.StatusVars[pos]
+			pos += 1
```
dveeden:
```suggestion
			pos += 2
```

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+		case 0x80:
```
dveeden, line 630:
```suggestion
		case 0x80: // Q_HRNOW, MariaDB Only
```

Do we need to register that this is a MariaDB event so we can correctly reconstruct the data from the `QueryStatusVars` type and get back the same as the original value?

replication/event.go
```diff
@@ -418,6 +418,203 @@ func (e *QueryEvent) Dump(w io.Writer) {
...
+			pos += 1
+
+			vars.Microseconds = uint32(low) + (uint32(high) << 16)
+		case 0x81:
+			pos += 1
+
+			vars.XID = binary.LittleEndian.Uint64(e.StatusVars[pos:])
+			pos += 8
+		default:
+			return nil
```
dveeden, line 646:
```suggestion
            slog.Warn("failed to decode query event variable", "type", e.StatusVars[pos])
			return nil
```

Or any other way to make this not ignore things silently...

dveeden (commented):
I think we need to combine/merge this with #1101 as both do more or less the same.

dbnski:
> What prompted you to work on this? Any interesting usecase?

I found some of the information from these events very helpful in tracking down and automated healing of double encoded writes.

replication/event.go
dbnski: You're right about the length, and it makes sense to use bool for this flag.

PR #1096, opened 2026-01-29
Decouple RSA auth key from TLS config

ramnes:
Decouple the RSA private key used for `caching_sha2_password` authentication from the TLS configuration by accepting it as a separate parameter.

Without this PR, the RSA key is always extracted from `tlsConfig.Certificates[0]`. This coupling causes issues when using GetCertificate for dynamic certificate loading (e.g., certificate rotation, or serving different certs per connection):

- Go's TLS only calls `GetCertificate` when `Certificates` is empty OR the client sends SNI
- MySQL clients don't send SNI during the TLS handshake
- Therefore, if `Certificates` was populated (to provide the RSA key), the server would serve that certificate instead of calling `GetCertificate`

This is particularly problematic when the TLS certificate uses ECDSA (common with modern CAs like Let's Encrypt and ZeroSSL) but RSA is still needed for the `caching_sha2_password` authentication flow over non-TLS connections.

To fix this, we pass the RSA private key directly to `NewServerWithAuth`. It allows users to:
- Use `GetCertificate` for TLS without populating `Certificates`
- Use ECDSA certificates for TLS while still supporting `caching_sha2_password` non-TLS connections
- Rotate TLS certificates dynamically without affecting authentication

Note: I can add a new constructor if you don't want to break the API on existing ones.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -118,7 +118,10 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	} else {
 		// client should send encrypted password
 		// decrypt
-		dbytes, err := rsa.DecryptOAEP(sha1.New(), rand.Reader, (c.serverConf.tlsConfig.Certificates[0].PrivateKey).(*rsa.PrivateKey), clientAuthData, nil)
+		if c.serverConf.rsaPrivateKey == nil {
+			return errors.New("RSA key not configured; non-TLS connections are not supported for this authentication method")
+		}
+		dbytes, err := rsa.DecryptOAEP(sha1.New(), rand.Reader, c.serverConf.rsaPrivateKey, clientAuthData, nil)
 		if err != nil {
 			return err
 		}
--- a/server/auth_switch_response.go
+++ b/server/auth_switch_response.go
@@ -51,7 +51,10 @@ func (c *Conn) handleCachingSha2PasswordFullAuth(authData []byte) error {
 		}
 		// the encrypted password
 		// decrypt
-		dbytes, err := rsa.DecryptOAEP(sha1.New(), rand.Reader, (c.serverConf.tlsConfig.Certificates[0].PrivateKey).(*rsa.PrivateKey), authData, nil)
+		if c.serverConf.rsaPrivateKey == nil {
+			return errors.New("RSA key not configured; non-TLS connections are not supported for this authentication method")
+		}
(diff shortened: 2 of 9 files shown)


lance6716:
@dveeden Can you take a look? I believe I have less knowledge than you about TLS things.

lance6716:
ping @dveeden

server/auth.go
```diff
@@ -118,7 +118,10 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	} else {
 		// client should send encrypted password
 		// decrypt
-		dbytes, err := rsa.DecryptOAEP(sha1.New(), rand.Reader, (c.serverConf.tlsConfig.Certificates[0].PrivateKey).(*rsa.PrivateKey), clientAuthData, nil)
+		if c.serverConf.rsaPrivateKey == nil {
+			return errors.New("RSA key not configured; non-TLS connections are not supported for this authentication method")
+		}
+		dbytes, err := rsa.DecryptOAEP(sha1.New(), rand.Reader, c.serverConf.rsaPrivateKey, clientAuthData, nil)
```
dveeden, line 124: Does this have to be `SHA1`? or can we use something more modern?

ramnes: As far as I can tell, it can't be changed:
* OpenSSL defaults to SHA-1 for RSA OAEP when not explicitly set: https://github.com/openssl/openssl/blob/openssl-3.0/crypto/rsa/rsa_oaep.c#L68-L71
* MySQL seems to use OpenSSL without overriding this default. As you can see here, it doesn't use `EVP_PKEY_CTX_set_rsa_oaep_md_name()`:  https://github.com/mysql/mysql-server/blob/8.0/sql/auth/auth_common.cc#L213-L236
* `crypto/rsa` requires explicit hash specification, so we use `sha1.New()` to match.

dveeden: Thanks. Clear.

Maybe consider opening a bug about this for MySQL and/or send them a PR. They have been working hard to remove SHA1 and MD5, both as database functions and for things that depend on it like `mysql_native_password`. Looks like they missed this as the OpenSSL API hides this (while `crypto/rsa` does not).

ramnes: SHA-1 in RSA-OAEP is just used as a PRF/MGF, not for collision resistance or signatures, so the risk profile is very different. My understanding is that the weaknesses of SHA-1 do not apply here and it's very likely why OpenSSL didn't change the default. Interesting discussion on that matter: https://security.stackexchange.com/questions/112029/should-sha-1-be-used-with-rsa-oaep

PR #1097, opened 2026-02-04
replication: Print header flags

dveeden:
```
$ for f in ~/mysql-bin.*; do echo $f; go run cmd/go-binlogparser/main.go -name $f | grep Header\ Flags; done
/home/dvaneeden/mysql-bin.000001
Header Flags: 
Header Flags: IGNORABLE
Header Flags: 
/home/dvaneeden/mysql-bin.000002
Header Flags: 
Header Flags: IGNORABLE
Header Flags: 
/home/dvaneeden/mysql-bin.000003
Header Flags: 
Header Flags: IGNORABLE
Header Flags: 
Header Flags: SUPPRESS_USE
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: SUPPRESS_USE
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: SUPPRESS_USE
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: SUPPRESS_USE
Header Flags: 
Header Flags: 
Header Flags: 
Header Flags: 
/home/dvaneeden/mysql-bin.000004
Header Flags: IN_USE
Header Flags: IGNORABLE
```

This shows the IN_USE flags and other flags.

### DIFF
--- a/replication/event.go
+++ b/replication/event.go
@@ -103,43 +103,40 @@ func (h *EventHeader) Decode(data []byte) error {
 
 // headerFlagsString is returning a pipe separated string with flag names
 func headerFlagsString(flags uint16) string {
-	var fstr string
+	var flagstr []string
 
 	if (flags & LOG_EVENT_BINLOG_IN_USE_F) != 0 {
-		fstr += "|IN_USE"
+		flagstr = append(flagstr, "IN_USE")
 	}
 	if (flags & LOG_EVENT_FORCED_ROTATE_F) != 0 {
-		fstr += "|ROTATE"
+		flagstr = append(flagstr, "ROTATE")
 	}
 	if (flags & LOG_EVENT_THREAD_SPECIFIC_F) != 0 {
-		fstr += "|THREAD_SPECIFIC"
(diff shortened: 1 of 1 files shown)


replication/event.go
```diff
@@ -101,11 +101,53 @@ func (h *EventHeader) Decode(data []byte) error {
...
+		fstr += "|IN_USE"
+	}
+	if (flags & LOG_EVENT_FORCED_ROTATE_F) != 0 {
+		fstr += "|ROTATE"
+	}
+	if (flags & LOG_EVENT_THREAD_SPECIFIC_F) != 0 {
+		fstr += "|THREAD_SPECIFIC"
+	}
+	if (flags & LOG_EVENT_SUPPRESS_USE_F) != 0 {
+		fstr += "|SUPPRESS_USE"
```
lance6716: we can use a `[]string` and strings.Join them at last. No intermediate strings are needed.

dveeden: Yes indeed!

PR #1104, opened 2026-02-21
driver: make DSN more flexible

dveeden:
Changes
- Dependencies: Update all dependencies
- Dependencies: Remove `github.com/go-sql-driver/mysql` dependency
- DSN: Allow `mysql://` as URL scheme. This allows us to accept the same URLs as MySQL Shell 
- DSN: Allow `user@tcp(127.0.0.1:3306)/db` style URLs to be compatible with go-sql-driver/mysql URLs (restricted to `tcp(...)` )
- DSN: Make `tls` an alias for `ssl`.
- DSN: Allow `false` and `skip-verify` as options for `ssl`/`tls`
- CI: Update MySQL versions
- CI: Update Go versions
- CI: Update actions


dveeden:
I also changed the two required checks to those for Go 1.24 and Go 1.25 (from Go 1.23 and Go 1.24).

Go 1.26 and Go 1.25 and the versions that are under active maintenance (see https://go.dev/doc/devel/release )

driver/driver.go
```diff
@@ -92,11 +94,20 @@ func parseDSN(dsn string) (connInfo, error) {
...
+	if !strings.HasPrefix(dsn, "mysql://") {
+		dsn = "mysql://" + dsn
+	}
+
 	parsedDSN, parseErr := url.Parse(dsn)
 	if parseErr != nil {
-		return ci, errors.Errorf("invalid dsn, must be user:password@addr[/db[?param=X]]")
+		// If parsing fails, try to rewrite `tcp(127.0.0.1:3306)` to `127.0.0.1:3306` and try again.
+		// This is for compatibility with go-sql-driver/mysql.
+		dsn = compatDSNre.ReplaceAllString(dsn, "$2")
```
lance6716, line 106: I'm a bit worried about this non-structural replacement will cause some problems, like this pattern also appears in password or other parts. I'll ask AI to improve it and give LGTM soon

dveeden: Maybe changing the pattern to look for `@tcp(` instead of `tcp(` would be good?

lance6716: Gemini web chat:

---

### 🚨 Potential Risks in the Current Regex Implementation

The current regular expression `(tcp\((.*)\))` and its usage with `ReplaceAllString(dsn, "$2")` introduce two critical risks to the DSN parsing logic:

**1. Greedy Matching (`.*`) corrupts the DSN structure**
Because `.*` is greedy, it will capture everything up to the *last* `)` in the entire string.

* **Scenario:** If the URL contains query parameters with parentheses (e.g., `mysql://user:pass@tcp(127.0.0.1:3306)/db?param=func(1)`).
* **Consequence:** The regex will match everything from `tcp(` all the way to the closing parenthesis of `func(1)`. The replacement will truncate and corrupt the DSN (e.g., turning it into `127.0.0.1:3306)/db?param=func(1`), causing `url.Parse` to fail or produce invalid data.

**2. Global Replacement risks altering valid Credentials**
The replacement is executed globally across the entire `dsn` string.

* **Scenario:** If a randomly generated strong password happens to contain the exact pattern `tcp(...)` (e.g., `mysql://root:my_tcp(123)pass@127.0.0.1:3306/db`).
* **Consequence:** The password will be silently modified during parsing (replacing `tcp(123)` with just `123`). This results in baffling authentication failures that are incredibly difficult to debug.

---

### 🛠️ Recommended Fix (Regex Optimization)

To fix this, we need to eliminate the greedy match and strictly anchor the replacement to the Host portion of the URL, ignoring the credentials part.

**Update the regex to:**

```go
// Matches "://" followed by optional credentials, then targets the "tcp(XXX)" host.
// Uses [^)]+ to ensure it only captures inside the parentheses, preventing greedy consumption.
var compatDSNre = regexp.MustCompile(`(://(?:[^@]*@)?)tcp\(([^)]+)\)`)

```

**Update the replacement logic to:**

```go
// Retains the scheme and credentials ($1), and safely extracts just the host ($2)
dsn = compatDSNre.ReplaceAllString(dsn, "${1}${2}")

```

**Why this works:**
By using `(://(?:[^@]*@)?)`, the regex safely captures the scheme and everything up to the *last* `@` symbol (the credentials boundary) as Group 1 (`$1`). The `[^)]+` ensures we only capture the actual host IP/port inside `tcp(...)` as Group 2 (`$2`), completely protecting the username, password, and URL parameters from being accidentally modified.

driver/driver_test.go
```diff
@@ -88,12 +88,20 @@ func TestParseDSN(t *testing.T) {
...
+		"user:password@4.domain.com/db?tls=custom":      {standardDSN: true, addr: "4.domain.com", user: "user", password: "password", db: "db", params: url.Values{"tls": []string{"custom"}}},
 		"user:password@5.domain.com/db?unused=param":    {standardDSN: true, addr: "5.domain.com", user: "user", password: "password", db: "db", params: url.Values{"unused": []string{"param"}}},
 		"user:password@5.domain.com/db?timeout=1s":      {standardDSN: true, addr: "5.domain.com", user: "user", password: "password", db: "db", params: url.Values{"timeout": []string{"1s"}}},
 		"user:password@5.domain.com/db?readTimeout=1m":  {standardDSN: true, addr: "5.domain.com", user: "user", password: "password", db: "db", params: url.Values{"readTimeout": []string{"1m"}}},
 		"user:password@5.domain.com/db?writeTimeout=1m": {standardDSN: true, addr: "5.domain.com", user: "user", password: "password", db: "db", params: url.Values{"writeTimeout": []string{"1m"}}},
 		"user:password@5.domain.com/db?compress=zlib":   {standardDSN: true, addr: "5.domain.com", user: "user", password: "password", db: "db", params: url.Values{"compress": []string{"zlib"}}},
```
lance6716, line 99: codex also find a bug for this test case, it will panic

```
		"mysql://127.0.0.1:3306":                        {standardDSN: true, addr: "127.0.0.1:3309", user: "", password: "", db: "", params: url.Values{}},
```

We can fix it in this PR or a separate one.

driver/driver.go
```diff
@@ -92,11 +94,20 @@ func parseDSN(dsn string) (connInfo, error) {
 		ci.standardDSN = strings.Contains(dsn, "/")
 	}
 
-	// Add a prefix so we can parse with url.Parse
-	dsn = "mysql://" + dsn
+	// Add a URL scheme if it isn't there so we can parse with url.Parse
+	if !strings.HasPrefix(dsn, "mysql://") {
```
lance6716, line 98: nit: maybe also ToLower to support `MySQL://`

Issue #1093, opened 2026-01-27
Canal silently skips events when table does not exist - possible data loss


dbnski:
The following code means that if a table does not exist on the server when the event is read, the event will be silently ignored potentially leading to data loss:
```
	t, err := c.GetTable(schemaName, tableName)
	if err != nil {
		e := errors.Cause(err)
		// ignore errors below
		if e == ErrExcludedTable || e == schema.ErrTableNotExist || e == schema.ErrMissingTableMeta {
			err = nil
		}

		return err
	}
```
It can only happen in rare circumstances (e.g. table is renamed while Canal is restarting), but given the consequences that may arise from this the error should not be muted by default. Similarly to `DiscardNoMetaRowEvent`, which only handles a very specific case, there should be `DiscardTableNotExistEvent` with configurable options to silently discard or pass every error to the reader (e.g. through `OnRowError` handler that receives the event, the table name and the error) .

dveeden:
Linking to the code:

https://github.com/go-mysql-org/go-mysql/blob/a71deaeb3c13cf957f3d4dc68b83820ca1b6deb3/canal/sync.go#L271-L280

Yes, this needs fixing

PR #1105, opened 2026-02-22
Adds OnTableNotFound() error handler to Canal

dbnski:
Addresses #1093 by introducing a new `OnTableNotFound()` handler that receives the complete event for whatever it wants to do with it. The handler can subsequently return an error (new default behavior), which will fail the stream, or `nil` to ignore and skip the problem (old default behavior). It also fixes a minor issue where the `e` variable holding the event would previously be shadowed by an error var taking the same name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,37 @@ func main() {
 }
 ```
 
+### Structured Connector
+
+If you prefer a structured configuration over a DSN string, you can use
+`database/sql.OpenDB` with `driver.Connector`:
+
+```go
+package main
+
+import (
+	"database/sql"
+	"net/url"
+
+	"github.com/go-mysql-org/go-mysql/driver"
+)
+
+func main() {
+	connector := driver.Connector{
+		Addr: "127.0.0.1:3306",
+		User: "root",
+		DB:   "test",
+		Params: url.Values{
+			// same option keys as the standard DSN form
+			"timeout": []string{"10s"},
+		},
+	}
+
+	db := sql.OpenDB(connector)
+	db.Close()
+}
+```
+
 ### Driver Options
 
 Configuration options can be provided by the standard DSN (Data Source Name).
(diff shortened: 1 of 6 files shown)


canal/sync.go
```diff
@@ -270,12 +270,15 @@ func (c *Canal) handleRowsEvent(e *replication.BinlogEvent) error {
...
+		cause := errors.Cause(err)
 		// ignore errors below
-		if e == ErrExcludedTable || e == schema.ErrTableNotExist || e == schema.ErrMissingTableMeta {
-			err = nil
+		if cause == ErrExcludedTable || cause == schema.ErrMissingTableMeta {
+			return nil
+		}
+		// Allow handler to decide what to do when table is missing.
+		if cause == schema.ErrTableNotExist {
+			return c.eventHandler.OnTableNotFound(e.Header, ev)
```
lance6716, line 280: Do you think `schemaName, tableName` is also helpful so they should be added to parameters?

canal/handler.go
```diff
@@ -51,6 +55,10 @@ func (h *DummyEventHandler) OnRowsQueryEvent(*replication.RowsQueryEvent) error
 	return nil
 }
 
+func (h *DummyEventHandler) OnTableNotFound(header *replication.EventHeader, e *replication.RowsEvent) error {
```
dveeden, line 59:
```suggestion
// OnTableNotFound is called for row events for reference tables that are not found
func (h *DummyEventHandler) OnTableNotFound(header *replication.EventHeader, e *replication.RowsEvent) error {
```

canal/sync.go
dbnski: The `replication.RowsEvent` object the handler is receiving has those - that's where `handleRowsEvent()` has them from too - so I don't think there's need to pass them as separate arguments.

PR #1106, opened 2026-02-23
driver: fix no db in DSN will panic

lance6716:
The added unit test reveals a panic. This PR fixed it.


lance6716:
@dveeden PTAL

PR #1107, opened 2026-02-23
driver: improve RE matching case

lance6716:
as the comment in https://github.com/go-mysql-org/go-mysql/pull/1104#discussion_r2836191338

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -82,7 +82,7 @@ type Connector struct {
 //
 //	mysql://user:pass@tcp(127.0.0.1:3306)/db?param=value
 //
-// into a standard URL form that net/url can parse:
+// into a standard RFC 3986 URL form that net/url can parse:
 //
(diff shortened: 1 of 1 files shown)


driver/driver.go
```diff
@@ -69,7 +69,17 @@ type connInfo struct {
 	params      url.Values
 }
 
-var compatDSNre = regexp.MustCompile(`(tcp\((.*)\))`)
+// compatDSNre rewrites go-sql-driver/mysql style addresses like:
+//
+//	mysql://user:pass@tcp(127.0.0.1:3306)/db?param=value
+//
+// into a standard URL form that net/url can parse:
```
dveeden:
```suggestion
// into a standard RFC 3986 URL form that net/url can parse:
```

Note that net/url mentions RFC 2396, but RFC 3986 obsoletes that.

PR #1116, opened 2026-03-07
mysql: Parsing UUIDSet with tags

### LINKED ISSUE #845: MySQL 8.3 will introduce new GTID format

dveeden:
Closes #845

### DIFF
--- a/mysql/mysql_gtid.go
+++ b/mysql/mysql_gtid.go
@@ -4,10 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"io"
-	"math"
-	"sort"
-	"strconv"
+	"slices"
 	"strings"
 
 	"github.com/go-mysql-org/go-mysql/utils"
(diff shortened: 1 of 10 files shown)


lance6716 (commented):
codex review says

1. this PR can'r parse `UUID:tag:1-3:5-6` where tags and multiple following interval
2. You can add some test logic about parsing and String, check they are not changed. Seems input `UUID:1-3:foobar:1` will be changed
3. Seems in https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html there are uppercase tags like `Domain_1`, which breaks their definitions on `[a-z_][a-z0-9_]{0,31}`

dveeden:
> codex review says
> 
>     1. this PR can'r parse `UUID:tag:1-3:5-6` where tags and multiple following interval
I'll check

>     2. You can add some test logic about parsing and String, check they are not changed. Seems input `UUID:1-3:foobar:1` will be changed
I'll check. 
 
>     3. Seems in https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html there are uppercase tags like `Domain_1`, which breaks their definitions on `[a-z_][a-z0-9_]{0,31}`

Yet another way in which those docs are wrong. I'll add this to https://bugs.mysql.com/bug.php?id=116789

Note that `Domain_1` is accepted as tag for `gtid_next`, but then is changed to lower case. The tag in the binlog is always lower case.

```
mysql-8.4.7 [test]> SELECT @@gtid_executed;
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| @@gtid_executed                                                                                                                                                                |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 40eb4bee-1972-11f1-acc9-324f96ede8f2:1-11:abcdefgh:1:abcdefghijklmnopqrstuvwxyz:1:abcdefghijklmnopqrstuvwxyz_____:1:abcdefghijklmnopqrstuvwxyz_____x:1:foobar:1-2:foobaz:1:x:1 |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.001 sec)

mysql-8.4.7 [test]> SET GTID_NEXT='AUTOMATIC:Domain_1';
Query OK, 0 rows affected (0.001 sec)

mysql-8.4.7 [test]> update t1 set c2='foobar' where id=7;
Query OK, 1 row affected (0.004 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql-8.4.7 [test]> SELECT @@gtid_executed;
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| @@gtid_executed                                                                                                                                                                           |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 40eb4bee-1972-11f1-acc9-324f96ede8f2:1-11:abcdefgh:1:abcdefghijklmnopqrstuvwxyz:1:abcdefghijklmnopqrstuvwxyz_____:1:abcdefghijklmnopqrstuvwxyz_____x:1:domain_1:1:foobar:1-2:foobaz:1:x:1 |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.001 sec)
```

dveeden:
Some more info:

https://github.com/mysql/mysql-server/blob/447eb26e094b444a88c532028647e48228c3c04f/libs/mysql/gtid/tag.h#L42-L48

So `SET GTID_NEXT='AUTOMATIC: Foo ';` results in a `foo` tag as leading and trailing whitespace is removed and the conversion to lowercase is done.

dveeden:
This removes these:

```
func (s IntervalSlice) Compare(o IntervalSlice) int {
func (s IntervalSlice) Less(i, j int) bool {
func (s IntervalSlice) Swap(i, j int) {
```

@lance6716 are they used by anything/anyone?

lance6716:
> This removes these:
> 
> ```
> func (s IntervalSlice) Compare(o IntervalSlice) int {
> func (s IntervalSlice) Less(i, j int) bool {
> func (s IntervalSlice) Swap(i, j int) {
> ```
> 
> @lance6716 are they used by anything/anyone?

I think we can delete them. They are internal concepts, I assume other developers only use GTID / GTID set rather than it

dveeden:
@gemini-code-assist review

dveeden:
API changes:

```
$ git show -U0 | egrep -E '^.func (\(.*\)) [A-Z]' | sort 
+func (i Interval) String() (s string) {
-func (i Interval) String() string {
+func (i Interval) StringWithoutTag() (s string) {
+func (s *MysqlGTIDSet) AddGTIDWithTag(uuid uuid.UUID, tag string, gno int64) {
+func (s *UUIDSet) Decode(data []byte, format GtidFormat) error {
+func (s *UUIDSet) Encode(format GtidFormat) []byte {
```

lance6716:
I posted inline review comments on the relevant code locations (focused on correctness + suggested regression tests).

---

- LLM model: gpt-5.2
- Thinking level: xhigh
- Client: Codex Desktop 26.309.31024 (build 962); codex-cli 0.114.0

mysql/mysql_uuidset.go
```diff
@@ -0,0 +1,287 @@
...
+	var err error
+	s := new(UUIDSet)
+	if s.SID, err = uuid.Parse(sep[0]); err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	// Handle interval
+	tag := ""
+	for i := 1; i < len(sep); i++ {
+		if tagRegexp.MatchString(sep[i]) {
```
lance6716, line 46: `ParseUUIDSet` currently treats any `tagRegexp`-matching token as a tag and keeps scanning. That means inputs like `UUID:foo:bar:1` are accepted and silently drop `foo` (the tag gets overwritten by `bar`).

If the intended grammar is `UUID[:tag]:interval[[:tag]:interval]`, consecutive tags should be rejected.

Suggested regression test (fails on this PR):

```go
func TestParseUUIDSet_InvalidConsecutiveTags(t *testing.T) {
    _, err := ParseUUIDSet("40eb4bee-1972-11f1-acc9-324f96ede8f2:foo:bar:1")
    require.Error(t, err)
}
```

Possible implementation: track an `expectInterval` bool; once you parse a tag, the next token must be an interval (not another tag).

mysql/mysql_gtid.go
```diff
@@ -557,13 +202,34 @@ func (s *MysqlGTIDSet) String() string {
...
+	sidcount := uint64(0)
 	for i := range s.Sets {
-		s.Sets[i].encode(&buf)
+		sidcount++
+		lasttag := s.Sets[i].Intervals[0].Tag
+		for j := range s.Sets[i].Intervals {
+			if s.Sets[i].Intervals[j].Tag != "" {
+				format = GtidFormatTagged
+			}
+			if s.Sets[i].Intervals[j].Tag != lasttag {
```
lance6716: `MysqlGTIDSet.Encode()` computes `sidcount` (number of UUID+tag groups / TSIDs) but `lasttag` is never updated after a tag change. If a tagged group has multiple intervals, `sidcount` gets overcounted and the encoded header claims more TSIDs than are actually serialized.

Concrete example: `de278ad0-2106-11e4-9f8e-6edd0ca20947:1-2:mytag:1-10:12-14:16:18-20` should have 2 groups ("" and "mytag"), but the current code produces a header sidnr of 5.

Suggested regression test (fails on this PR):

```go
func TestMysqlGTIDSet_Encode_TaggedSidCountMatchesUUIDTagGroups(t *testing.T) {
    gs, err := ParseMysqlGTIDSet("de278ad0-2106-11e4-9f8e-6edd0ca20947:1-2:mytag:1-10:12-14:16:18-20")
    require.NoError(t, err)

    b := gs.Encode()
    format, sidnr := DecodeSid(b)
    require.Equal(t, GtidFormatTagged, format)
    require.Equal(t, uint64(2), sidnr)
}
```

Fix idea: when you detect `Intervals[j].Tag != lasttag`, increment `sidcount` *and* update `lasttag = Intervals[j].Tag`.

mysql/mysql_gtid.go
```diff
@@ -557,13 +202,34 @@ func (s *MysqlGTIDSet) String() string {
...
+	// If any of the intervals have a tag, we have to use the tagged
+	// format for the full event.
+	format := GtidFormatClassic
 
+	// Count number of UUID+Tag combinations
+	sidcount := uint64(0)
 	for i := range s.Sets {
-		s.Sets[i].encode(&buf)
+		sidcount++
+		lasttag := s.Sets[i].Intervals[0].Tag
```
lance6716: Minor but potentially crashy: `lasttag := s.Sets[i].Intervals[0].Tag` will panic if a `UUIDSet` ever has `len(Intervals)==0`.

`UUIDSet.Decode` currently accepts an interval-count of 0 (see the "Possibly invalid" test cases), so it seems possible to end up with an empty `UUIDSet` in-memory and then `Encode()` would panic.

Suggested regression test (currently panics on this PR):

```go
func TestMysqlGTIDSet_Encode_DoesNotPanicOnEmptyUUIDSet(t *testing.T) {
    sid, err := uuid.Parse("071a4ecc-1bf9-11f1-a838-e6dd1807d029")
    require.NoError(t, err)

    gs := &MysqlGTIDSet{Sets: map[string]*UUIDSet{
        sid.String(): {SID: sid, Intervals: IntervalSlice{}},
    }}

    require.NotPanics(t, func() { _ = gs.Encode() })
}
```

One simple fix is to skip empty UUIDSets while counting/encoding (or at least guard `len()==0`).

mysql/mysql_interval.go
```diff
@@ -0,0 +1,200 @@
...
+		} else {
+			stop := max(last.Stop, s[i].Stop)
+			n[len(n)-1] = Interval{last.Start, stop, last.Tag}
+		}
+	}
+
+	return n
+}
+
+func (s *IntervalSlice) InsertInterval(interval Interval) {
```
lance6716, line 133: `InsertInterval` currently merges intervals purely based on `Start/Stop` and does not consider `Tag`. If it is used with tagged intervals, it can merge across tags and effectively drop/blur tag information.

Example:
- existing: `[{Start:1, Stop:3, Tag:""}]`
- insert: `{Start:2, Stop:4, Tag:"foo"}`
- current behavior: merges into a single interval with `Tag==""` (tag lost)

Suggested unit test (fails on this PR):

```go
func TestIntervalSlice_InsertInterval_DoesNotMergeDifferentTags(t *testing.T) {
    s := IntervalSlice{{Start: 1, Stop: 3, Tag: ""}}
    s.InsertInterval(Interval{Start: 2, Stop: 4, Tag: "foo"})

    // Expect: tags keep intervals disjoint; order can be normalized.
    require.Equal(t,
        IntervalSlice{{Start: 1, Stop: 3, Tag: ""}, {Start: 2, Stop: 4, Tag: "foo"}}.Normalize(),
        s.Normalize(),
    )
}
```

This also impacts `MysqlGTIDSet.AddGTIDWithTag` which calls `Intervals.InsertInterval(...)` with `Tag` set; a higher-level regression test could assert that adding `foo:6` to `uuid:1-5:foo:10-12` yields `uuid:1-5:foo:6:10-12` (and not `uuid:1-6:foo:10-12`).

Potential fixes: make `InsertInterval` tag-aware, or document/ensure it is only called on single-tag slices (and avoid it for tagged GTIDs).

mysql/mysql_uuidset.go
```diff
@@ -0,0 +1,305 @@
...
+	var err error
+
+	for pos < len(data) {
+		if s.SID == uuid.Nil {
+			if s.SID, err = uuid.FromBytes(data[0:16]); err != nil {
+				return 0, err
+			}
+		} else if pos+16 > len(data) {
+			return pos, errors.New("invalid uuid set buffer, expected 16 bytes beyond pos")
+		} else if !bytes.Equal(data[0:16], data[pos:pos+16]) {
```
lance6716, line 230: codex says 

This works when decoding into a fresh UUIDSet, but it doesn’t validate against a pre-set s.SID and can yield an inconsistent UUIDSet if Decode is called on a non-zero receiver. A more robust pattern is: read UUID at pos, set s.SID if nil, else compare UUID == s.SID, and stop/return when it changes.

I'll take a look soon and understand it

replication/event.go
```diff
@@ -313,7 +282,7 @@ func (e *PreviousGTIDsEvent) Decode(data []byte) error {
 		uuid := e.decodeUuid(data[pos : pos+16])
 		pos += 16
 		var tag string
-		if format == GtidFormatTagged {
+		if format == mysql.GtidFormatTagged {
 			tagLength := int(data[pos]) / 2
 			pos += 1
 			if tagLength > 0 { // 0 == no tag, >0 == tag
```
lance6716, line 288: in below lines, codex says

"PreviousGTIDsEvent.Decode reads the interval count with binary.LittleEndian.Uint16(data[pos:pos+8]) even though the count field is 8 bytes in the COM_BINLOG_DUMP_GTID style encoding (should be Uint64/uint64). Also, the string reconstruction prints the UUID only when tag==\"\"; if the first TSID for a UUID has a non-empty tag (no untagged group), this would emit a leading :tag... without the UUID. Worth adding a regression test for 'UUID has only tagged intervals' and adjusting the builder logic accordingly."

lance6716 (commented):
I'll also check codex comments later

mysql/mysql_uuidset.go
```diff
@@ -0,0 +1,305 @@
...
+// - Removing any length of leading and trailing whitespace (tabs, spaces).
+// - Lowercasing the tag
+var tagRegexp = regexp.MustCompile(`^[a-z_][a-z0-9_]{0,31}$`)
+
+// Refer:
+// - https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html
+// - https://bugs.mysql.com/bug.php?id=116789
+type UUIDSet struct {
+	SID       uuid.UUID
+	Intervals IntervalSlice
```
lance6716, line 27: Oh, I see this is not a nested map (which I mentioned in https://github.com/go-mysql-org/go-mysql/pull/1120#issuecomment-4067374623)

Maybe there are some difficulties in current code, because we didn't have a map key of tag? And tag is repeated many times every Interval

mysql/mysql_interval.go
```diff
@@ -0,0 +1,178 @@
...
+	*s = append(*s, interval)
+	*s = s.Normalize()
+}
+
+// Contain returns true if sub in s
+func (s IntervalSlice) Contain(sub IntervalSlice) bool {
+	j := 0
+	for i := range sub {
+		for ; j < len(s); j++ {
+			if sub[i].Tag != s[j].Tag {
```
lance6716, line 143: If there's another map level of `Tag`, it could be simplified

PR #1117, opened 2026-03-12
replication: decode unsigned integer types correctly in RowsEvent

yanhao-pro:
decodeValue now accepts an isUnsigned flag and returns the appropriate unsigned Go type (uint8/uint16/uint32/uint64) for TINY/SHORT/INT24/LONG/ LONGLONG columns. decodeImage derives the flag per column from the TableMapEvent UnsignedMap so values above the signed maximum are no longer misrepresented as negative numbers.


lance6716 (approved):
A helpful improvement!

PR #1119, opened 2026-03-13
replication: fix IsNumericColumn to include MYSQL_TYPE_YEAR and MYSQL_TYPE_DECIMAL

### LINKED ISSUE #1118: replication: IsNumericColumn missing MYSQL_TYPE_YEAR causes silent off-by-one in UnsignedMap() for tables with YEAR columns
---
## Bug: `IsNumericColumn` excludes `MYSQL_TYPE_YEAR`, causing `UnsignedMap()` to return wrong signedness for all numeric columns after a `YEAR` column

**Library version:** `v1.11.0`
**File:** `replication/row_event.go`

---

### Summary

`IsNumericColumn()` does not include `MYSQL_TYPE_YEAR`, but MySQL server **always** writes a signedness bit for `YEAR` columns in the `TABLE_MAP_EVENT`'s `SignednessBitmap`. Because `UnsignedMap()` only consumes a bitmap bit when `IsNumericColumn()` returns `true`, it silently skips the `YEAR` bit — causing every numeric column that comes **after** a `YEAR` column in the table to be assigned the **wrong** signedness value. This is a silent off-by-one bug in bitmap traversal.

---

### Reproduction: Table schema

```sql
CREATE TABLE t (
  id        INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
  birth_year YEAR,                   -- YEAR column in the middle
  salary    INT UNSIGNED,            -- numeric column AFTER year
  score     DECIMAL(10,2) UNSIGNED
);
```

With this schema, MySQL writes a `SignednessBitmap` with bits for columns at positions:

| Column index | Column | Bit in bitmap | Expected by `UnsignedMap()` |
|---|---|---|---|
| 0 | `id` (INT UNSIGNED) | bit 0 → `1` (unsigned) | ✅ correct |
| 1 | `birth_year` (YEAR) | bit 1 → `1` (YEAR is always unsigned internally) | **skipped** by `IsNumericColumn` |
| 2 | `salary` (INT UNSIGNED) | bit 2 → `1` | reads bit 1 → gets wrong value |
| 3 | `score` (DECIMAL UNSIGNED) | bit 3 → `1` | reads bit 2 → gets wrong value |

The result: `UnsignedMap()` returns `salary` and `score` as **signed** even though they are `UNSIGNED`.

---

#### Step 1: MySQL server writes the bitmap (`sql/field_common_properties.h`)

MySQL decides which columns get a bit using `has_signedess_information_type()`:

```cpp
// sql/field_common_properties.h
// https://github.com/mysql/mysql-server/blob/8.0/sql/field_common_properties.h

/**
 Test if the field type contains information on being signed/unsigned.
 This includes numeric but also YEAR that still contains sign modifiers
 even if ignored.
*/
inline bool has_signedess_information_type(enum_field_types type) {
  switch (type) {
    case MYSQL_TYPE_TINY:
    case MYSQL_TYPE_SHORT:
    case MYSQL_TYPE_INT24:
    case MYSQL_TYPE_LONG:
    case MYSQL_TYPE_LONGLONG:
    case MYSQL_TYPE_YEAR:          // ← YEAR gets a bit
    case MYSQL_TYPE_FLOAT:
    case MYSQL_TYPE_DOUBLE:
    case MYSQL_TYPE_DECIMAL:
    case MYSQL_TYPE_NEWDECIMAL:
      return true;
    default:
      return false;
  }
}
```

This is called from `is_numeric_field()` in `sql/log_event.cc`, which drives `init_signedness_field()`:

```cpp
// sql/log_event.cc — init_signedness_field()
for (auto field : *m_column_view) {
  if (is_numeric_field(field)) {       // true for YEAR
    Field_num *field_num = dynamic_cast<Field_num *>(field);
    if (field_num->is_unsigned()) flag |= mask;
    mask >>= 1;   // ← bit is consumed/written for YEAR
    ...
  }
}
```

MySQL's own `print_columns()` in `log_event.cc` also confirms this — it reads the YEAR bit but suppresses its display:

```cpp
if (*signedness_it == true &&
    // the UNSIGNED modifier is encoded for YEAR but not used
    field_type_code != MYSQL_TYPE_YEAR)
  my_b_printf(file, " UNSIGNED");
signedness_it++;   // ← iterator always advances — YEAR bit IS in the bitmap
```

The docstring on `has_signedess_information_type` explicitly states:
> *"This includes numeric but also YEAR that still contains sign modifiers even if ignored."*

#### Step 2: `go-mysql` stores the raw bitmap bytes unchanged

```go
// replication/row_event.go:266-267
case TABLE_MAP_OPT_META_SIGNEDNESS:
    e.SignednessBitmap = v  // raw bytes from the wire — includes YEAR's bit
```

#### Step 3: `UnsignedMap()` iterates the bitmap

```go
// replication/row_event.go:619-638
func (e *TableMapEvent) UnsignedMap() map[int]bool {
    if len(e.SignednessBitmap) == 0 {
        return nil
    }
    ret := make(map[int]bool)
    i := 0
    for _, field := range e.SignednessBitmap {
        for c := 0x80; c != 0; {
            if e.IsNumericColumn(i) {
                ret[i] = field&byte(c) != 0
                c >>= 1   // ← bit consumed ONLY when IsNumericColumn is true
            }
            i++
            if i >= int(e.ColumnCount) {
                return ret
            }
        }
    }
    return ret
}
```

The critical invariant: **`c >>= 1` (bit consumption) happens only when `IsNumericColumn(i)` is `true`**. Column index `i` always advances, but the bitmap cursor `c` only advances for columns `go-mysql` considers numeric.

#### Step 4: `IsNumericColumn()` excludes `MYSQL_TYPE_YEAR`

```go
// replication/row_event.go:792-807
func (e *TableMapEvent) IsNumericColumn(i int) bool {
    switch e.realType(i) {
    case MYSQL_TYPE_TINY,
        MYSQL_TYPE_SHORT,
        MYSQL_TYPE_INT24,
        MYSQL_TYPE_LONG,
        MYSQL_TYPE_LONGLONG,
        MYSQL_TYPE_NEWDECIMAL,
        MYSQL_TYPE_FLOAT,
        MYSQL_TYPE_DOUBLE:
        return true    // MYSQL_TYPE_YEAR is absent ← bug

    default:
        return false
    }
}
```

`MYSQL_TYPE_YEAR` is missing. So when the iterator hits a `YEAR` column, `c` is **not advanced**, but the MySQL bitmap has already consumed a bit for it. Every subsequent numeric column reads from the wrong bit position.

---

### Visual: what happens with our example table

```
Bitmap bits (MySQL wrote):   [id=1] [year=1] [salary=1] [score=1]
                               bit0    bit1     bit2      bit3

go-mysql cursor c:
  i=0  id         IsNumeric=true  → reads bit0 ✅, c shifts
  i=1  birth_year IsNumeric=false → reads nothing, c does NOT shift  ← desync
  i=2  salary     IsNumeric=true  → reads bit1 ❌ (should be bit2)
  i=3  score      IsNumeric=true  → reads bit2 ❌ (should be bit3)
```

---

### Fix

Add `MYSQL_TYPE_YEAR` to `IsNumericColumn()` to match MySQL's `has_signedess_information_type`:

```go
func (e *TableMapEvent) IsNumericColumn(i int) bool {
    switch e.realType(i) {
    case MYSQL_TYPE_TINY,
        MYSQL_TYPE_SHORT,
        MYSQL_TYPE_INT24,
        MYSQL_TYPE_LONG,
        MYSQL_TYPE_LONGLONG,
        MYSQL_TYPE_NEWDECIMAL,
        MYSQL_TYPE_FLOAT,
        MYSQL_TYPE_DOUBLE,
        MYSQL_TYPE_YEAR: // added: MySQL encodes a signedness bit for YEAR
        return true

    default:
        return false
    }
}
```

Note: `MYSQL_TYPE_DECIMAL` may also need to be added (it appears in `has_signedess_information_type` but not in `IsNumericColumn`) — worth verifying.

---

### References

- MySQL source — `has_signedess_information_type`: [`sql/field_common_properties.h`](https://github.com/mysql/mysql-server/blob/8.0/sql/field_common_properties.h)
- MySQL source — `init_signedness_field` + `print_columns`: `sql/log_event.cc`
- Protocol definition — `SIGNEDNESS` TLV: [`libbinlogevents/include/rows_event.h`](https://github.com/mysql/mysql-server/blob/8.0/libbinlogevents/include/rows_event.h) line 421–427
- Official MySQL blog: [More Metadata Is Written Into Binary Log](https://dev.mysql.com/blog-archive/more-metadata-is-written-into-binary-log/)
- go-mysql `UnsignedMap()`: `replication/row_event.go:619`
- go-mysql `IsNumericColumn()`: `replication/row_event.go:792`
- go-mysql `decodeOptionalMeta()`: `replication/row_event.go:251`

vikaxsh:
IsNumericColumn() was missing MYSQL_TYPE_YEAR and MYSQL_TYPE_DECIMAL, causing UnsignedMap() to silently return the wrong signedness for every numeric column appearing after a YEAR column in the table. MySQL writes a SignednessBitmap bit for these types (as documented in has_signedess_information_type() in sql/field_common_properties.h), but go-mysql's bitmap cursor was not advancing for them — producing a silent off-by-one desync. This PR adds both types to IsNumericColumn() to match MySQL's behaviour, along with a regression test that directly reproduces the bug.

fixes #1118

### DIFF
--- a/replication/row_event_test.go
+++ b/replication/row_event_test.go
@@ -1279,7 +1279,7 @@ func TestTableMapHelperMaps(t *testing.T) {
 	// Index 18 is `t_year` (YEAR). MySQL writes a signedness bitmap bit for
 	// YEAR (has_signedess_information_type includes YEAR), so IsNumericColumn
 	// must include YEAR to keep the bitmap cursor in sync.
-	// MySQL 8.0 sets the YEAR bit to 0 (false); MariaDB 10.5 sets it to 1 (true).
+	// MySQL sets the YEAR bit to 0 (false); MariaDB sets it to 1 (true).
 	mysqlUnsignedMap := map[int]bool{}
 	mariadbUnsignedMap := map[int]bool{}
 	for i := 1; i <= 9; i++ {
@@ -1436,7 +1436,7 @@ func TestUnsignedMapWithYearColumn(t *testing.T) {
 			mysql.MYSQL_TYPE_NEWDECIMAL,
 		},
 		ColumnMeta:       []uint16{0, 0, 0, (10 << 8) | 2},
-		SignednessBitmap: []byte{0xF0}, // bits: id=1 year=1 salary=1 score=1
+		SignednessBitmap: []byte{0xB0}, // bits: id=1 year=0 salary=1 score=1
 	}
 
 	got := e.UnsignedMap()
@@ -1445,10 +1445,10 @@ func TestUnsignedMapWithYearColumn(t *testing.T) {
 	// id (0), birth_year (1), salary (2), score (3) are all numeric.
 	// All four are unsigned in the bitmap.
 	want := map[int]bool{
-		0: true, // id        – unsigned
-		1: true, // birth_year – YEAR bit present in bitmap (always unsigned)
-		2: true, // salary    – unsigned (was wrongly false before the fix)
-		3: true, // score     – unsigned (was wrongly false before the fix)
+		0: true,  // id        – unsigned
+		1: false, // birth_year – YEAR bit present in bitmap (always unsigned)
+		2: true,  // salary    – unsigned (was wrongly false before the fix)
+		3: true,  // score     – unsigned (was wrongly false before the fix)
 	}
 	require.Equal(t, want, got)
 }
--- a/replication/row_event_test.go
+++ b/replication/row_event_test.go
@@ -1275,13 +1275,23 @@ func TestTableMapHelperMaps(t *testing.T) {
 		);
 	*/
 
-	unsignedMap := map[int]bool{}
+	// Indices 1-9: signed numerics; 10-17: unsigned numerics.
+	// Index 18 is `t_year` (YEAR). MySQL writes a signedness bitmap bit for
+	// YEAR (has_signedess_information_type includes YEAR), so IsNumericColumn
+	// must include YEAR to keep the bitmap cursor in sync.
+	// MySQL 8.0 sets the YEAR bit to 0 (false); MariaDB 10.5 sets it to 1 (true).
+	mysqlUnsignedMap := map[int]bool{}
+	mariadbUnsignedMap := map[int]bool{}
 	for i := 1; i <= 9; i++ {
-		unsignedMap[i] = false
+		mysqlUnsignedMap[i] = false
+		mariadbUnsignedMap[i] = false
 	}
 	for i := 10; i <= 17; i++ {
-		unsignedMap[i] = true
+		mysqlUnsignedMap[i] = true
+		mariadbUnsignedMap[i] = true
 	}
+	mysqlUnsignedMap[18] = false
+	mariadbUnsignedMap[18] = true
 
 	// collation id | collatation
 	//     28       | gbk_chinese_ci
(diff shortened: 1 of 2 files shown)


serprex (approved):

replication/row_event_test.go
```diff
@@ -1389,6 +1399,60 @@ func TestTableMapHelperMaps(t *testing.T) {
...
+// Schema used to generate the test data (MySQL 8.0):
+//
+//	CREATE TABLE t (
+//	  id         INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+//	  birth_year YEAR,
+//	  salary     INT UNSIGNED,
+//	  score      DECIMAL(10,2) UNSIGNED
+//	);
+func TestUnsignedMapWithYearColumn(t *testing.T) {
+	// TableMapEvent for the schema above captured from a real MySQL 8.0 instance.
```
lance6716, line 1418: I see you said

> MySQL 8.0 sets the YEAR bit to 0

But in this test, you have captured

> 1: true, // birth_year – YEAR bit present in bitmap (always unsigned)

🤔

replication/row_event_test.go
```diff
@@ -1275,13 +1275,23 @@ func TestTableMapHelperMaps(t *testing.T) {
 		);
 	*/
 
-	unsignedMap := map[int]bool{}
+	// Indices 1-9: signed numerics; 10-17: unsigned numerics.
+	// Index 18 is `t_year` (YEAR). MySQL writes a signedness bitmap bit for
+	// YEAR (has_signedess_information_type includes YEAR), so IsNumericColumn
+	// must include YEAR to keep the bitmap cursor in sync.
+	// MySQL 8.0 sets the YEAR bit to 0 (false); MariaDB 10.5 sets it to 1 (true).
```
lance6716: codex did a simple research, and said that more that all MySQL version set to 0 and MariaDB set to 1. So I think better not to mention a specific version

lance6716 (commented):
rest lgtm

replication/row_event_test.go
vikaxsh: fixed

replication/row_event_test.go
vikaxsh: done

lance6716 (approved):

PR #1120, opened 2026-03-16
mysql: Parsing UUIDSet with tags (alternative implementation)

### LINKED ISSUE #845: MySQL 8.3 will introduce new GTID format

dveeden:
Closes #845 

Similar to #1116 

The difference is this:

Take `de278ad0-2106-11e4-9f8e-6edd0ca20947:1-2:mytag:1-10`:
- In #1116 this is considered a single SID (uuid) with two intervals: `1-2` and `mytag:1-10`.
- In this PR this is considered two TSIDs (uuid+tag), one with a empty tag and `1-2` and the other with `mytag` and `1-10`

Note that the text representation suggests that for a single UUID all intervals and tags belong together. However in the binary representation there is always a `<uuid><tag>` combination, which suggests that they key should be the uuid+tag.

This PR:
- Adds a `TSID` type to store the UUID+Tag.
- Changes the key for the `map` that `MysqlGTIDSet` uses from a string to a `TSID`

API changes

```
+func (s *MysqlGTIDSet) AddGTIDWithTag(tsid TSID, gno int64) {
+func (s *MysqlGTIDSet) AddSets(sets []UUIDSet) {
+func (s *UUIDSet) Decode(data []byte, format GtidFormat) error {
+func (s *UUIDSet) Encode(format GtidFormat) []byte {
+func (t *TSID) MarshalBinary() ([]byte, error) {
+func (t *TSID) String() string {
-func NewUUIDSet(sid uuid.UUID, in ...Interval) *UUIDSet {
+func NewUUIDSet(tsid TSID, in ...Interval) *UUIDSet {
+func DecodeSid(data []byte) (format GtidFormat, sidnr uint64) {
+func ParseUUIDSets(str string) ([]UUIDSet, error) {
```

### DIFF
--- a/mysql/const.go
+++ b/mysql/const.go
@@ -270,3 +270,10 @@ const (
 	SESSION_TRACK_TRANSACTION_CHARACTERISTICS
 	SESSION_TRACK_TRANSACTION_STATE
 )
+
+type GtidFormat int
+
+const (
+	GtidFormatClassic GtidFormat = iota
+	GtidFormatTagged
+)
--- a/mysql/mysql_gtid.go
+++ b/mysql/mysql_gtid.go
@@ -4,10 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"io"
-	"math"
-	"sort"
-	"strconv"
+	"slices"
 	"strings"
 
 	"github.com/go-mysql-org/go-mysql/utils"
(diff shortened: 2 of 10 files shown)


lance6716:
> Take `de278ad0-2106-11e4-9f8e-6edd0ca20947:1-2:mytag:1-10`:
> 
> * In [mysql: Parsing UUIDSet with tags #1116](https://github.com/go-mysql-org/go-mysql/pull/1116) this is considered a single SID (uuid) with two intervals: `1-2` and `mytag:1-10`.
> * In this PR this is considered two TSIDs (uuid+tag), one with a empty tag and `1-2` and the other with `mytag` and `1-10`

I think the duplication (of the uuid part) should be avoided, so the first implementation seems better 🤔

> in the binary representation there is always a <uuid><tag> combination, which suggests that they key should be the uuid+tag.

This fact shows MySQL chooses the other one, maybe the overhead of repeating a UUID (16 bytes) is acceptable, and writing simple code is more important? Or in C/C++ a nested map (UUID -> tag -> interval) is worse than single-level map (UUID+tag -> interval)?

After all, I still think we should choose (UUID -> tag -> interval). Should follow the correct concept, and golang's builtin map is easier to use than C/C++.

dveeden:
@lance6716 

Originally we had this:

```go
type Interval struct {
        Start int64
        Stop int64
}

type IntervalSlice []Interval

type UUIDSet struct {
        SID uuid.UUID
        Intervals IntervalSlice
}

type MysqlGTIDSet struct {
        Sets map[string]*UUIDSet
}
```

Then in #1116 (`tagged_gtid`) `Tag` was added to `Interval`:

```go
type Interval struct {
        Start int64
        Stop int64
        Tag string
}

type IntervalSlice []Interval

type UUIDSet struct {
        SID       uuid.UUID
        Intervals IntervalSlice
}

type MysqlGTIDSet struct {
        Sets map[string]*UUIDSet
}
```

Then in #1120 (`tagged_gtid_v2`) we have the following. This changes the key of `MysqlGTIDSet` from SID.String() to `TSID`.

```go
type Interval struct {
        Start int64
        Stop int64
}

type IntervalSlice []Interval

type TSID struct {
        SID uuid.UUID
        Tag string
}

type UUIDSet struct {
        TSID      TSID  
        Intervals IntervalSlice
}  

type  struct {
        Sets map[TSID]*UUIDSet
}
```

What you are suggesting is something like this?

```go
type Interval struct {
        Start int64
        Stop int64
}

type IntervalSlice []Interval

type UUIDSet struct {
        SID uuid.UUID
        Intervals IntervalSlice
}

type MysqlGTIDSet struct {
        Sets map[string]map[string]*UUIDSet
}
```

Example:
```go
func main() {
	x := MysqlGTIDSet{
		Sets: map[string]map[string]*UUIDSet{
			"de278ad0-2106-11e4-9f8e-6edd0ca20947": {
				"": &UUIDSet{
					SID: uuid.MustParse("de278ad0-2106-11e4-9f8e-6edd0ca20947"),
					Intervals: IntervalSlice{
						Interval{1, 3},
					},
				},
				"mytag": &UUIDSet{
					SID: uuid.MustParse("de278ad0-2106-11e4-9f8e-6edd0ca20947"),
					Intervals: IntervalSlice{
						Interval{1, 11},
					},
				},
			},
		},
	}

	fmt.Printf("x = %#v\n", x)
}
```

Note that the UUID is used in string version as key for the map, but also in the `SID`. Another issue
is that the `UUIDSet` has the intervals and the SID, but doesn't have the Tag.

Another issue I see is that using string in multiple places instead of a more specific type doesn't let
the compiler help with type checks.

Another option is this:

```go
type Interval struct {
        Start int64
        Stop int64
}

type IntervalSlice []Interval

type tag string

type MysqlGTIDSet map[uuid.UUID]map[tag]IntervalSlice
```

With this as example:
```go
func main() {
	x := MysqlGTIDSet{
		uuid.MustParse("de278ad0-2106-11e4-9f8e-6edd0ca20947"): {
			"": IntervalSlice{
				Interval{1, 3},
			},
			"mytag": IntervalSlice{
				Interval{1, 11},
			},
		},
	}

	fmt.Printf("x = %#v\n", x)
}
```

What do you think is best?

lance6716:
```go
type Interval struct {
        Start int64
        Stop int64
}

type IntervalSlice []Interval

type Tag string

type MysqlGTIDSet map[uuid.UUID]map[Tag]IntervalSlice
```

I prefer this one

dveeden:
> ```go
> type Interval struct {
>         Start int64
>         Stop int64
> }
> 
> type IntervalSlice []Interval
> 
> type Tag string
> 
> type MysqlGTIDSet map[uuid.UUID]map[Tag]IntervalSlice
> ```
> 
> I prefer this one

Ok, I'll create a separate PR for this. For now I'll mark the existing two PRs as draft.

PR #1121, opened 2026-03-26
client/server: preserve OK payload for procedure multi-results

nzin-alayacare:
Context: using go-mysql to create a mysql proxy server (with a 8.0 target mysql server), I encountered 2 issues

- I couldn't drain properly some multi-results connections
- doing a store procedure call with an out parameter through Connector/J was not working.

So this PR is about: keep trailing OK packet payload bytes available for proxy passthrough and expose a statement helper to drain stored procedure multi-results safely, improving Connector/J CALL + OUT compatibility.

### DIFF
--- a/client/ok_payload_test.go
+++ b/client/ok_payload_test.go
@@ -1,0 +1,26 @@
+package client
+
+import (
+	"encoding/binary"
+	"testing"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHandleOKPacketPreservesSuffix(t *testing.T) {
+	c := &Conn{}
+	c.capability = mysql.CLIENT_PROTOCOL_41 | mysql.CLIENT_SESSION_TRACK
+
+	data := []byte{mysql.OK_HEADER}
+	data = append(data, mysql.PutLengthEncodedInt(0)...)
+	data = append(data, mysql.PutLengthEncodedInt(0)...)
+	data = binary.LittleEndian.AppendUint16(data, 0) // status
+	data = binary.LittleEndian.AppendUint16(data, 0) // warnings
+	suffix := []byte{0x01, 0xfe, 0xaa, 0xbb}
+	data = append(data, suffix...)
+
+	r, err := c.handleOKPacket(data)
+	require.NoError(t, err)
+	require.Equal(t, suffix, r.OKPayloadSuffix)
+}
--- a/client/resp.go
+++ b/client/resp.go
@@ -46,6 +46,12 @@ func (c *Conn) handleOKPacket(data []byte) (*mysql.Result, error) {
 		pos += 2
 	}
 
+	// Preserve trailing OK payload (info + session tracking bytes) for
+	// protocol-transparent proxy forwarding.
+	if pos < len(data) {
+		r.OKPayloadSuffix = append([]byte(nil), data[pos:]...)
+	}
+
 	if (c.capability&mysql.CLIENT_SESSION_TRACK > 0) &&
 		(c.status&mysql.SERVER_SESSION_STATE_CHANGED > 0) {
 		var err error
--- a/client/stmt.go
+++ b/client/stmt.go
@@ -49,6 +49,44 @@ func (s *Stmt) ExecuteSelectStreaming(result *mysql.Result, perRowCb SelectPerRo
 	return s.conn.readResultStreaming(true, result, perRowCb, perResCb)
 }
 
+// StmtProcedureMultiResultForward is called for each logical result returned by
+// COM_STMT_EXECUTE. When err is non-nil, res is nil.
+type StmtProcedureMultiResultForward func(res *mysql.Result, err error) error
+
+// ExecuteProcedureMultiResults runs COM_STMT_EXECUTE and drains all procedure
+// results until SERVER_MORE_RESULTS_EXISTS is no longer set.
+func (s *Stmt) ExecuteProcedureMultiResults(args []any, forward StmtProcedureMultiResultForward) (*mysql.Result, error) {
+	if err := s.write(args...); err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	var forwardErr error
+	for {
+		res, err := s.conn.readResult(true)
+		if forwardErr == nil {
+			if err != nil {
+				forwardErr = forward(nil, err)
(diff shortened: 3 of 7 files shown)


dveeden:
So we don't decode what's in `OKPayloadSuffix`? I guess that's ok for most 1-to-1 proxy implementations, but maybe we should decode the contents so the proxy itself can use it?

dveeden:
For session tracking: This is what we don on the client side: https://github.com/go-mysql-org/go-mysql/blob/2271d4c5af8df282db7191355a28a3417b8d1bb3/client/resp.go#L79

dveeden:
Do you have packet traces and/or Wireshark screenshots showing what's being send in these cases by Connector/J? What version of connj are you using?

nzin-alayacare:
> Do you have packet traces and/or Wireshark screenshots showing what's being send in these cases by Connector/J? What version of connj are you using?

Thanks for the quick answer.

I dont have (today) wireshark traces or screenshots, but I have a sample that reproduced the issue on my proxy sql implementation.
- I am using mysql-connector-j version 8.4.0
- the code is doing a "simple" jdbc connection, and try to call a stored procedure like "{call my_stored_procedure(?,?,...,?)}";
- the important point is that the last parameter is an OUT parameter
- without the fix I received a "SQL Error [S1000]: Cannot invoke "Object.getClass()" because "currentProtocolEntity" is null"
- the problem is mostly for procedure calls, not for regular queries


I can spend some time to reproduce into a different test suite, but I would need some guidance in which context/component testing suite I should try to reproduce the issue

client/ok_payload_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"encoding/binary"
+	"testing"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHandleOKPacketPreservesSuffix(t *testing.T) {
```
lance6716, line 11: this test is short, better to move into existing test files

client/resp.go
```diff
@@ -46,6 +46,12 @@ func (c *Conn) handleOKPacket(data []byte) (*mysql.Result, error) {
 		pos += 2
 	}
 
+	// Preserve trailing OK payload (info + session tracking bytes) for
+	// protocol-transparent proxy forwarding.
+	if pos < len(data) {
+		r.OKPayloadSuffix = append([]byte(nil), data[pos:]...)
```
lance6716, line 52: Did you check that we don't need to clone that data, which is to say, using a reference to remaining slice is OK?

client/stmt.go
```diff
@@ -49,6 +49,48 @@ func (s *Stmt) ExecuteSelectStreaming(result *mysql.Result, perRowCb SelectPerRo
...
+
+	if err := s.write(args...); err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	var forwardErr error
+	for {
+		res, err := s.conn.readResult(true)
+		if forwardErr == nil {
+			if err != nil {
```
lance6716, line 71: Can we directly use `forwardErr = forward(res, err)`?

client/stmt.go
```diff
@@ -49,6 +49,48 @@ func (s *Stmt) ExecuteSelectStreaming(result *mysql.Result, perRowCb SelectPerRo
...
+		}
+		if res == nil || res.Status&mysql.SERVER_MORE_RESULTS_EXISTS == 0 {
+			break
+		}
+	}
+
+	if forwardErr != nil {
+		return nil, forwardErr
+	}
+	rs := mysql.NewResultset(1)
```
lance6716, line 88: can you add some comments to explain how to use this reselt set?

server/server_conf.go
```diff
@@ -54,7 +54,8 @@ func NewDefaultServer() *Server {
 		protocolVersion: 10,
 		capability: mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_LONG_FLAG | mysql.CLIENT_CONNECT_WITH_DB | mysql.CLIENT_PROTOCOL_41 |
 			mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_SECURE_CONNECTION | mysql.CLIENT_PLUGIN_AUTH | mysql.CLIENT_SSL |
-			mysql.CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA | mysql.CLIENT_CONNECT_ATTRS,
+			mysql.CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA | mysql.CLIENT_CONNECT_ATTRS |
+			mysql.CLIENT_MULTI_RESULTS | mysql.CLIENT_PS_MULTI_RESULTS | mysql.CLIENT_SESSION_TRACK,
```
lance6716, line 58: I think if the proxy should not always tell its clients CLIENT_SESSION_TRACK. If the upstream of proxy does not enable CLIENT_SESSION_TRACK, proxy itself should add suffix to satisfy CLIENT_SESSION_TRACK, which make thing more complicated 🤔

dveeden: Yes, if this flag is enabled the proxy must make sure the right protocol is used. It does not have to send session tracking, but it might need to send an empty list etc. if the flag is enabled.

dveeden:
> > Do you have packet traces and/or Wireshark screenshots showing what's being send in these cases by Connector/J? What version of connj are you using?
> 
> Thanks for the quick answer.
> 
> I dont have (today) wireshark traces or screenshots, but I have a sample that reproduced the issue on my proxy sql implementation.

If you could record some traffic with `tcpdump` or some other tool (filter on port 3306, disable SSL/TLS, don't use any passwords that must remain secret) this could be useful.


>     * I am using mysql-connector-j version 8.4.0
> 
>     * the code is doing a "simple" jdbc connection, and try to call a stored procedure like "{call my_stored_procedure(?,?,...,?)}";
> 
>     * the important point is that the last parameter is an OUT parameter
> 
>     * without the fix I received a "SQL Error [S1000]: Cannot invoke "Object.getClass()" because "currentProtocolEntity" is null"
> 
>     * the problem is mostly for procedure calls, not for regular queries
> 
> 
> I can spend some time to reproduce into a different test suite, but I would need some guidance in which context/component testing suite I should try to reproduce the issue

If you have a simple standalone test that I can run that would be useful.

Note that none of this is required to get this merged. The only thing that is required is that you resolve the review comments.

PR #1122, opened 2026-03-28
fix: MariadbGTIDSet.AddSet() replace server entry on same domain to avoid duplicate domain id error

triplefighter7619:
## Problem

`MariadbGTIDSet.AddSet()` stores multiple server IDs per domain in its
`map[uint32]map[uint32]*MariadbGTID` structure. When a MariaDB master failover changes
the server ID within the same domain, `AddSet()` adds a new server entry alongside
the old one instead of replacing it.

This causes `String()` to output GTIDSet strings like `0-1013-100,0-963-200` with
duplicate domain ID 0. MariaDB rejects this in `START SLAVE UNTIL MASTER_GTID_POS`
with error #1943: `GTID 0-963-200 and 0-1013-100 conflict (duplicate domain id 0)`.

## Fix

When `AddSet()` encounters a GTID with the same domain but different server ID,
clear existing server entries before adding the new one. MariaDB only allows one
GTID per domain in `gtid_slave_pos` / `MASTER_GTID_POS`.

## Test

Updated `TestMariaDBGTIDSetEqual` test case to reflect correct behavior.

### DIFF
--- a/mysql/mariadb_gtid.go
+++ b/mysql/mariadb_gtid.go
@@ -102,7 +102,9 @@ func (gtid *MariadbGTID) forward(newer *MariadbGTID) error {
 	return nil
 }
 
-// MariadbGTIDSet is a set of mariadb gtid
+// MariadbGTIDSet represents a MariaDB GTID position (one GTID per domain_id).
+// Despite the name, this is a position, not a set — unlike MySQL's executed GTID set.
+// See https://github.com/go-mysql-org/go-mysql/pull/1122
 type MariadbGTIDSet struct {
 	Sets map[uint32]map[uint32]*MariadbGTID
 }
@@ -121,7 +123,9 @@ func ParseMariadbGTIDSet(str string) (GTIDSet, error) {
 	return s, nil
 }
 
-// AddSet adds mariadb gtid into mariadb gtid set
+// AddSet adds or updates a GTID position for a domain. If the domain already has
+// a GTID with a different server_id (e.g. after primary failover), the old entry
+// is replaced to maintain one position per domain_id.
 func (s *MariadbGTIDSet) AddSet(gtid *MariadbGTID) error {
 	if gtid == nil {
 		return nil
--- a/mysql/mariadb_gtid.go
+++ b/mysql/mariadb_gtid.go
@@ -132,6 +132,9 @@ func (s *MariadbGTIDSet) AddSet(gtid *MariadbGTID) error {
 			gtid.ServerID: gtid,
 		}
 	} else if o, ok := serverSets[gtid.ServerID]; !ok {
+		// Same domain, different server (e.g. master failover). MariaDB only allows one GTID per domain.
+		slog.Warn("replacing server entries for domain", slog.Uint64("domainID", uint64(gtid.DomainID)), slog.Any("new", gtid))
+		clear(serverSets)
 		serverSets[gtid.ServerID] = gtid
 	} else {
 		err := o.forward(gtid)
--- a/mysql/mariadb_gtid.go
+++ b/mysql/mariadb_gtid.go
@@ -132,7 +132,6 @@ func (s *MariadbGTIDSet) AddSet(gtid *MariadbGTID) error {
 			gtid.ServerID: gtid,
 		}
 	} else if o, ok := serverSets[gtid.ServerID]; !ok {
-		// Same domain, different server (e.g. master failover). MariaDB only allows one GTID per domain.
 		slog.Warn("replacing server entries for domain", slog.Uint64("domainID", uint64(gtid.DomainID)), slog.Any("new", gtid))
 		clear(serverSets)
 		serverSets[gtid.ServerID] = gtid
--- a/mysql/mariadb_gtid_test.go
+++ b/mysql/mariadb_gtid_test.go
@@ -179,7 +179,7 @@ func TestMariaDBGTIDSetEqual(t *testing.T) {
 		{"1-1-1,2-2-2", "1-1-1", false},
 		{"1-1-1,2-2-2", "1-1-1,2-2-2", true},
 		{"1-1-1,2-2-2", "1-1-1,2-2-3", false},
-		{"0-1-1,0-2-2", "0-2-2", false},
+		{"0-1-1,0-2-2", "0-2-2", true}, // MariaDB allows only one GTID per domain; 0-2-2 replaces 0-1-1
 	}
 
 	for _, cs := range cases {
--- a/mysql/mariadb_gtid_test.go
+++ b/mysql/mariadb_gtid_test.go
@@ -179,7 +179,7 @@ func TestMariaDBGTIDSetEqual(t *testing.T) {
 		{"1-1-1,2-2-2", "1-1-1", false},
 		{"1-1-1,2-2-2", "1-1-1,2-2-2", true},
 		{"1-1-1,2-2-2", "1-1-1,2-2-3", false},
-		{"0-1-1,0-2-2", "0-2-2", true}, // MariaDB allows only one GTID per domain; 0-2-2 replaces 0-1-1
+		{"0-1-1,0-2-2", "0-2-2", true},
 	}
 
 	for _, cs := range cases {


lance6716:
> MariaDB only allows one GTID per domain in gtid_slave_pos / MASTER_GTID_POS.

Can you provide some reference about this behaviour?

Also @GMHDBJD please help check it. I have forgotten the behaviour of MariaDB gtid now 😂

GMHDBJD (approved):

triplefighter7619:
The MariaDB official documentation explicitly states this constraint. From the [Global Transaction ID docs](https://mariadb.com/docs/server/ha-and-performance/standard-replication/gtid) under `START SLAVE UNTIL`:

```
If multiple GTIDs are specified, then they must be with distinct replication domain ID
```

And under gtid_slave_pos:

```
This system variable contains the GTID of the last transaction applied to the database by the server's slave threads for each replication domain.
```

A replica tracks its position as one GTID per domain — one position per replication stream. `START SLAVE UNTIL MASTER_GTID_POS` enforces this by requiring distinct domain IDs.

Real-world reproduction: In our environment, after a MariaDB master failover (server ID changed from 1013 to 963, same domain 0), the `AddSet()` method accumulated both server entries under domain 0. The resulting `String()` output was `0-1013-18149634291,0-963-18149657635`, which was passed to:
```
START SLAVE UNTIL MASTER_GTID_POS = '0-1013-18149634291,0-963-18149657635'
```

MariaDB rejected this with error #1943:
```
Received #1943 error from MySQL server: "GTID 0-963-18149657635 and 0-1013-18149634291 conflict (duplicate domain id 0)"
```
This caused a fatal crash in our application that uses this library for MariaDB binlog replication — the consumer could not flush transactions to the follower because the GTIDSet had two entries for domain 0.

The fix ensures that when a new server ID appears under the same domain (which is expected during master failovers), the old server entry is replaced rather than added alongside, keeping one GTID per domain.

dveeden:
What MySQL records the range of transactions that it has seen: `<uuid>:<1-100>`

MariaDB records a unique identifier for a position in the stream: So `0-1-2` for example. This doesn't explicitly say that -1-1 was executed, it does that implicitly.

So '0-1013-18149634291,0-963-18149657635' is two positions in the same stream, similar to 0-1-1,0-1-2.

When switching from serverid 963 to serverid 1013 I would expect only one of these positions (the newest) to be supplied to the `START SLAVE UNTIL MASTER_GTID_POS` statement.

MariaDB can run with multiple streams where each stream is a separate domain-id. This is with multi-master (circular) replication of multi source. But this should not be needed for primary switchovers.

dveeden:
If MariaDBGTIDSet has `0-963-18149657635` and then `MariadbGTIDSet.AddSet()` is called with `0-1013-18149634291` it probably should either give an error (duplicate domain-id). 

Another option would be for it to replace the current position, but `MariadbGTIDSet.Update()` is probably better for that.

dveeden (changes requested):
Please see my earlier comments.

triplefighter7619:
@dveeden, regarding your suggestion to either error in `AddSet()` or handle replacement in `Update()`:

**Erroring in `AddSet()` would break `binlogsyncer.go`** — it calls `AddSet(&event.GTID)` directly for every GTID event from the binlog stream:

```go
// replication/binlogsyncer.go:894-897
prev := b.currGset.Clone()
err := b.currGset.(*mysql.MariadbGTIDSet).AddSet(&event.GTID)
if err != nil {
    return errors.Trace(err)
}
```

After a primary switchover, the server ID changes within the same domain (e.g., `0-1013-100` → `0-963-101`). If `AddSet()` errors on duplicate domain, the binlog syncer would crash during normal failover scenarios.

**The replacement logic belongs in `AddSet()`**, not `Update()`, because both `Update()` and `binlogsyncer` funnel through `AddSet()`:

```go
// mysql/mariadb_gtid.go - Update() calls AddSet()
func (s *MariadbGTIDSet) Update(GTIDStr string) error {
    sp := strings.Split(GTIDStr, ",")
    for i := 0; i < len(sp); i++ {
        gtid, err := ParseMariadbGTID(sp[i])
        // ...
        err = s.AddSet(gtid)  // <-- all paths go through AddSet()
    }
}
```

Handling replacement only in `Update()` would leave `binlogsyncer`'s direct `AddSet()` calls unprotected.

**History:** PR #272 ([`cb167c0`](https://github.com/go-mysql-org/go-mysql/commit/cb167c088cc0213a613dc00ef4870d98b488598e)) introduced `MariadbGTIDSet` with a flat `map[uint32]*MariadbGTID` where `AddSet()` correctly replaced within the same domain via `forward()`:

```go
// Original AddSet() from PR #272 — flat map, one entry per domain:
func (s *MariadbGTIDSet) AddSet(gtid *MariadbGTID) error {
    o, ok := s.Sets[gtid.DomainID]
    if ok {
        o.forward(gtid)  // replaces ServerID + SequenceNumber
    } else {
        s.Sets[gtid.DomainID] = gtid
    }
    return nil
}
```

PR #852 ([`034c7ab`](https://github.com/go-mysql-org/go-mysql/commit/034c7abd13cb4f22c5e41d8179eebd8bd8d6e409), 2024-03-14 — "separate serverID of Mariadb GTID set") changed the map to `map[uint32]map[uint32]*MariadbGTID`, which inadvertently broke the same-domain replacement — `AddSet()` started adding new server entries alongside old ones instead of replacing.

This fix restores PR #272's replacement behavior while keeping PR #852's nested map structure. Happy to refactor if you have a preferred approach.

dveeden:
@triplefighter7619 

1. Thanks for the full explanation
2. I think your PR is probably the most pragmatic solution.

However this shows some problems with the code: For MariaDB `AddSet()` doesn't add a set as the name suggests, but does an add-or-update of a position. Note that `GTIDSet` indicates that this is a set, but that isn't the case on MariaDB. 

It looks like `Contains()` for MariaDB ( https://github.com/go-mysql-org/go-mysql/blob/2271d4c5af8df282db7191355a28a3417b8d1bb3/mysql/mariadb_gtid.go#L66 ) is also a problem. It checks if the domainID is the same and then compares the sequence number. However if a GTID has a different ServerID then comparing the sequence doesn't yield any useful information. The `Contains()` function makes sense for a real set, but not for a position.

For MySQL you can tell if a specific GTID was executed in the past based on whether or not it is in the executed set. For MariaDB you can't do this as the current GTID position may have a different serverId than the GTID you're comparing against (and assuming the DomainID for both being the same).

Looking at your code I think there is one issue: I think it now stored the last position for each serverID for each domainID. It think it should only store a single serverID+position per domainID. However this is probably also a pre-existing problem that maybe should be handled outside of this PR.

What I think should be done in this PR is to add a clarifying comment to explain some of this and maybe a reference to this issue as well.

cc @lance6716

dveeden (approved):

triplefighter7619:
Thanks @dveeden! I've added clarifying comments to `MariadbGTIDSet` and `AddSet()` as suggested, and opened #1123 to track the broader data model and `Contain()` issues separately.

dveeden (approved):

mysql/mariadb_gtid.go
```diff
@@ -102,7 +102,9 @@ func (gtid *MariadbGTID) forward(newer *MariadbGTID) error {
 	return nil
 }
 
-// MariadbGTIDSet is a set of mariadb gtid
+// MariadbGTIDSet represents a MariaDB GTID position (one GTID per domain_id).
+// Despite the name, this is a position, not a set — unlike MySQL's executed GTID set.
+// See https://github.com/go-mysql-org/go-mysql/pull/1122
 type MariadbGTIDSet struct {
 	Sets map[uint32]map[uint32]*MariadbGTID
```
lance6716, line 109: How about change the structure to `map[uint32]struct{serverID uint32, gtid *MariadbGTID}`? We can do it in this PR, or in the following PR that solves #1123.

lance6716 (approved):
Please check my only comment. I agree to merge this PR as-is if you think better to alter the struct in another PR @triplefighter7619 

I guess https://github.com/go-mysql-org/go-mysql/pull/852 has its reason, but I forget why 😂 This PR seems is a more correct behaviour as native MariaDB. Just hope https://github.com/go-mysql-org/go-mysql/pull/852 can have another way to fix its original problem.

mysql/mariadb_gtid.go
triplefighter7619: I agree `map[uint32]*MariadbGTID` (one entry per domain, since `MariadbGTID` already contains `ServerID`) would be the cleanest. Your suggestion of `map[uint32]struct{serverID uint32, gtid *MariadbGTID}` works too, though `MariadbGTID` already has the `ServerID` field so just `map[uint32]*MariadbGTID` should suffice.

I'd prefer to do the struct change in the follow-up PR for #1123 to keep this PR minimal and focused on the immediate fix.

PR #1124, opened 2026-03-31
refactor: flatten MariadbGTIDSet to map[uint32]*MariadbGTID (one entry per domain)

### LINKED ISSUE #1123: MariadbGTIDSet: data model stores multiple server IDs per domain; Contain() comparison is incorrect across server IDs
## Background

Identified during the fix in #1122: `MariadbGTIDSet` uses `map[uint32]map[uint32]*MariadbGTID` (domain_id → server_id → GTID), which stores multiple server entries per domain. However, MariaDB only tracks **one position per domain_id** in `gtid_slave_pos` and `MASTER_GTID_POS`.

### 1. Data model allows multiple server IDs per domain

The nested `map[serverID]*MariadbGTID` per domain can accumulate entries from different servers. While #1122 fixes `AddSet()` to clear old entries on server change, the underlying data model still permits it. Ideally, the structure should enforce one entry per domain.

**Introduced in:** PR #852 (`034c7ab` — "separate serverID of Mariadb GTID set", 2024-03-14)
**Previous behavior:** PR #272 (`cb167c0`) used a flat `map[uint32]*MariadbGTID` with one entry per domain.

### 2. `Contain()` is incorrect across server IDs

`MariadbGTID.Contain()` checks if `DomainID` matches and `SequenceNumber >= other.SequenceNumber`. However, if the two GTIDs have different `ServerID` values (e.g., after a primary failover), comparing sequence numbers is not meaningful — sequences are per-server, not globally ordered across servers within a domain.

As @dveeden noted: *"For MariaDB you can't [tell if a specific GTID was executed] as the current GTID position may have a different serverId than the GTID you're comparing against."*

### 3. Naming: `MariadbGTIDSet` is a position, not a set

Unlike MySQL's `GTIDSet` which tracks ranges of executed transactions, MariaDB's GTID represents a **position** in each replication stream. The current naming can be misleading.

## Suggested improvements

1. Consider flattening the map back to `map[uint32]*MariadbGTID` (one entry per domain)
2. Update `Contain()` to account for server ID differences
3. Consider renaming or adding documentation to clarify the position-vs-set distinction

cc @dveeden @lance6716

triplefighter7619:
Fixes #1123

## Problem

`MariadbGTIDSet` used `map[uint32]map[uint32]*MariadbGTID` (domain_id → server_id → GTID),
which allowed multiple server entries per domain. This is incorrect for MariaDB, which tracks
exactly **one GTID position per domain_id** in `gtid_slave_pos` and `MASTER_GTID_POS`.

This structure was introduced in PR #852 (`034c7ab`). Before that, PR #272 (`cb167c0`) used a
flat `map[uint32]*MariadbGTID` which correctly enforced one entry per domain.

### `mariadb_gtid.go`

- **`MariadbGTIDSet.Sets`**: flattened from `map[uint32]map[uint32]*MariadbGTID` to `map[uint32]*MariadbGTID`
- **`AddSet()`**: simplified — if domain exists, call `forward()` to update server ID + sequence; otherwise add new entry. The `forward()` method already handles server ID changes (as happens during primary failovers)
- **`String()`**, **`Encode()`**, **`Clone()`**, **`Equal()`**, **`Contain()`**: all simplified from nested iteration to single-level map iteration
- Removed unused `"bytes"` import

### `mariadb_gtid_test.go`

- Updated all test expectations from `map[uint32]map[uint32]string` to `map[uint32]string`
- Added `TestMariaDBGTIDSetSameDomainDifferentServer` with comprehensive coverage:
  - Basic server replacement (primary failover scenario)
  - Multi-domain isolation (only affected domain changes)
  - Direct `AddSet()` call (as `binlogsyncer` does)
  - Equality after forward
  - Clone independence
- Removed unused `"slices"` import

### No changes to `binlogsyncer.go`

`AddSet(&event.GTID)` passes `*MariadbGTID` — same signature, works with the flat map.

## Context

As discussed in #1122 and noted by @dveeden and @lance6716, MariaDB's GTID is a **position**
(one per domain), not a **set** (like MySQL's executed GTID set). The flat map structure
enforces this at the type level, making it impossible to accidentally accumulate multiple
server entries under the same domain.

### DIFF
--- a/mysql/mariadb_gtid.go
+++ b/mysql/mariadb_gtid.go
@@ -1,7 +1,6 @@
 package mysql
 
 import (
-	"bytes"
 	"fmt"
 	"log/slog"
 	"sort"
@@ -104,15 +103,15 @@ func (gtid *MariadbGTID) forward(newer *MariadbGTID) error {
 
 // MariadbGTIDSet represents a MariaDB GTID position (one GTID per domain_id).
 // Despite the name, this is a position, not a set — unlike MySQL's executed GTID set.
-// See https://github.com/go-mysql-org/go-mysql/pull/1122
+// See https://github.com/go-mysql-org/go-mysql/issues/1123
 type MariadbGTIDSet struct {
-	Sets map[uint32]map[uint32]*MariadbGTID
+	Sets map[uint32]*MariadbGTID
 }
 
 // ParseMariadbGTIDSet parses str into mariadb gtid sets
 func ParseMariadbGTIDSet(str string) (GTIDSet, error) {
 	s := new(MariadbGTIDSet)
-	s.Sets = make(map[uint32]map[uint32]*MariadbGTID)
+	s.Sets = make(map[uint32]*MariadbGTID)
 	if str == "" {
 		return s, nil
 	}
@@ -125,25 +124,18 @@ func ParseMariadbGTIDSet(str string) (GTIDSet, error) {
 
 // AddSet adds or updates a GTID position for a domain. If the domain already has
 // a GTID with a different server_id (e.g. after primary failover), the old entry
-// is replaced to maintain one position per domain_id.
+// is replaced via forward() to maintain one position per domain_id.
 func (s *MariadbGTIDSet) AddSet(gtid *MariadbGTID) error {
 	if gtid == nil {
 		return nil
 	}
 
-	if serverSets, ok := s.Sets[gtid.DomainID]; !ok {
-		s.Sets[gtid.DomainID] = map[uint32]*MariadbGTID{
-			gtid.ServerID: gtid,
-		}
-	} else if o, ok := serverSets[gtid.ServerID]; !ok {
-		slog.Warn("replacing server entries for domain", slog.Uint64("domainID", uint64(gtid.DomainID)), slog.Any("new", gtid))
-		clear(serverSets)
-		serverSets[gtid.ServerID] = gtid
-	} else {
-		err := o.forward(gtid)
-		if err != nil {
+	if existing, ok := s.Sets[gtid.DomainID]; ok {
+		if err := existing.forward(gtid); err != nil {
 			return errors.Trace(err)
 		}
+	} else {
+		s.Sets[gtid.DomainID] = gtid.Clone()
 	}
 
 	return nil
@@ -168,10 +160,8 @@ func (s *MariadbGTIDSet) Update(GTIDStr string) error {
 
 func (s *MariadbGTIDSet) String() string {
 	sets := make([]string, 0, len(s.Sets))
-	for _, set := range s.Sets {
-		for _, gtid := range set {
-			sets = append(sets, gtid.String())
-		}
+	for _, gtid := range s.Sets {
+		sets = append(sets, gtid.String())
 	}
 	sort.Strings(sets)
 
@@ -180,29 +170,16 @@ func (s *MariadbGTIDSet) String() string {
 
 // Encode encodes mariadb gtid set
 func (s *MariadbGTIDSet) Encode() []byte {
-	var buf bytes.Buffer
-	sep := ""
-	for _, set := range s.Sets {
-		for _, gtid := range set {
-			buf.WriteString(sep)
-			buf.WriteString(gtid.String())
-			sep = ","
-		}
-	}
-
-	return buf.Bytes()
+	return []byte(s.String())
 }
 
 // Clone clones a mariadb gtid set
 func (s *MariadbGTIDSet) Clone() GTIDSet {
 	clone := &MariadbGTIDSet{
-		Sets: make(map[uint32]map[uint32]*MariadbGTID),
+		Sets: make(map[uint32]*MariadbGTID),
 	}
-	for domainID, set := range s.Sets {
-		clone.Sets[domainID] = make(map[uint32]*MariadbGTID)
-		for serverID, gtid := range set {
-			clone.Sets[domainID][serverID] = gtid.Clone()
-		}
+	for domainID, gtid := range s.Sets {
+		clone.Sets[domainID] = gtid.Clone()
 	}
 
 	return clone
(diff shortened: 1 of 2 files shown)


mysql/mariadb_gtid.go
```diff
@@ -121,23 +122,20 @@ func ParseMariadbGTIDSet(str string) (GTIDSet, error) {
...
-		serverSets[gtid.ServerID] = gtid
-	} else {
-		err := o.forward(gtid)
-		if err != nil {
+	if existing, ok := s.Sets[gtid.DomainID]; ok {
+		if err := existing.forward(gtid); err != nil {
 			return errors.Trace(err)
 		}
+	} else {
+		s.Sets[gtid.DomainID] = gtid
```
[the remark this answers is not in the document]
triplefighter7619: Implemented this

mysql/mariadb_gtid.go
```diff
@@ -174,29 +170,24 @@ func (s *MariadbGTIDSet) String() string {
...
-			sep = ","
-		}
+	for _, gtid := range s.Sets {
+		buf.WriteString(sep)
+		buf.WriteString(gtid.String())
+		sep = ","
 	}
 
-	return buf.Bytes()
+	return []byte(buf.String())
```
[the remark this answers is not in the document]
triplefighter7619: Implemented this

dveeden (approved):

triplefighter7619:
@lance6716 can you also take a look at this

lance6716:
Sorry I'm busy these days. Maybe I'll have more spare time ~5 days later. @dveeden free to merge for now or request some AI review

triplefighter7619:
@dveeden can you please merge this PR?

mysql/mariadb_gtid_test.go
```diff
@@ -92,14 +91,16 @@ func TestMariaDBForward(t *testing.T) {
...
-		{"0-1-1", map[uint32]map[uint32]string{0: {1: "0-1-1"}}, []string{"0-1-1"}, false},
-		{"", nil, []string{""}, false},
-		{"0-1-1,1-2-3", map[uint32]map[uint32]string{0: {1: "0-1-1"}, 1: {2: "1-2-3"}}, []string{"0-1-1,1-2-3", "1-2-3,0-1-1"}, false},
-		{"0-1--1", nil, nil, true},
+		{"0-1-1", map[uint32]string{0: "0-1-1"}, "0-1-1", false},
+		{"", nil, "", false},
+		{"0-1-1,1-2-3", map[uint32]string{0: "0-1-1", 1: "1-2-3"}, "0-1-1,1-2-3", false},
+		{"0-1--1", nil, "", true},
+		// Same domain, different server — last one wins (forward replaces)
+		{"0-1-1,0-2-2", map[uint32]string{0: "0-2-2"}, "0-2-2", false},
```
dveeden, line 111: Might be good to test for a non-numeric domain-id (`x-1-1`) which should result in an error.

mysql/mariadb_gtid.go
```diff
@@ -125,25 +123,18 @@ func ParseMariadbGTIDSet(str string) (GTIDSet, error) {
...
-		}
-	} else if o, ok := serverSets[gtid.ServerID]; !ok {
-		slog.Warn("replacing server entries for domain", slog.Uint64("domainID", uint64(gtid.DomainID)), slog.Any("new", gtid))
-		clear(serverSets)
-		serverSets[gtid.ServerID] = gtid
-	} else {
-		err := o.forward(gtid)
-		if err != nil {
+	if existing, ok := s.Sets[gtid.DomainID]; ok {
+		if err := existing.forward(gtid); err != nil {
```
dveeden, line 133: Looks like we don't have testing for a GTID that causes `forward()` to return an error. This can happen if the domain-id is different.

I think this can also happen when `Update()` calls `AddSet()`

mysql/mariadb_gtid.go
dveeden, line 191: I think we are not testing this.

mysql/mariadb_gtid.go
```diff
@@ -219,21 +195,14 @@ func (s *MariadbGTIDSet) Equal(o GTIDSet) bool {
 		return false
 	}
 
-	for domainID, set := range other.Sets {
-		serverSet, ok := s.Sets[domainID]
+	for domainID, gtid := range other.Sets {
+		ourGTID, ok := s.Sets[domainID]
 		if !ok {
 			return false
```
dveeden, line 201: I think we are not testing this

mysql/mariadb_gtid.go
```diff
@@ -246,17 +215,13 @@ func (s *MariadbGTIDSet) Contain(o GTIDSet) bool {
 		return false
```
dveeden, line 215: I think we are not testing this.

dveeden (approved):
I think this is all good and I will merge this later today or tomorrow morning unless there are objections.

I had a look at the test coverage and there are a few very minor gaps that we could probably fix easily.

mysql/mariadb_gtid_test.go
dveeden: Maybe also something like `4294967295-1-2` and `4294967296-1-2`. I think this is `<uint32>-<uint32>-<uint64>`.

And I assume `1e5-1-2` doesn't work...

triplefighter7619: Added test cases for `"x-1-1"` (non-numeric domain), `"0-x-1"` (non-numeric server), `"1e5-1-2"` (scientific notation), plus uint32/uint64 boundary values (`4294967295-1-2` valid, `4294967296-1-2` overflow error, `0-1-18446744073709551616` overflow error).

mysql/mariadb_gtid.go
triplefighter7619: Added `TestMariaDBGTIDSetForwardError` — the forward() error path is unreachable through normal `AddSet()`/`Update()` usage since the map key guarantees domain match, but tested via injected mismatch for completeness. Also added `Update()` with invalid strings (`"x-1-1"`, `"0-1--1"`).

mysql/mariadb_gtid.go
triplefighter7619: Added: `mariaSet.Equal(mysqlSet)` → `false` (tests the type assertion `!ok` branch).

mysql/mariadb_gtid.go
triplefighter7619: Added: `{"1-1-1", "2-1-1", false}` — same length sets but different domain IDs, tests the `s.Sets[domainID]` not-found branch.

mysql/mariadb_gtid.go
triplefighter7619: Added: `mariaSet.Contain(mysqlSet)` → `false` (type assertion branch), and `{"1-1-1", "2-1-1", false}` (domain not found branch).

Issue #937, opened 2024-10-28
CI: Add revive


dveeden:
Using [revive](https://github.com/mgechev/revive) as linter might be good. However this would need a reasonable config and probably fixing of some issues like missing comments on exported functions.

@lance6716 @atercattus what do you think?

lance6716:
Do you mean turn on revive in current CI framework

https://github.com/go-mysql-org/go-mysql/blob/5acb5690fa62d58bdfd470336ec7df7c88990dee/.github/workflows/ci.yml#L84-L93

https://golangci-lint.run/usage/linters/#revive

LGTM then.

dveeden:
We should still do this and also add `modernize` to golangci-lint

PR #1125, opened 2026-04-01
mysql: Parsing MySQL GTIDs with tags (second alternative implementation)

### LINKED ISSUE #845: MySQL 8.3 will introduce new GTID format

dveeden:
Closes #845 

Similar to #1116 #1120

This changes the `MysqlGTIDSet` more drastically and gets rid of the `UUIDSet` type.

Function changes:
```diff
-func decodeSid(data []byte) (format GtidFormat, sidnr uint64) {
+func DecodeSid(data []byte) (format GtidFormat, sidnr uint64) {
+func encodeSid(format GtidFormat, sidnr uint64) []byte {
+func (f GtidFormat) String() string {
+func FuzzDecodeSid(f *testing.F) {
+func FuzzParseInterval(f *testing.F) {
-func (gtid *MysqlGTIDSet) Clone() GTIDSet {
+func (i Interval) String() (s string) {
-func (i Interval) String() string {
+func NewMysqlGTIDSet() MysqlGTIDSet {
-func NewUUIDSet(sid uuid.UUID, in ...Interval) *UUIDSet {
-func ParseUUIDSet(str string) (*UUIDSet, error) {
-func (s IntervalSlice) Compare(o IntervalSlice) int {
-func (s IntervalSlice) Less(i, j int) bool {
-func (s IntervalSlice) Swap(i, j int) {
-func (s *MysqlGTIDSet) Add(addend MysqlGTIDSet) error {
+func (s *MysqlGTIDSet) AddGTIDWithTag(uuid uuid.UUID, tag Tag, gno int64) {
-func (s *MysqlGTIDSet) AddSet(set *UUIDSet) {
+func (s *MysqlGTIDSet) Clone() GTIDSet {
-func (s *MysqlGTIDSet) MinusSet(set *UUIDSet) {
-func (s *MysqlGTIDSet) Minus(subtrahend MysqlGTIDSet) error {
-func (s *UUIDSet) AddInterval(in IntervalSlice) {
-func (s *UUIDSet) Bytes() []byte {
-func (s *UUIDSet) Clone() *UUIDSet {
-func (s *UUIDSet) Contain(sub *UUIDSet) bool {
-func (s *UUIDSet) Decode(data []byte) error {
-func (s *UUIDSet) decode(data []byte) (int, error) {
-func (s *UUIDSet) Encode() []byte {
-func (s *UUIDSet) encode(w io.Writer) {
-func (s *UUIDSet) MinusInterval(in IntervalSlice) {
-func (s *UUIDSet) String() string {
+func (t Tag) MarshalBinary() []byte {
```

Based on `git show -U0 | grep -E '^.func ' | sort -k2,3 | grep -v 'func Test'` with some functions that didn't actually change removed.


mysql/mysql_gtid.go
```diff
@@ -4,582 +4,344 @@ import (
...
-	}
-	for sid, uuidSet := range gtid.Sets {
-		clone.Sets[sid] = uuidSet.Clone()
+func encodeSid(format GtidFormat, sidnr uint64) []byte {
+	sid := make([]byte, 8)
+	if format == GtidFormatClassic {
+		_, _ = binary.Encode(sid, binary.LittleEndian, sidnr)
+		return sid
 	}
+	_, _ = binary.Encode(sid, binary.LittleEndian, sidnr<<8)
```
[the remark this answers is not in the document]
dveeden: https://pkg.go.dev/encoding/binary#Encode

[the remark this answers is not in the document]
dveeden: @gemini-code-assist Please check https://pkg.go.dev/encoding/binary#Encode

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,344 @@ import (
...
-	if ok {
-		o.AddInterval(set.Intervals)
-	} else {
-		s.Sets[sid] = set
+func (gtid MysqlGTIDSet) Clone() GTIDSet {
+	g := NewMysqlGTIDSet()
+	for k, v := range gtid {
+		g[k] = v
 	}
+	return g
```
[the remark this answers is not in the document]
dveeden: Please recheck

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
-func (gtid *MysqlGTIDSet) Clone() GTIDSet {
-	clone := &MysqlGTIDSet{
-		Sets: make(map[string]*UUIDSet),
-	}
-	for sid, uuidSet := range gtid.Sets {
-		clone.Sets[sid] = uuidSet.Clone()
+func encodeSid(format GtidFormat, sidnr uint64) []byte {
+	sid := make([]byte, 8)
+	if format == GtidFormatClassic {
+		_, _ = binary.Encode(sid, binary.LittleEndian, sidnr)
```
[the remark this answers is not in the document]
dveeden: /gemini Read https://pkg.go.dev/encoding/binary#Encode

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
+		return sid
 	}
+	_, _ = binary.Encode(sid, binary.LittleEndian, sidnr<<8)
```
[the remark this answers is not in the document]
dveeden: /gemini Read https://pkg.go.dev/encoding/binary#Encode

mysql/mysql_gtid.go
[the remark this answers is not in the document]
dveeden: @gemini-code-assist please check https://pkg.go.dev/encoding/binary#Encode

mysql/mysql_gtid.go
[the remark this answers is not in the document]
dveeden: @gemini-code-assist please check https://pkg.go.dev/encoding/binary#Encode

dveeden:
@lance6716 could you give this a review?

lance6716:
> @lance6716 could you give this a review?

I guess I have more time tomorrow 😄

mysql/testdata/fuzz/FuzzDecodeMysqlGTIDSet/8cfae21719bb90af
```diff
@@ -0,0 +1,2 @@
+go test fuzz v1
```
lance6716, line 1: Do we want to check in fuzz corpus files into the repo? If not, consider keeping them local / adding an ignore entry; if yes, maybe add a short rationale since these files tend to grow over time.

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
+// - Lowercasing the tag
+var tagRegexp = regexp.MustCompile(`^[a-z_][a-z0-9_]{0,31}$`)
 
-type IntervalSlice []Interval
+// Tag is a GTID Tag
+type Tag string
 
-func (s IntervalSlice) Len() int {
-	return len(s)
+func (t Tag) MarshalBinary() []byte {
```
lance6716: Minor API nit: `Tag.MarshalBinary()` looks like an `encoding.BinaryMarshaler` implementation, but the signature does not match (`MarshalBinary() ([]byte, error)`).
Consider renaming it (e.g. `encodeBinary`) or implementing the full interface.
Also, since `Tag` is a string type, it can be constructed without `tagRegexp`; a small guard against `uint8(len(tag)<<1)` overflow would make this more robust.

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
-		return nil, errors.Trace(err)
-	}
-
-	// Handle interval
-	for i := 1; i < len(sep); i++ {
-		if in, err := parseInterval(sep[i]); err != nil {
-			return nil, errors.Trace(err)
-		} else {
-			s.Intervals = append(s.Intervals, in)
+		intervalCount := int(binary.LittleEndian.Uint64(data[pos : pos+8]))
```
lance6716: Robustness: `intervalCount := int(binary.LittleEndian.Uint64(...))` converts before bounds checking.
If the uint64 value is > MaxInt it can wrap to a small/negative `int`, which may bypass the `== 0` / `> math.MaxInt/16` checks and later panic on slicing (or silently store nil intervals).
Safer pattern: keep it as `uint64`, validate `>0` and `intervalCount <= uint64((len(data)-pos)/16)` (and that `pos + intervalCount*16` doesn't overflow), then convert to `int`.
Also: on trailing bytes you currently return a partially decoded set + error; returning `nil, err` would make error handling more consistent.

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
+				_ = binary.Write(&buf, binary.LittleEndian, interval.Stop)
+			}
 		}
 	}
 
-	return true
+	return buf.Bytes()
 }
 
 func (s *MysqlGTIDSet) Equal(o GTIDSet) bool {
```
lance6716, line 287: Semantics: `reflect.DeepEqual` makes `Equal` depend on representation details (e.g. nil vs empty maps/slices). That can be surprising across different constructors/decoders.
I'd rather implement semantic equality: type-assert to `*MysqlGTIDSet`, then compare UUID/tag keys and compare normalized interval slices.

mysql/mysql_gtid.go
```diff
@@ -4,582 +4,387 @@ import (
...
-	}
+func (s *MysqlGTIDSet) IsEmpty() bool {
+	return len(*s) == 0
+}
 
-	for key, set := range sub.Sets {
-		o, ok := s.Sets[key]
-		if !ok {
-			return false
+func (s *MysqlGTIDSet) String() string {
```
lance6716, line 312: Perf: `String()` uses repeated `fmt.Sprintf` and `s += ...` in nested loops, which allocates a lot for large sets.
Using `strings.Builder` (plus `strconv.AppendInt` / writing ints without fmt) would cut allocations significantly.

mysql/mysql_interval.go
```diff
@@ -0,0 +1,154 @@
...
+			stop := max(last.Stop, s[i].Stop)
+			n[len(n)-1] = Interval{last.Start, stop}
+		}
+	}
+
+	return n
+}
+
+// InsertInterval is merging an Interval into an IntervalSlice
+func (s *IntervalSlice) InsertInterval(interval Interval) {
```
lance6716: Perf: `InsertInterval` appends then calls `Normalize()` (sort+merge) on every insertion.
If this is used in hot paths (e.g. binlog GTID tracking), consider an incremental merge / fast path to avoid full sort each time.

replication/event.go
```diff
@@ -268,41 +268,10 @@ type PreviousGTIDsEvent struct {
...
-		masked := make([]byte, 8)
-		copy(masked, data[1:7])
-		sidnr = binary.LittleEndian.Uint64(masked)
-		return format, sidnr
-	}
-	sidnr = binary.LittleEndian.Uint64(data[:8])
-	return format, sidnr
-}
-
 func (e *PreviousGTIDsEvent) Decode(data []byte) error {
```
lance6716, line 272: Safety: `PreviousGTIDsEvent.Decode` doesn't check `len(data) >= 8`; since `mysql.DecodeSid` returns (Classic, 0) on short input, truncated events can be treated as empty sets.
Also `sliceCount := binary.LittleEndian.Uint16(data[pos : pos+8])` seems inconsistent with other GTID encodings that use an 8-byte interval count; at least it truncates >65535.
I'd consider validating buffer lengths before slicing, parsing the interval count as `uint64`, and adding bounds checks to avoid panics on malformed input.

lance6716 (commented):
Thanks for the PR — I left inline comments on a few robustness/perf items.

Two additional issues I couldn't place as inline comments because the relevant lines/files aren't touched in this PR:

1) Tagged GTID events don't currently advance the BinlogSyncer's GTID position.
   `replication/parser.go` can emit `*GtidTaggedLogEvent` (GTID_TAGGED_LOG_EVENT), but `BinlogSyncer.handleEventAndACK` only handles `*GTIDEvent`.
   That means `currGset/prevGset` won't be updated for tagged transactions. Suggest adding a `case *GtidTaggedLogEvent:` and updating via
   `AddGTIDWithTag(uuid, NormalizeTag(event.Tag), event.GNO)`.

2) `(*GTIDEvent).GTIDNext()` ignores `Tag`. For tagged events, returning `uuid:tag:gno` (instead of `uuid:gno`) would preserve tag information.

(I did not run tests locally; relying on CI.)

---

- LLM model: gpt-5.2
- Thinking level: xhigh
- Client: Codex Desktop 26.325.31654 (build 1272); codex-cli 0.117.0

mysql/testdata/fuzz/FuzzDecodeMysqlGTIDSet/8cfae21719bb90af
dveeden: These files are only the fuzz corpus files that are failing inputs. So they should be small.

This then serves as regression test and is picked up by `go test`.

```
$ go test -v -run FuzzDecodeMysqlGTIDSet/8cfae21719bb90af
=== RUN   FuzzDecodeMysqlGTIDSet
=== RUN   FuzzDecodeMysqlGTIDSet/8cfae21719bb90af
--- PASS: FuzzDecodeMysqlGTIDSet (0.00s)
    --- PASS: FuzzDecodeMysqlGTIDSet/8cfae21719bb90af (0.00s)
PASS
ok  	github.com/go-mysql-org/go-mysql/mysql	0.002s
```

However I'm also not really sure if we should include this or not.

lance6716: I'll do some search about the best practice of go fuzz files soon

lance6716: Google says we should track this kind of file

dveeden:
@lance6716 Could you give this another review?

dump/parser.go
```diff
@@ -59,6 +59,7 @@ func Parse(r io.Reader, h ParseHandler, parseBinlogPos bool) error {
 			if m := gtidExp.FindAllStringSubmatch(line, -1); len(m) == 1 {
 				gtidStr := m[0][1]
 				if gtidStr != "" {
+					fmt.Printf("GTID STR: %s\n", gtidStr)
```
lance6716: Is it just your debug log?

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,417 @@ package mysql
...
+// And each tag has one or more Intervals.
+type MysqlGTIDSet map[uuid.UUID]map[Tag]IntervalSlice
 
-	s.Sort()
+// This ensures that MysqlGTIDSet implements the GTIDSet interface
+var _ GTIDSet = &MysqlGTIDSet{}
 
-	n = append(n, s[0])
+func DecodeMysqlGTIDSet(data []byte) (*MysqlGTIDSet, error) {
+	s := NewMysqlGTIDSet()
```
lance6716: better to move after `len(data) < 8` check to avoid creating that object in error path

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,417 @@ package mysql
...
 }
 
-func (s *MysqlGTIDSet) Encode() []byte {
-	var buf bytes.Buffer
-
-	_ = binary.Write(&buf, binary.LittleEndian, uint64(len(s.Sets)))
-
-	for i := range s.Sets {
-		s.Sets[i].encode(&buf)
+// Decode the number of sids (source identifiers) and if it is using
```
lance6716:
```suggestion
// DecodeSid the number of sids (source identifiers) and if it is using
```

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,417 @@ package mysql
...
+// tagged GTIDs or classic (non-tagged) GTIDs.
+//
+// Note that each gtid tag increases the sidno here, so a single UUID
```
lance6716: the second return value is `sidno` or `sidnr`?

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,417 @@ package mysql
...
-		}
+			taglen := int(data[pos] >> 1)
+			pos++
 
-		if sub[i].Start < s[j].Start || sub[i].Stop > s[j].Stop {
-			return false
+			if pos+taglen > len(data) {
+				return nil, errors.New("invalid gtid set buffer, tag extends beyond data")
+			}
+			tag = Tag(data[pos : pos+taglen])
```
lance6716, line 89: Do you think we should always use `NormalizeTag` (or rename it to `NewTag` to create it? Don't use raw type conversions.

lance6716 (commented):
(reviewing)

dump/parser.go
dveeden: Yes. I've removed it now.

mysql/mysql_gtid.go
dveeden: I think renaming it to `NewTag` isn't efficient as then we might normalize already normalized tags. 

So you want a `NewTag()` instead of `Tag()`?

lance6716: `NewTag(string)`

I haven't finished reading all codes. My idea is just normalize all `Tag` when creating it, so `NewTag` is the prefered and only way to do this.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
-				i++
-			} else if minuend.Start < subtrahend.Start && minuend.Stop > subtrahend.Stop {
-				n = append(n, Interval{minuend.Start, subtrahend.Start})
-				minuend = Interval{subtrahend.Stop, minuend.Stop}
-				j++
-			} else {
-				panic("should never be here")
+			s[sid] = map[Tag]IntervalSlice{
+				tag: intervals.Normalize(),
 			}
```
lance6716, line 121: maybe run `func (s IntervalSlice) Normalize()` once after all is set, to reduce repeated computing.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
+		if _, ok := s[u]; !ok {
+			s[u] = make(map[Tag]IntervalSlice, 1)
+		}
+		// Handle interval(s) and tags
+		tag := Tag("")
+		for i := 1; i < len(sep); i++ {
+			if tagRegexp.MatchString(sep[i]) {
+				tag = NormalizeTag(sep[i])
+				if _, ok := s[u][tag]; !ok {
+					s[u][tag] = nil
```
lance6716, line 171: no need to do this? the zero value of uninitialized map entry is nil already

lance6716 (commented):
(will reviewing hours later)

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
 }
 
-func (s *MysqlGTIDSet) MinusSet(set *UUIDSet) {
-	if set == nil {
-		return
-	}
-	sid := set.SID.String()
-	uuidSet, ok := s.Sets[sid]
+func (s *MysqlGTIDSet) AddGTIDWithTag(uuid uuid.UUID, tag Tag, gno int64) {
+	_, ok := (*s)[uuid]
```
lance6716, line 196: Oh `(*s)[uuid]` this is a little awkward. Maybe directly use 

```
func (s MysqlGTIDSet) AddGTIDWithTag
```

and also let `DecodeMysqlGTIDSet` returns `MysqlGTIDSet`.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
 	if ok {
-		uuidSet.MinusInterval(set.Intervals)
-		if uuidSet.Intervals == nil {
-			delete(s.Sets, sid)
+		(*s)[uuid][tag] = append((*s)[uuid][tag], Interval{gno, gno + 1}).Normalize()
```
lance6716, line 198: I'm considering for these small changes to intervals, we can have a more specialized logic to normalized. Maybe add a TODO and we will do it later.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
-	}
-	for _, uuidSet := range gtidSet.(*MysqlGTIDSet).Sets {
-		s.AddSet(uuidSet)
+func (s *MysqlGTIDSet) Clone() GTIDSet {
+	g := NewMysqlGTIDSet()
+	for k, v := range *s {
+		newInnerMap := make(map[Tag]IntervalSlice, len(v))
+		for k2, v2 := range v {
+			newSlice := make(IntervalSlice, len(v2))
+			copy(newSlice, v2)
```
lance6716: we can use slices.Clone now

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
-	if ok {
-		o.Intervals.InsertInterval(Interval{gno, gno + 1})
-	} else {
-		s.Sets[sid] = &UUIDSet{uuid, IntervalSlice{Interval{gno, gno + 1}}}
+func (s *MysqlGTIDSet) Contain(o GTIDSet) bool {
+	if om, ok := o.(*MysqlGTIDSet); ok {
+		for k := range *om {
+			if _, ok := (*s)[k]; !ok {
+				return false
+			} else {
```
lance6716: My IDE says because the if branch has `return`, it's no need to write `else` level

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
-	for key, set := range sub.Sets {
-		o, ok := s.Sets[key]
-		if !ok {
-			return false
+	slices.SortFunc(uuids, func(a, b uuid.UUID) int {
+		return bytes.Compare(a[:], b[:])
+	})
+	for _, uuid := range uuids {
+		var tags []Tag
+		for tag := range (*s)[uuid] {
```
lance6716: `slices.Collect(maps.Keys(...))`

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,415 @@ package mysql
...
+		slices.Sort(tags)
 
-		if !o.Contain(set) {
-			return false
+		for _, tag := range tags {
+			ubin, _ := uuid.MarshalBinary()
+			buf.Write(ubin)
+
+			if format == GtidFormatTagged {
+				tbin, _ := tag.MarshalBinary()
```
lance6716, line 272: Because we ignored the error from `tag.MarshalBinary()`, I think we should also check the length in `NewTag` and afterwards assume all Tags have legal length

lance6716 (commented):
large PR 😂 will review later

mysql/mysql_gtid.go
dveeden: Yes this is needed:

```go
> [Breakpoint 1] github.com/go-mysql-org/go-mysql/mysql.ParseMysqlGTIDSet-range1() ./mysql_gtid.go:161 (hits goroutine(23):1 total:1) (PC: 0x76e882)
   156:			tag := Tag("")
   157:			for i := 1; i < len(sep); i++ {
   158:				if tagRegexp.MatchString(sep[i]) {
   159:					tag = NormalizeTag(sep[i])
   160:					if _, ok := s[u][tag]; !ok {
=> 161:						s[u][tag] = nil
   162:					}
   163:				} else {
   164:					if in, err := parseInterval(sep[i]); err != nil {
   165:						return nil, errors.Trace(err)
   166:					} else {
(dlv) p u
github.com/google/uuid.UUID [222,39,138,208,33,6,17,228,159,142,110,221,12,162,9,71]
(dlv) p s[u]
map[github.com/go-mysql-org/go-mysql/mysql.Tag]github.com/go-mysql-org/go-mysql/mysql.IntervalSlice [
	"": [
		(*"github.com/go-mysql-org/go-mysql/mysql.Interval")(0xc000125610),
	], 
]
(dlv) p tag
"foo"
(dlv) n
> github.com/go-mysql-org/go-mysql/mysql.ParseMysqlGTIDSet-range1() ./mysql_gtid.go:157 (PC: 0x76e8f8)
   152:			if _, ok := s[u]; !ok {
   153:				s[u] = make(map[Tag]IntervalSlice, 1)
   154:			}
   155:			// Handle interval(s) and tags
   156:			tag := Tag("")
=> 157:			for i := 1; i < len(sep); i++ {
   158:				if tagRegexp.MatchString(sep[i]) {
   159:					tag = NormalizeTag(sep[i])
   160:					if _, ok := s[u][tag]; !ok {
   161:						s[u][tag] = nil
   162:					}
(dlv) p s[u]
map[github.com/go-mysql-org/go-mysql/mysql.Tag]github.com/go-mysql-org/go-mysql/mysql.IntervalSlice [
	"": [
		(*"github.com/go-mysql-org/go-mysql/mysql.Interval")(0xc000125610),
	], 
	"foo": nil, 
]
```

mysql/mysql_gtid.go
dveeden: So this would be an optimization for something like `<uuid>:1-10:12-20:22-30:...`, where there are multiple intervals for the same uuid+tag.

dveeden: When I try this it doesn't seem to call `Normalize()` more than needed. But maybe my test is wrong..

mysql/mysql_gtid.go
dveeden: But there is nothing that prevents people from directly calling `Tag()`, right?

dveeden: We could have something like this:
```go
type Tag struct {
  tag string
  normalized bool
}
```

And then we would have to only use the normalization once. But I don't like the complexity of that.

mysql/mysql_gtid.go
dveeden: I think we should keep `AddGTIDWithTag()` and `AddGTID() as similar as possible. There are already quite a few API changes and I'd like to avoid adding more of them.

I do agree that `(*s)[uuid]` feels a bit weird. But is it really a problem?

mysql/mysql_gtid.go
dveeden: @lance6716 I pushed a commit that renames `NormalizeTag()` to `NewTag()`. I change most `Tag(...)` to `NewTag(...)`. Could you check if this is ok?

lance6716: > But there is nothing that prevents people from directly calling Tag(), right?

Yes. 😂 Only through comment.

mysql/mysql_gtid.go
lance6716: I'll keep this comment open and check it later

mysql/mysql_gtid.go
lance6716: Not a big problem. I'll resolve this comment and maybe create similar comments wiht more detailed suggestions later

mysql/mysql_interval.go
```diff
@@ -0,0 +1,148 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/pingcap/errors"
+)
+
+// Like MySQL GTID Interval struct, [start, stop), left closed and right open
+// See MySQL rpl_gtid.h
+type Interval struct {
+	// The first GID of this interval.
```
lance6716: What's GID? I'm not familiar with it, do you mean GNO or something? Maybe add comments

mysql/mysql_interval.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Interval is [start, stop), but the GTID string's format is [n] or [n1-n2], closed interval
+func parseInterval(str string) (i Interval, err error) {
+	p := strings.Split(str, "-")
+
+	switch len(p) {
+	case 1:
+		i.Start, err = strconv.ParseInt(p[0], 10, 64)
+		if err != nil {
+			return i, errors.Errorf("invalid interval format, not numeric: %v", err)
```
lance6716: please check the raw input `str` is also displayed in error message, to help debug.

mysql/mysql_interval.go
```diff
@@ -0,0 +1,148 @@
...
+	return i, err
+}
+
+func (i Interval) String() (s string) {
+	if i.Stop == i.Start+1 {
+		s += fmt.Sprintf("%d", i.Start)
+	} else {
+		s += fmt.Sprintf("%d-%d", i.Start, i.Stop-1)
+	}
+	return s
```
lance6716:
```suggestion
	if i.Stop == i.Start+1 {
		return fmt.Sprintf("%d", i.Start)
	} else {
		return fmt.Sprintf("%d-%d", i.Start, i.Stop-1)
	}
```

mysql/mysql_interval.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (s IntervalSlice) Normalize() IntervalSlice {
+	if s == nil {
+		return nil
+	}
+
+	n := IntervalSlice{}
+	if len(s) == 0 {
+		return n
```
lance6716:
```suggestion
		return s
```

and

1. move creating `n` after this branch
2. preserve the capacity of `n` to the length of `s`

Or maybe modify `s` in-place, to avoid create the new object `n` 🤔

mysql/mysql_interval.go
```diff
@@ -0,0 +1,148 @@
...
+			stop := max(last.Stop, s[i].Stop)
+			n[len(n)-1] = Interval{last.Start, stop}
+		}
+	}
+
+	return n
+}
+
+// Contain returns true if sub in s
+func (s IntervalSlice) Contain(sub IntervalSlice) bool {
```
lance6716, line 113: Should `s` and `sub` be Sorted and Normalized? I think this is like the properties of Tag. Maybe we can make the public type has a `inner` private member, to maintain the properties of `inner` member

```go
type IntervalSlice struct {
  inner []Interval
}
```

mysql/mysql_gtid.go
dveeden: I could create a `^[a-z_][a-z0-9_]{0,31}$` regexp. And in `NewTag()` return an error if it doesn't match.  Or I could use the existing regexp to match the input before it is normalized.

However that means that the signature for `NewTag()` needs to change.

Another option is to check the error that `MarshalBinary()` returns, but then either need to make `Encode()` return an error or have it return a `[]byte` of zero lenght.

mysql/mysql_gtid.go
dveeden: Could you add a suggestion for the TODO line?

lance6716: Sure — here's a TODO suggestion that captures the intent without changing behavior:

```suggestion
		// TODO: Avoid full Normalize() on each AddGTIDWithTag; we can merge the single appended interval into an already-normalized slice.
```

mysql/mysql_interval.go
lance6716: Not blocking from my side. The current `Normalize()` looks fine for correctness; we can revisit prealloc / in-place normalization if perf profiling shows it matters.

mysql/mysql_gtid.go
lance6716: I'd avoid silently ignoring `tag.MarshalBinary()` errors since `Encode()` can't propagate them. Two pragmatic options:

1) Make the error impossible in normal flows by validating at the boundaries (e.g. in `DecodeMysqlGTIDSet`: reject `taglen > 32` + invalid normalized tag bytes; `ParseMysqlGTIDSet` already validates; for `AddGTIDWithTag` either document preconditions or add a cheap length check).

2) If an invalid `Tag` still sneaks in (external user constructed `Tag("...")`), handle it explicitly in `Encode()` (panic with a clear message, or return `nil` to fail closed) rather than writing a malformed buffer.

Changing `NewTag()` to return `(Tag, error)` would be stricter but is a larger API change.

mysql/mysql_interval.go
lance6716: Yep — `Contain`/`Equal` currently assume both sides are already sorted + normalized (non-overlapping). Internally we seem to enforce that (Parse/Decode/Add/Update normalize), so correctness is fine.

For external callers, I'd prefer the simple route: document the precondition on `IntervalSlice.Contain`/`Equal` (and maybe add a small helper/constructor if needed) rather than wrapping into a struct with private `inner`, which is a bigger API change.

lance6716 (changes requested):
Resolved the addressed review threads and re-reviewed the current head.

**Needs changes**
1) `replication/binlogsyncer.go`: in the `*GtidTaggedLogEvent` branch, `prevGset` is updated using the *current* `event.Tag` together with `b.prevMySQLGTIDEvent.GNO`. This should use `b.prevMySQLGTIDEvent.Tag` (tags can change between transactions). Also the `*GTIDEvent` branch should avoid losing the previous tag if the previous event was tagged — simplest is to consistently use `AddGTIDWithTag(u, <event>.Tag, <event>.GNO)` for both branches.

2) `replication/event.go` (`PreviousGTIDsEvent.Decode`):
   - Switching `sliceCount` to `uint64` is correct, but the new `sliceCount > math.MaxUint16` rejection is arbitrary and may reject valid inputs. Better to validate based on remaining buffer size (and overflow) similarly to `mysql.DecodeMysqlGTIDSet`.
   - The tagged string-building logic currently depends on `tag == ""` to emit the UUID/comma. If the first entry for a UUID has a non-empty tag (no empty-tag SID), this can produce an invalid string (starts with `:tag`). Safer approach: track `lastUUID` and group by UUID changes.
   - Also consider adding bounds checks for `tagLength` and interval decoding to avoid panics on truncated/corrupted events.

3) `mysql/mysql_gtid.go`: `Encode()` ignores the `tag.MarshalBinary()` error. Since `Encode()` can’t return an error, please either (a) validate at the boundaries so this cannot happen in normal flows (e.g. reject `taglen > 32` / invalid normalized tag bytes in `DecodeMysqlGTIDSet`, document/guard `AddGTIDWithTag`), or (b) fail closed (panic / return `nil`) instead of silently emitting a malformed buffer.

**Nice to have / follow-ups**
- Performance: `AddGTIDWithTag` currently does `append(...).Normalize()` on every add; old code used an insertion/merge path. If this is on a hot path, consider reintroducing a small “insert one interval into a normalized slice” helper. (I left a TODO suggestion inline.)
- API compatibility: this PR removes the exported `UUIDSet` / `ParseUUIDSet` / `NewUUIDSet` surface; if we want to avoid breaking users, consider keeping wrappers/deprecation.

**Local test signal**
- `go test ./mysql ./replication` passes.
- Full `go test ./...` fails locally because some suites expect a MySQL instance on `127.0.0.1:3306` (unrelated to this PR).

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,425 @@ package mysql
...
-	if err != nil {
-		return i, err
-	}
+// Normalized tags should match: `^[a-z_][a-z0-9_]{0,31}$`
 
-	if i.Stop <= i.Start {
-		err = errors.Errorf("invalid interval format, must n[-n] and the end must >= start")
-	}
+// Tag is a GTID Tag
+type Tag string
```
lance6716: If we use 

```
type Tag struct {
 normalized string
}
```

The inner field can't be accessed by user, we can do the one time normalization when create `Tag`.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,425 @@ package mysql
...
-func (i Interval) String() string {
-	if i.Stop == i.Start+1 {
-		return fmt.Sprintf("%d", i.Start)
-	} else {
-		return fmt.Sprintf("%d-%d", i.Start, i.Stop-1)
+func (t Tag) MarshalBinary() ([]byte, error) {
+	if len(t) > 32 {
+		return nil, errors.New("tag length too long")
 	}
+	taglen := uint8(len(t) << 1)
```
lance6716:
```suggestion
	tagLen := uint8(len(t) << 1)
```

golang's style

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,425 @@ package mysql
...
-	for i := 1; i < len(s); i++ {
-		last := n[len(n)-1]
-		if s[i].Start > last.Stop {
-			n = append(n, s[i])
-			continue
-		} else {
-			stop := max(last.Stop, s[i].Stop)
-			n[len(n)-1] = Interval{last.Start, stop}
-		}
+func DecodeMysqlGTIDSet(data []byte) (*MysqlGTIDSet, error) {
```
lance6716, line 70: move this function after `NewMysqlGTIDSet`? I think it's better to see `NewXXX` after defining `XXX`

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,425 @@ package mysql
...
+	if len(data) < 8 {
```
lance6716: there's another check len < 8 inside `DecodeSid`. And I have checked all call site of `DecodeSid` has already checked len < 8 🤔 Maybe we can delete the len < 8 branch inside `DecodeSid`

lance6716 (commented):
some human review

mysql/mysql_gtid.go
dveeden: We can't remove it from `DecodeSid()`:
- This is exported, so no guarantee it is only called from `DecodeMysqlGTIDSet()`
- We can't unexport it easily as it is also used from `replication/`.

We can't remove the check from `DecodeMysqlGTIDSet()`:
- While `DecodeSid()` checks for the length, the signature doesn't allow it to return an error.

mysql/mysql_interval.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		i.Stop = i.Start + 1
+	case 2:
+		i.Start, err = strconv.ParseInt(p[0], 10, 64)
+		if err == nil {
+			i.Stop, err = strconv.ParseInt(p[1], 10, 64)
+			i.Stop++
+		}
+	default:
+		err = errors.New("invalid interval format, must n[-n]")
```
lance6716, line 40: this `n[-n]` might mean the start and end is the same. Maybe `must n or m-n`, `n1-n2` etc.

mysql/mysql_interval.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if err != nil {
+		return i, err
+	}
+
+	if i.Stop <= i.Start {
+		err = errors.New("invalid interval format, must n[-n] and the end must >= start")
```
lance6716, line 48: ditto

mysql/mysql_interval.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (i Interval) String() (s string) {
+	if i.Stop == i.Start+1 {
+		return fmt.Sprintf("%d", i.Start)
+	} else {
+		return fmt.Sprintf("%d-%d", i.Start, i.Stop-1)
+	}
+}
+
+type IntervalSlice []Interval
```
lance6716, line 62: we can also use

```go
type IntervalSlice struct {
  normalized []Interval
}
```

to do one-time normalization at init time. And `Sort` and `Normalize` may have no need to be public.

mysql/mysql_interval.go
dveeden: This is what was already in master:

https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/mysql/mysql_gtid.go#L41

I think `n` here means number (regexp: `[0-9]+`) and not a variable.

Changing this might be good to do, but maybe outside of this PR is best.

mysql/mysql_interval.go
dveeden: This message is also in master: https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/mysql/mysql_gtid.go#L49

mysql/mysql_interval.go
dveeden: `IntervalSlice.Sort` and `IntervalSlice.Normalize` were already public:

https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/mysql/mysql_gtid.go#L83

https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/mysql/mysql_gtid.go#L87

This PR is already making a lot of API changes, I rather not add more to it.

And the `IntervalSlice` type didn't change in this PR, it was already like this in `master`.

https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/mysql/mysql_gtid.go#L63

This could be a good set of changes that we could consider after this PR is merged.

mysql/mysql_gtid.go
lance6716: I see in old codes it's not exported. So we have a chance to change the signature of `DecodeSid` to add the returned error type. And remove the length check at the two caller.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
-				break
+		if format == GtidFormatTagged {
+			if pos >= len(data) {
+				return nil, errors.New("invalid gtid set buffer, tag length expected")
 			}
-		}
-		if j == len(s) {
-			return false
-		}
+			taglen := int(data[pos] >> 1)
```
lance6716: nit:

```suggestion
			tagLen := int(data[pos] >> 1)
```

(also rename other references)

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
+		}
 
-			s.AddSet(set)
+		if _, ok := s[u]; !ok {
+			s[u] = make(map[Tag]IntervalSlice, 1)
+		}
+		// Handle interval(s) and tags
+		tag := NewTag("")
+		for i := 1; i < len(sep); i++ {
+			if tagRegexp.MatchString(sep[i]) {
```
lance6716, line 175: Does MySQL support that a tag appears more than once?

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
-	}
-	sid := set.SID.String()
-	o, ok := s.Sets[sid]
-	if ok {
-		o.AddInterval(set.Intervals)
-	} else {
-		s.Sets[sid] = set
-	}
+func (s *MysqlGTIDSet) AddGTID(uuid uuid.UUID, gno int64) {
+	s.AddGTIDWithTag(uuid, Tag{}, gno)
```
lance6716, line 199: nit: in line 173 we have used `NewTag("")`, their coding style can be unified

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
 
-func (s *MysqlGTIDSet) Update(GTIDStr string) error {
-	gtidSet, err := ParseMysqlGTIDSet(GTIDStr)
-	if err != nil {
-		return err
-	}
-	for _, uuidSet := range gtidSet.(*MysqlGTIDSet).Sets {
-		s.AddSet(uuidSet)
+func (s *MysqlGTIDSet) Clone() GTIDSet {
+	g := NewMysqlGTIDSet()
```
lance6716: we can also reserve the capacity for `g` to `len(*s)`

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
 
-func (s *MysqlGTIDSet) AddGTID(uuid uuid.UUID, gno int64) {
-	sid := uuid.String()
-	o, ok := s.Sets[sid]
-	if ok {
-		o.Intervals.InsertInterval(Interval{gno, gno + 1})
-	} else {
-		s.Sets[sid] = &UUIDSet{uuid, IntervalSlice{Interval{gno, gno + 1}}}
+func (s *MysqlGTIDSet) Contain(o GTIDSet) bool {
+	if om, ok := o.(*MysqlGTIDSet); ok {
```
lance6716: nit: the ident can be shorter

```
om, ok := o.(*MysqlGTIDSet)
if !ok {
  return false
}
...
```

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
+		for k := range *om {
+			if _, ok := (*s)[k]; !ok {
+				return false
+			}
+			for k2 := range (*om)[k] {
+				if i, ok := (*s)[k][k2]; !ok {
+					return false
+				} else {
```
lance6716: and here. We have returned `false` so no need for `else`. Not sure why linter does not report it.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
-	return nil
-}
+func (s *MysqlGTIDSet) Encode() []byte {
+	var buf bytes.Buffer
 
-func (s *MysqlGTIDSet) Minus(subtrahend MysqlGTIDSet) error {
-	for _, uuidSet := range subtrahend.Sets {
-		s.MinusSet(uuidSet)
+	format := GtidFormatClassic
+	sidcount := uint64(0)
```
lance6716:
```suggestion
	sidCount := uint64(0)
```

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
+	var uuids []uuid.UUID
```
lance6716: reserve the capacity

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
+	for uuid := range *s {
+		uuids = append(uuids, uuid)
+		for tag := range (*s)[uuid] {
+			sidcount++
+			if format != GtidFormatTagged && (tag != Tag{""}) {
```
lance6716: there are 3 styles. NewTag, Tag{}, Tag{""}. I think Tag{} is short and OK.

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
-	}
-
-	if len(sub.Sets) != len(s.Sets) {
-		return false
-	}
-
-	for key, set := range sub.Sets {
-		o, ok := s.Sets[key]
-		if !ok {
+	if om, ok := o.(*MysqlGTIDSet); ok {
```
lance6716: nit: ident is less in the old code

mysql/mysql_gtid.go
```diff
@@ -2,584 +2,434 @@ package mysql
...
+		return err
 	}
-	sort.Strings(sets)
-
-	sep := ""
-	for _, set := range sets {
-		buf.WriteString(sep)
-		buf.WriteString(set)
-		sep = ","
+	if om, ok := o.(*MysqlGTIDSet); ok {
```
lance6716: nit: ident

lance6716 (commented):
Human review :P 

I have finished reviewing mysql_gtid.go. Will take a look at mysql_interval.go tomorrow

mysql/mysql_interval.go
```diff
@@ -0,0 +1,147 @@
...
+			stop := max(last.Stop, s[i].Stop)
+			n[len(n)-1] = Interval{last.Start, stop}
+		}
+	}
+
+	return n
+}
+
+// Contain returns true if sub in s
+func (s IntervalSlice) Contain(sub IntervalSlice) bool {
```
lance6716, line 115: maybe `sub` is not Sorted

mysql/mysql_interval.go
```diff
@@ -0,0 +1,147 @@
...
+	return true
+}
+
+func (s IntervalSlice) Equal(o IntervalSlice) bool {
+	if len(s) != len(o) {
+		return false
+	}
+
+	for i := range s {
+		if s[i].Start != o[i].Start || s[i].Stop != o[i].Stop {
```
lance6716: maybe we can directly compare `Interval` struct

lance6716:
rest lgtm

mysql/mysql_gtid.go
dveeden: I assume you mean a UUID+Tag combination? That should be unique. The same tag can appear multiple times for different UUID's.

Let me see what MySQL does when the UUID+Tag combo isn't unique.

dveeden: It gets accepted and is then normalized to a format that has unique tag/uuid combinations.

```
mysql-9.7.0 [(none)]> set global gtid_purged='01e1f2be-470e-11f1-9d68-12bf2d2c403a:foo:1-10:bar:10-20:foo:8-15';
Query OK, 0 rows affected (0.011 sec)

mysql-9.7.0 [(none)]> SELECT @@global.gtid_purged;
+---------------------------------------------------------+
| @@global.gtid_purged                                    |
+---------------------------------------------------------+
| 01e1f2be-470e-11f1-9d68-12bf2d2c403a:bar:10-20:foo:1-15 |
+---------------------------------------------------------+
1 row in set (0.001 sec)
```

mysql/mysql_gtid.go
dveeden: I think #937 is needed for this, but I think we should do that after this PR has merged.

dveeden:
@lance6716 This should be ready for another review

lance6716 (approved):

PR #1128, opened 2026-04-21
replication: fix BinlogSyncer.Close hang on SSH tunnels

serprex:
SetReadDeadline unblocks ReadPacket so onStream notices ctx cancellation. Some transports (notably SSH tunnels) refuse deadlines with an error, in which case close the underlying connection to force ReadPacket to return. Otherwise wg.Wait below parks forever when KILL also fails to reach server (e.g. thread already reaped, "ERROR 1094 unknown thread id")

see https://github.com/PeerDB-io/peerdb/pull/4195

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -240,9 +240,17 @@ func (b *BinlogSyncer) close() {
 	b.cancel()
 
 	if b.c != nil {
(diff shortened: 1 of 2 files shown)


lance6716 (approved):
Sorry for the late review

Issue #1037, opened 2025-05-13
Security concerns with existing password implementation for the Server


gdsmith:
I was looking to implement a version of the `CredentialProvider` for a MySQL proxy. I was trying to understand how to store the password so I could return what was expected. Looking at how the password is used during authentication it is _required_ that the server gets the password in plain text, therefore it is _required_ for the `CredentialProvider` to be able to provide the password in plain text.

I believe this is a big security issue in the code that can be easily avoided if you correctly follow the MySQL implementations.

As an example, if you look at how this is implemented for the [`mysql_native_password`](https://github.com/mysql/mysql-server/blob/ff05628a530696bc6851ba6540ac250c7a059aa7/libmysql/authentication_native_password/mysql_native_password.cc#L39) in MySQL proper you will see that they rather use the salt/seed to encrypt the `hash_stage1` and then generate the `hash_stage2` from that for comparison with what is stored on the server.

```
  SERVER:  public_seed=generate_user_salt()
           send(public_seed)

  CLIENT:  recv(public_seed)
           hash_stage1=sha1("password")
           hash_stage2=sha1(hash_stage1)
           reply=xor(hash_stage1, sha1(public_seed,hash_stage2)

           // this three steps are done in scramble()

           send(reply)


  SERVER:  recv(reply)
           hash_stage1=xor(reply, sha1(public_seed,hash_stage2))
           candidate_hash2=sha1(hash_stage1)
           check(candidate_hash2==hash_stage2)
```
`
The reason for doing this is so the actual password does not need to be stored anywhere on the server, you just need to store the `hash_stage2` of the password.

lance6716:
Thanks for report this problem. Currently I don't have enough knowledge to discuss about it. It seems your suggestion is correct and improves the security. I'll take a look some days later. Can you provide more details like how we should change the code when I'm learning?

gdsmith:
I would suggest looking at the [implementation in vitess](https://github.com/vitessio/vitess/blob/main/go/mysql/auth_server.go#L299) for an example

dveeden:
@lance6716 I can have a look at this over the next few weeks

gdsmith:
Happy to make a stab at this, but looking for some input on some questions:

1. When looking at the implementation, it seems like the library currently does on the fly password negotiation (because you have access to the underlying password in the store). If we are looking at updating the store to rather keep the hashes, we can either:
[...]

PR #1129, opened 2026-04-27
server: support pre-computed hashed passwords in Credential (closes part of #1037)

takaidohigasi:
## Summary

Adds an opt-in `Credential.HashedPasswords [][]byte` field alongside the existing `Passwords []string`, plus a corresponding helper `InMemoryAuthenticationHandler.AddUserWithHashedPassword(...)`. This lets callers configure user credentials when only the **server-side hash** is available, without ever supplying the plaintext.

This is the next step after #1040 (which removed plaintext-at-rest from the in-memory handler internally but kept the plaintext-only public API). Issue #1037 raised this exact use case and is still open — this PR addresses the API gap it identified.

## Why

Concrete motivation: a MySQL proxy (in our case [`takaidohigasi/mysql-interceptor`](https://github.com/takaidohigasi/mysql-interceptor)) that mirrors its user list from another source — another server's `mysql.user` table or a ProxySQL configuration — where passwords are stored as standard `mysql_native_password` hashes (`*XXXXXXXX...` 41-character form) and the plaintext is intentionally not held anywhere. The proxy needs to authenticate clients using only those hashes.

Today there is no clean way to do this: `Credential.Passwords` is plaintext-only, and even the new pluggable `AuthenticationProvider` (`NewServerWithAuth`) can't help because `Conn.salt` and `Conn.credential` are unexported, so a custom provider outside the `server` package can't run `mysql.CompareNativePassword(...)` itself.

## Behavior

- `compareNativePasswordAuthData` consults `HashedPasswords` first (cheaper per connect — no `SHA1(SHA1(plaintext))` step) and falls back to the existing plaintext path. A credential with **only** `HashedPasswords` set is fully usable.
- A new `Credential.hasAnyCredential()` helper is used by `acquireCredential` so hashed-only users aren't reported as `ER_NO_SUCH_USER`.
- `AddUserWithHashedPassword` validates the auth plugin up front: only `mysql_native_password` is supported in this PR (`caching_sha2_password` and `sha256_password` can be extended later if there's interest). Empty hashes are rejected so misconfigured callers fail loudly instead of silently registering an unauthenticatable user.
- Fully backward-compatible: existing callers using `AddUser(username, password)` are unaffected.

## Hash format

The bytes expected by `HashedPasswords` (for `mysql_native_password`) are the raw 20-byte `SHA1(SHA1(plaintext))` value — the same bytes that the existing `mysql.NativePasswordHash` returns and that `mysql.DecodePasswordHex(\"*XXXX...\")` produces from MySQL's standard hex form. Either helper works:

```go
// From a MySQL/ProxySQL stored hash string:
hash, _ := mysql.DecodePasswordHex("*6BB4837EB74329105EE4568DDA7DC67ED2CA2AD9")
handler.AddUserWithHashedPassword("alice", hash)

// Or directly from a known plaintext (rare in this use case, but for symmetry):
hash := mysql.NativePasswordHash([]byte("s3cr3t"))
handler.AddUserWithHashedPassword("alice", hash)
```

## Tests

- `TestAddUserWithHashedPassword` — end-to-end: registers a hashed user, then a real `database/sql` client (using the project's own driver) connects with the corresponding plaintext (success) and with a different plaintext (denied), all without the handler ever seeing the real password. Also round-trips through `EncodePasswordHex` / `DecodePasswordHex` to confirm the byte format matches.
- `TestAddUserWithHashedPasswordRejectsUnsupportedPlugin` — passing `caching_sha2_password` returns an explicit error.
- `TestAddUserWithHashedPasswordRejectsEmptyHash` — empty / nil hash returns an explicit error.

`go test -race ./server/...` passes; `go vet ./...` and `gofmt -l server/` clean.

## Diff size

3 files changed, 149 insertions(+), 4 deletions(−).

## Out of scope / follow-ups

- `caching_sha2_password` and `sha256_password` hashed-credential support — different stored-hash formats, separate change.
- Server-side rotation / cache invalidation for `caching_sha2_password` (existing concern called out in the package doc) — unchanged here.

Closes part of #1037.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -10,6 +10,8 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/pingcap/tidb/pkg/parser/auth"
+
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
 )
@@ -76,6 +78,46 @@ func scrambleValidation(cached, nonce, scramble []byte) bool {
 	return subtle.ConstantTimeCompare(m, cached) == 1
 }
 
+// safeNativeCompare wraps mysql.CompareNativePassword in a recover so a
+// malformed stored hash (installed by constructing Credential directly,
+// bypassing validateHashedPassword) cannot crash the connection goroutine.
+// Returns (match, err); on panic, err is non-nil and match is false.
+func safeNativeCompare(clientAuthData, hash, salt []byte) (match bool, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			match = false
+			err = errors.Errorf("CompareNativePassword panicked: %v", r)
+		}
+	}()
+	return mysql.CompareNativePassword(clientAuthData, hash, salt), nil
+}
+
+// safeSha256Check wraps mysql.Check256HashingPassword in a recover for the
+// same reason as safeNativeCompare. The verifier slices into parts[2][:16]
+// which can panic on malformed stored hashes.
+func safeSha256Check(hash []byte, password string) (match bool, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			match = false
+			err = errors.Errorf("Check256HashingPassword panicked: %v", r)
+		}
+	}()
+	return mysql.Check256HashingPassword(hash, password)
+}
+
+// safeCachingSha2Check wraps auth.CheckHashingPassword in a recover. The
+// upstream verifier slices into parts[3][:SALT_LENGTH] which can panic on
+// malformed stored hashes.
+func safeCachingSha2Check(hash []byte, password, plugin string) (match bool, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			match = false
+			err = errors.Errorf("CheckHashingPassword panicked: %v", r)
+		}
+	}()
+	return auth.CheckHashingPassword(hash, password, plugin)
+}
+
 func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential Credential) error {
 	if isEmptyPassword(clientAuthData) {
 		if credential.hasEmptyPassword() {
@@ -88,7 +130,12 @@ func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential C
 	// credentials when only the server-side hash is available (no plaintext),
 	// and they're cheaper per connect because we skip the SHA1(SHA1(...)) step.
 	for _, hash := range credential.HashedPasswords {
-		if mysql.CompareNativePassword(clientAuthData, hash, c.salt) {
+		match, err := safeNativeCompare(clientAuthData, hash, c.salt)
+		if err != nil {
+			log.Printf("server: native_password hash compare error for user %q: %v", c.user, err)
+			continue
+		}
+		if match {
 			return nil
 		}
 	}
@@ -145,16 +192,17 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	// (e.g. mirroring `mysql.user.authentication_string`), and we skip
 	// the per-connect hashPassword work.
 	for _, hash := range credential.HashedPasswords {
-		check, err := mysql.Check256HashingPassword(hash, string(clientAuthData))
+		check, err := safeSha256Check(hash, string(clientAuthData))
 		if err != nil {
 			// Stored hashes registered via AddUserWithHashedPassword have
 			// already been shape-checked by validateHashedPassword, so
 			// reaching this branch means either a malformed hash was
-			// installed by constructing Credential directly, or the
-			// upstream verifier changed format expectations. Log so the
-			// silent skip is auditable; auth still falls through to the
-			// plaintext loop and ultimately ErrAccessDenied if nothing
-			// matches.
+			// installed by constructing Credential directly (which can
+			// trigger an upstream slice-out-of-bounds panic that
+			// safeSha256Check converts to an error), or the upstream
+			// verifier changed format expectations. Log so the silent
+			// skip is auditable; auth still falls through to the plaintext
+			// loop and ultimately ErrAccessDenied if nothing matches.
 			log.Printf("server: sha256_password hash compare error for user %q: %v", c.user, err)
 			continue
 		}
--- a/server/auth.go
+++ b/server/auth.go
@@ -129,8 +129,8 @@ func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential C
 	// Pre-computed hashes are checked first: they let callers configure
 	// credentials when only the server-side hash is available (no plaintext),
 	// and they're cheaper per connect because we skip the SHA1(SHA1(...)) step.
-	for _, hash := range credential.HashedPasswords {
-		match, err := safeNativeCompare(clientAuthData, hash, c.salt)
+	for _, hp := range credential.HashedPasswords {
+		match, err := safeNativeCompare(clientAuthData, hp.data, c.salt)
 		if err != nil {
 			log.Printf("server: native_password hash compare error for user %q: %v", c.user, err)
 			continue
@@ -191,8 +191,8 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	// credentials when only the server-side stored hash is available
 	// (e.g. mirroring `mysql.user.authentication_string`), and we skip
 	// the per-connect hashPassword work.
-	for _, hash := range credential.HashedPasswords {
-		check, err := safeSha256Check(hash, string(clientAuthData))
+	for _, hp := range credential.HashedPasswords {
+		check, err := safeSha256Check(hp.data, string(clientAuthData))
 		if err != nil {
 			// Stored hashes registered via AddUserWithHashedPassword have
 			// already been shape-checked by validateHashedPassword, so
--- a/server/auth.go
+++ b/server/auth.go
@@ -41,14 +41,14 @@ func (c *Conn) compareAuthData(authPluginName string, clientAuthData []byte) err
 }
 
 func (c *Conn) acquireCredential() error {
-	if len(c.credential.Passwords) > 0 {
+	if c.credential.hasAnyCredential() {
 		return nil
 	}
 	credential, found, err := c.authHandler.GetCredential(c.user)
 	if err != nil {
 		return err
 	}
-	if !found || len(credential.Passwords) == 0 {
+	if !found || !credential.hasAnyCredential() {
 		return mysql.NewDefaultError(mysql.ER_NO_SUCH_USER, c.user, c.RemoteAddr().String())
 	}
 	c.credential = credential
@@ -81,6 +81,15 @@ func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential C
 			return nil
 		}
 		return ErrAccessDeniedNoPassword
+	}
+
+	// Pre-computed hashes are checked first: they let callers configure
+	// credentials when only the server-side hash is available (no plaintext),
+	// and they're cheaper per connect because we skip the SHA1(SHA1(...)) step.
+	for _, hash := range credential.HashedPasswords {
+		if mysql.CompareNativePassword(clientAuthData, hash, c.salt) {
+			return nil
+		}
 	}
 
 	for _, password := range credential.Passwords {
--- a/server/auth.go
+++ b/server/auth.go
@@ -8,6 +8,7 @@ import (
 	"crypto/subtle"
 	"crypto/tls"
 	"fmt"
+	"log"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
@@ -146,6 +147,15 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	for _, hash := range credential.HashedPasswords {
 		check, err := mysql.Check256HashingPassword(hash, string(clientAuthData))
 		if err != nil {
+			// Stored hashes registered via AddUserWithHashedPassword have
+			// already been shape-checked by validateHashedPassword, so
+			// reaching this branch means either a malformed hash was
+			// installed by constructing Credential directly, or the
+			// upstream verifier changed format expectations. Log so the
+			// silent skip is auditable; auth still falls through to the
+			// plaintext loop and ultimately ErrAccessDenied if nothing
+			// matches.
+			log.Printf("server: sha256_password hash compare error for user %q: %v", c.user, err)
 			continue
 		}
 		if check {
--- a/server/auth.go
+++ b/server/auth.go
@@ -8,9 +8,6 @@ import (
 	"crypto/subtle"
 	"crypto/tls"
 	"fmt"
-	"log"
-
-	"github.com/pingcap/tidb/pkg/parser/auth"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
@@ -78,46 +75,6 @@ func scrambleValidation(cached, nonce, scramble []byte) bool {
 	return subtle.ConstantTimeCompare(m, cached) == 1
 }
 
-// safeNativeCompare wraps mysql.CompareNativePassword in a recover so a
-// malformed stored hash (installed by constructing Credential directly,
-// bypassing validateHashedPassword) cannot crash the connection goroutine.
-// Returns (match, err); on panic, err is non-nil and match is false.
-func safeNativeCompare(clientAuthData, hash, salt []byte) (match bool, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			match = false
-			err = errors.Errorf("CompareNativePassword panicked: %v", r)
-		}
-	}()
-	return mysql.CompareNativePassword(clientAuthData, hash, salt), nil
-}
-
-// safeSha256Check wraps mysql.Check256HashingPassword in a recover for the
-// same reason as safeNativeCompare. The verifier slices into parts[2][:16]
-// which can panic on malformed stored hashes.
-func safeSha256Check(hash []byte, password string) (match bool, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			match = false
-			err = errors.Errorf("Check256HashingPassword panicked: %v", r)
-		}
-	}()
-	return mysql.Check256HashingPassword(hash, password)
-}
-
-// safeCachingSha2Check wraps auth.CheckHashingPassword in a recover. The
-// upstream verifier slices into parts[3][:SALT_LENGTH] which can panic on
-// malformed stored hashes.
-func safeCachingSha2Check(hash []byte, password, plugin string) (match bool, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			match = false
-			err = errors.Errorf("CheckHashingPassword panicked: %v", r)
-		}
-	}()
-	return auth.CheckHashingPassword(hash, password, plugin)
-}
-
 func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential Credential) error {
 	if isEmptyPassword(clientAuthData) {
 		if credential.hasEmptyPassword() {
@@ -130,12 +87,7 @@ func (c *Conn) compareNativePasswordAuthData(clientAuthData []byte, credential C
 	// credentials when only the server-side hash is available (no plaintext),
 	// and they're cheaper per connect because we skip the SHA1(SHA1(...)) step.
 	for _, hp := range credential.HashedPasswords {
-		match, err := safeNativeCompare(clientAuthData, hp.data, c.salt)
-		if err != nil {
-			log.Printf("server: native_password hash compare error for user %q: %v", c.user, err)
-			continue
-		}
-		if match {
+		if mysql.CompareNativePassword(clientAuthData, hp.data, c.salt) {
 			return nil
 		}
 	}
@@ -192,18 +144,15 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 	// (e.g. mirroring `mysql.user.authentication_string`), and we skip
 	// the per-connect hashPassword work.
 	for _, hp := range credential.HashedPasswords {
-		check, err := safeSha256Check(hp.data, string(clientAuthData))
+		check, err := mysql.Check256HashingPassword(hp.data, string(clientAuthData))
 		if err != nil {
-			// Stored hashes registered via AddUserWithHashedPassword have
-			// already been shape-checked by validateHashedPassword, so
-			// reaching this branch means either a malformed hash was
-			// installed by constructing Credential directly (which can
-			// trigger an upstream slice-out-of-bounds panic that
-			// safeSha256Check converts to an error), or the upstream
-			// verifier changed format expectations. Log so the silent
+			// Stored hashes are shape-checked at construction time
+			// (validateHashedPassword via NewHashedPassword), so reaching
+			// this branch implies the upstream verifier changed format
+			// expectations. Log via the configurable logger so the silent
 			// skip is auditable; auth still falls through to the plaintext
 			// loop and ultimately ErrAccessDenied if nothing matches.
-			log.Printf("server: sha256_password hash compare error for user %q: %v", c.user, err)
+			c.serverConf.logger().Error("sha256_password hash compare error", "user", c.user, "error", err)
 			continue
 		}
 		if check {
--- a/server/auth.go
+++ b/server/auth.go
@@ -139,6 +139,20 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
 			clientAuthData = clientAuthData[:l-1]
 		}
 	}
+	// Pre-computed hashes are checked first: callers can configure
+	// credentials when only the server-side stored hash is available
+	// (e.g. mirroring `mysql.user.authentication_string`), and we skip
+	// the per-connect hashPassword work.
+	for _, hash := range credential.HashedPasswords {
+		check, err := mysql.Check256HashingPassword(hash, string(clientAuthData))
+		if err != nil {
+			continue
+		}
+		if check {
+			return nil
+		}
+	}
+
 	for _, password := range credential.Passwords {
 		hash, err := credential.hashPassword(password)
 		if err != nil {
--- a/server/authentication_handler.go
+++ b/server/authentication_handler.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"bytes"
 	"slices"
 	"sync"
 
@@ -52,10 +53,16 @@ func NewInMemoryAuthenticationHandler(defaultAuthMethod ...string) *InMemoryAuth
 // rehoming users from another server's `mysql.user` table) configure credentials when only the stored
 // hash is available. The byte format depends on AuthPluginName:
 //
-//   - mysql_native_password: the 20-byte SHA1(SHA1(plaintext)) value, i.e. what NativePasswordHash
-//     returns, or what DecodePasswordHex returns from MySQL's standard "*XXXX..." (41-char) hex form.
-//
-// Other auth plugins do not currently consume HashedPasswords; for them, populate Passwords instead.
+//   - mysql_native_password: the 20-byte SHA1(SHA1(plaintext)) value, i.e. what
+//     mysql.NativePasswordHash returns, or what mysql.DecodePasswordHex returns from MySQL's standard
+//     "*XXXX..." (41-char) hex form.
+//   - caching_sha2_password: the bytes of the standard "$A$<iter>$<salt>$<hash>" stored form, i.e.
+//     what auth.NewHashPassword(plaintext, AUTH_CACHING_SHA2_PASSWORD) returns. Note that this auth
+//     plugin's full-auth flow requires either TLS or a configured RSA key on the server (same
+//     constraint as plaintext Passwords). After the first successful full auth the server caches
+//     SHA256(SHA256(plaintext)) per user@host so subsequent connections can take the fast-auth path.
+//   - sha256_password: the bytes of the standard "$<iter>$<salt>$<hashHex>" stored form, i.e. what
+//     mysql.NewSha256PasswordHash returns. Same TLS/RSA requirement as caching_sha2_password.
 //
 // Both fields can be set on the same Credential: HashedPasswords is checked first (cheaper, no
 // hashing per connect), then Passwords.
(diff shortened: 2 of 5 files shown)


server/authentication_handler.go
```diff
@@ -119,6 +141,45 @@ func (h *InMemoryAuthenticationHandler) AddUser(username, password string, optio
...
+	}
+	if authPluginName != mysql.AUTH_NATIVE_PASSWORD {
+		// Future work: extend HashedPasswords semantics to caching_sha2_password
+		// and sha256_password. Reject up front so callers don't silently get
+		// "auth always fails" for unsupported plugins.
+		return errors.Errorf("AddUserWithHashedPassword does not yet support auth plugin '%s'; only '%s' is supported", authPluginName, mysql.AUTH_NATIVE_PASSWORD)
+	}
+	if len(hash) == 0 {
+		return errors.New("hashed password must not be empty")
+	}
```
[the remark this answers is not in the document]
takaidohigasi: Good catch — applied in b5cc857. The check now requires exactly `mysqlNativePasswordHashLen` (20) bytes for `mysql_native_password`, with a clearer error message that includes the auth plugin name and the actual length. Added a parameterized test covering nil / empty / 19 bytes / 21 bytes / the easy-to-make mistake of passing the `*XXXX...` hex string as `[]byte` directly without `mysql.DecodePasswordHex`.

server/authentication_handler.go
```diff
@@ -119,6 +141,45 @@ func (h *InMemoryAuthenticationHandler) AddUser(username, password string, optio
...
+
+	h.userPool.Store(username, Credential{
+		HashedPasswords: [][]byte{hash},
```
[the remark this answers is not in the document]
takaidohigasi: Good catch — applied in b5cc857 via `slices.Clone`. Added `TestAddUserWithHashedPasswordCopiesSlice` that mutates the caller's slice after registration and verifies the stored credential is unaffected.

takaidohigasi:
cope with another password plugins.

takaidohigasi:
Pushed `ab12007` — extends `HashedPasswords` support to **`caching_sha2_password`** and **`sha256_password`** in addition to `mysql_native_password`. Single byte-slice API; the shape per plugin is documented on the `Credential` struct:

- **mysql_native_password** — 20 bytes, `SHA1(SHA1(plaintext))`.
- **caching_sha2_password** — bytes of `$A$<iter>$<salt><hash>` (`auth.NewHashPassword` output).
- **sha256_password** — bytes of `$<iter>$<salt>$<hashHex>` (`mysql.NewSha256PasswordHash` output).

**Verification wiring:**

- `compareSha256PasswordAuthData` iterates `HashedPasswords` via `mysql.Check256HashingPassword` before falling back to the plaintext loop.
- `checkSha2CacheCredentials` (caching_sha2 full-auth) iterates via `auth.CheckHashingPassword`. Fast-auth path needs no change — it's a cache lookup; for hashed-only users the cache starts empty so the server falls through to full auth, and after the first successful full auth the cache populates and subsequent connections use fast auth.

**`AddUserWithHashedPassword`** now dispatches per-plugin format validation via a small `validateHashedPassword` helper. Each plugin has explicit length / structure checks tight enough to avoid the existing implementations' fixed-offset slicing panics (e.g. caching_sha2's `parts[3][:SALT_LENGTH]`). `mysql_clear_password` is explicitly redirected to `AddUser` since there's no meaningful hashed form.

**Caveat documented**: caching_sha2 / sha256 require TLS or an RSA key on the server because their full-auth flow sends the plaintext on the wire. Same constraint that already applies to plaintext `Passwords` with these plugins, just spelled out.

**Tests:**

- `TestAddUserWithHashedPassword_CachingSha2` — end-to-end with `auth.NewHashPassword(plaintext, AUTH_CACHING_SHA2_PASSWORD)` as the stored value; two consecutive `Ping`s exercise both the cache-miss full-auth path and the cache-hit fast-auth path under TLS.
- `TestAddUserWithHashedPassword_Sha256Password` — same shape, no cache layer.
- `TestAddUserWithHashedPasswordRejectsBadFormat` — table test, 12 failure modes across all three plugins (wrong length, missing prefix, wrong hash type, non-decimal iterations, salt-only-no-hash, etc.).
- `TestAddUserWithHashedPasswordRejectsUnknownPlugin` (made-up name) and `TestAddUserWithHashedPasswordRejectsClearPassword` (plaintext-only plugin) replace the previous "rejects caching_sha2" test that's no longer applicable.

Full server suite + race detector pass; gofumpt and go vet clean.

dveeden (commented):
Some questions:

1. Do we limit this to two accounts per user or is it an unbounded array of hashes?
2. We don't store the type of each hashed password? Should we? 
3. Looks like we now don't use the `AuthPluginName`, but just try all hashes and password until one matches?
4. Could one account have multiple types of hashes? 
5. Would this allow us to implement the caching part of `caching_sha2` later on where we keep info in memory after the first authentication to allow for fast path in subsequent authentication attempts?

Maybe there should be a `type HashedPasssword struct { data []byte, ht hashType }` and then have `HashedPasswords []HashedPassword` for the `Credential` struct?

dveeden:
Some remarks from Claude:

  - isAuthMethodSupported + validateHashedPassword's default case are redundant. Both emit "unknown authentication plugin name". isAuthMethodSupported could be      
  dropped from AddUserWithHashedPassword since validateHashedPassword already covers it (and handles AUTH_CLEAR_PASSWORD with a better message).
  - The err != nil / continue pattern in the hash loops silently swallows errors (auth.go:97-99, auth.go:148-150, etc.). For stored hashes that the server itself    
  formatted, this should be impossible, but a debug log or a continue comment would clarify intent.                                                                  
  - validateHashedPassword is unexported but handles a wide surface area — if you go with the HashedPassword struct approach, this function could move into a
  constructor like NewHashedPassword(plugin, data) which returns (HashedPassword, error), which is more idiomatic than a bare validation function.

dveeden:
Claude also said that this could panic:

```
dvaneeden@dve-carbon:~/dev/go-mysql/server$ go test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB test
2026/04/29 11:42:48 Received: UseDB 
2026/04/29 11:42:48 Received: UseDB 
2026/04/29 11:42:48 Received: UseDB test
--- FAIL: FuzzCheckHashingPassword (0.00s)
    --- FAIL: FuzzCheckHashingPassword/seed#0 (0.00s)
panic: runtime error: slice bounds out of range [:20] with capacity 0 [recovered, repanicked]

goroutine 427 [running]:
testing.tRunner.func1.2({0xb5ae20, 0x36b2457142d0})
	/usr/lib/golang/src/testing/testing.go:1974 +0x232
testing.tRunner.func1()
	/usr/lib/golang/src/testing/testing.go:1977 +0x349
panic({0xb5ae20?, 0x36b2457142d0?})
	/usr/lib/golang/src/runtime/panic.go:860 +0x13a
github.com/pingcap/tidb/pkg/parser/auth.CheckHashingPassword({0x36b245782965, 0x7, 0x0?}, {0xb9aed2, 0x8}, {0xba4763, 0x15})
	/home/dvaneeden/go/pkg/mod/github.com/pingcap/tidb/pkg/parser@v0.0.0-20260219190905-9b9281fa8d6d/auth/caching_sha2.go:209 +0x2b1
github.com/go-mysql-org/go-mysql/server.FuzzCheckHashingPassword.func1(0x0?, {0x36b245782965?, 0x0?, 0x491213?})
	/home/dvaneeden/dev/go-mysql/server/authentication_handler_test.go:360 +0x3a
reflect.Value.call({0xac22e0?, 0xbdda98?, 0x13?}, {0xb99a3b, 0x4}, {0x36b2453e90b0, 0x2, 0x2?})
	/usr/lib/golang/src/reflect/value.go:586 +0xf06
reflect.Value.Call({0xac22e0?, 0xbdda98?, 0x1100540?}, {0x36b2453e90b0?, 0xb92580?, 0xf241c0?})
	/usr/lib/golang/src/reflect/value.go:369 +0xb9
testing.(*F).Fuzz.func1.1(0x36b24543f208?)
	/usr/lib/golang/src/testing/fuzz.go:341 +0x30a
testing.tRunner(0x36b24543f208, 0x36b24543a360)
	/usr/lib/golang/src/testing/testing.go:2036 +0xea
created by testing.(*F).Fuzz.func1 in goroutine 426
	/usr/lib/golang/src/testing/fuzz.go:328 +0x668
exit status 2
FAIL	github.com/go-mysql-org/go-mysql/server	1.150s
```

```patch
diff --git a/server/authentication_handler_test.go b/server/authentication_handler_test.go
index 8191c19..02f2f49 100644
--- a/server/authentication_handler_test.go
+++ b/server/authentication_handler_test.go
@@ -3,6 +3,7 @@ package server
 import (
        "database/sql"
        "net"
+       "strings"
        "sync/atomic"
        "testing"
        "time"
@@ -334,6 +335,32 @@ func TestAddUserWithHashedPassword_Sha256Password(t *testing.T) {
        require.Error(t, dbBad.Ping())
 }
 
+// FuzzCheckHashingPassword verifies that auth.CheckHashingPassword never panics.
+// The function slices pwhashParts[3][:SALT_LENGTH] (20 bytes) without a length
+// check, so inputs that pass the early error-return guards (4 "$"-parts, type
+// "A", parseable iterations) but carry a short final segment can panic. Whether
+// a specific length panics is capacity-dependent: bytes.Split returns sub-slices
+// of the original backing array, and the fuzz engine allocates each input
+// tightly, making panics reliable.
+//
+// The seeds include known panicking inputs so "go test ./server/" catches the
+// bug without the fuzzer running. Use -fuzz=FuzzCheckHashingPassword to search
+// for additional cases.
+func FuzzCheckHashingPassword(f *testing.F) {
+       // Known panicking inputs: pass all pre-checks but have a short final segment.
+       f.Add([]byte("$A$005$"))                               // empty final segment
+       f.Add([]byte("$A$005$" + strings.Repeat("x", 9)))     // 9-byte final segment
+
+       // Valid hashes as seeds so the fuzzer also explores from known-good shapes.
+       f.Add([]byte(auth.NewHashPassword("password", mysql.AUTH_CACHING_SHA2_PASSWORD)))
+
+       f.Fuzz(func(t *testing.T, hash []byte) {
+               // Must not panic. A panic is caught by the fuzz framework, reported
+               // as a test failure, and the triggering input is saved to the corpus.
+               _, _ = auth.CheckHashingPassword(hash, "password", mysql.AUTH_CACHING_SHA2_PASSWORD)
+       })
+}
+
 func TestOnAuthFailureCalled(t *testing.T) {
        handler := &hookTrackingAuthenticationHandler{
                InMemoryAuthenticationHandler: NewInMemoryAuthenticationHandler(mysql.AUTH_NATIVE_PASSWORD),
```

takaidohigasi:
Thanks @dveeden — these are good questions. Let me go through them:

  **1. Limit to two, or unbounded?** Unbounded, mirroring the existing
  `Passwords []string`. MySQL itself caps at primary + `RETAIN CURRENT
  PASSWORD` secondary, so callers will typically set 1 or 2; nothing in
  the library forces that.

  **2. Storing type per hash?** Currently type is per-`Credential` via
  `AuthPluginName` — all hashes in one Credential share a plugin type,
  which matches `mysql.user.plugin` (one value per user row). So I
  deliberately mirrored that shape rather than promoting type to a
  per-entry property.

  **3. Is `AuthPluginName` still used?** Yes — the handshake negotiation
  selects the plugin, which decides which compare function runs
  (`compareNativePasswordAuthData` vs `compareSha256PasswordAuthData` vs
  `checkSha2CacheCredentials`). Inside each, we iterate `HashedPasswords`
  expecting that plugin's stored format. There's no cross-plugin try-all
  behavior; per-format validation in `validateHashedPassword` rejects
  mixing at registration time.

  **4. Multiple types per account?** Not today — see #2. One Credential =
  one plugin type.

  **5. Future fast-auth caching for `caching_sha2`?** Honest answer: not
  cleanly with the current shape. The fast-auth cache uses
  `SHA256(SHA256(plaintext))` (a different format from the
  `$A$<iter>$<salt><hash>` we store), so pre-warming would need either a
  second field or a per-entry-type discriminator — i.e. exactly your
  proposed `HashedPassword{data, ht}` shape.

  ---

  **On your structural suggestion:** the `[]HashedPassword{...}` design
  is more flexible — it's a clean answer to (5) and unlocks the
  mixed-plugin case in (4). The trade-offs are:

  - More complex API for the common case (one user, one plugin)
  - Inconsistency with existing `Passwords []string` until that's also
    reworked to carry per-entry types
  - Slightly bigger surface to migrate later if we want to support
    the fast-auth cache pre-warm

  How would you like me to scope this PR? Two options:

  **(a)** Land this as-is for the immediate use case (proxy mirroring
  `mysql.user`'s native_password / sha256 / caching_sha2 stored hashes)
  and treat the `[]HashedPassword{data, ht}` redesign as a separate PR
  when fast-auth caching is on the roadmap.

  **(b)** Restructure here to `HashedPasswords []HashedPassword` so
  fast-auth caching can be added incrementally without another API
  change. Happy to do this if you'd prefer — would just be an extra
  commit on top of what's there.

  I'd lean toward (a) for incrementalism, but I'm equally happy with
  (b) if you want the longer-term shape locked in now.

takaidohigasi:
Thanks for the review! Pushed four commits:

- 1a 5f3e98a — drop redundant isAuthMethodSupported gate; validateHashedPassword already covers it.
- 1b 03eabd0 — log unexpected hash-compare errors instead of silently swallowing them.  
- 1c 164d985 — add HashedPassword type + NewHashedPassword constructor; AddUserWithHashedPassword now delegates to it. Kept storage as [][]byte for now — happy to promote to []HashedPassword as a follow-up if you'd prefer.
 - panic 41ddb27 — the panic shape was already blocked by validation; relabeled the existing test as "panic shape", added a sibling case, and documented the bypass caveat on Credential.HashedPasswords.

dveeden:
> How would you like me to scope this PR? Two options:
> 
> **(a)** Land this as-is for the immediate use case (proxy mirroring `mysql.user`'s native_password / sha256 / caching_sha2 stored hashes) and treat the `[]HashedPassword{data, ht}` redesign as a separate PR when fast-auth caching is on the roadmap.
> 
> **(b)** Restructure here to `HashedPasswords []HashedPassword` so fast-auth caching can be added incrementally without another API change. Happy to do this if you'd prefer — would just be an extra commit on top of what's there.
> 
> I'd lean toward (a) for incrementalism, but I'm equally happy with (b) if you want the longer-term shape locked in now.

I don't have a strong opinion on this either. @lance6716 what do you think about this?

server/authentication_handler.go
```diff
@@ -42,14 +43,82 @@ func NewInMemoryAuthenticationHandler(defaultAuthMethod ...string) *InMemoryAuth
...
+//     fast-auth path.
+//   - sha256_password: the bytes of the standard "$<iter>$<salt>$<hashHex>" stored form, i.e. what
+//     mysql.NewSha256PasswordHash returns. Same TLS-or-RSA requirement as caching_sha2_password,
+//     for the same reason.
+//
+// Both fields can be set on the same Credential: HashedPasswords is checked first (cheaper, no
+// hashing per connect), then Passwords.
+//
+// Hashes installed via AddUserWithHashedPassword / NewHashedPassword are
+// shape-checked up front. Constructing a Credential directly and inserting
```
lance6716: > Constructing a Credential directly

If we don't export that field, we can always validate it at the init time. Do you think it's necessary to export it?

And seems `HashedPasswords []HashedPassword` can also solve this problem

server/authentication_handler.go
```diff
@@ -42,14 +43,82 @@ func NewInMemoryAuthenticationHandler(defaultAuthMethod ...string) *InMemoryAuth
...
+// HashedPassword wrapping a defensive copy of the bytes. authPluginName must
+// be one of the supported hash-bearing plugins (mysql_native_password,
+// caching_sha2_password, sha256_password); mysql_clear_password has no
+// hashed form and is rejected. See Credential.HashedPasswords for the
+// expected byte layout per plugin.
+func NewHashedPassword(authPluginName string, hash []byte) (HashedPassword, error) {
+	if err := validateHashedPassword(authPluginName, hash); err != nil {
+		return HashedPassword{}, err
+	}
+	return HashedPassword{plugin: authPluginName, data: slices.Clone(hash)}, nil
```
lance6716, line 107: if we chance the type of `hash` to `string`, string is immutable so don't need to clone it. But it's less natural to store binary data in string 🤔

server/authentication_handler.go
```diff
@@ -79,6 +148,30 @@ func (c Credential) hasEmptyPassword() bool {
...
 }
 
+// hasAnyCredential reports whether the credential has at least one usable
+// entry — either a plaintext password (including the empty string, which
+// signals "empty password is OK") or a pre-computed hash.
+func (c Credential) hasAnyCredential() bool {
+	return len(c.Passwords) > 0 || len(c.HashedPasswords) > 0
+}
+
+// cloneCredential returns a deep copy of c so that callers cannot mutate
```
lance6716: I think the comment is not needed. clone always means that.

server/authentication_handler.go
```diff
@@ -119,6 +216,182 @@ func (h *InMemoryAuthenticationHandler) AddUser(username, password string, optio
...
+func (h *InMemoryAuthenticationHandler) AddUserHashed(username string, hp HashedPassword) error {
+	if err := validateHashedPassword(hp.plugin, hp.data); err != nil {
+		return err
+	}
+	// Defensive copy so a caller that constructed HashedPassword via
+	// NewHashedPassword (which already cloned the input) is still
+	// protected from later AddUserHashed calls accidentally aliasing
+	// the same backing array, and so HashedPassword.Bytes returning a
+	// fresh copy stays consistent with the credential storage model.
+	h.userPool.Store(username, Credential{
```
lance6716, line 270: if we want to add 2 hash to 1 user, calling `AddUserHashed` will overwrite the previous data. Is this expected? And anyway please use function comment to explain the behaviour.

lance6716 (commented):
rest lgtm.

server/authentication_handler.go
takaidohigasi: Good call — went with `HashedPasswords []HashedPassword` in 4506440. Each element is now self-validated at construction (`NewHashedPassword`), so direct struct assignment can no longer install a malformed hash, and the panic-shape caveat that was on the field doc is gone. `HashedPassword.data` stays unexported and `Bytes()` returns a copy, so the value is effectively immutable from outside the package.

server/authentication_handler.go
takaidohigasi: Kept `[]byte` for now — `mysql_native_password` stores raw 20-byte SHA1×2, which is awkward to put in a Go string (it's not text), and the new typed `HashedPassword` wrapper already removes the caller-mutation risk that was the original motivation for considering `string`. Happy to revisit if you'd rather see all three plugins use `string` for symmetry.

server/authentication_handler.go
takaidohigasi: Agreed — dropped in 4506440. The function name `cloneCredential` already says it.

server/authentication_handler.go
takaidohigasi: Good catch — overwrite was the intended behavior (matches `AddUser`'s existing semantics) but was never spelled out. In 4506440: doc on `AddUser`, `AddUserWithHashedPassword`, and `AddUserHashed` now states the overwrite explicitly, and added a separate `AppendUserHashed` for the password-rotation use case (validates the user exists, validates plugin matches, returns an error otherwise). Let me know if you'd prefer different semantics.

server/auth.go
```diff
@@ -75,6 +78,46 @@ func scrambleValidation(cached, nonce, scramble []byte) bool {
 	return subtle.ConstantTimeCompare(m, cached) == 1
 }
 
+// safeNativeCompare wraps mysql.CompareNativePassword in a recover so a
+// malformed stored hash (installed by constructing Credential directly,
+// bypassing validateHashedPassword) cannot crash the connection goroutine.
+// Returns (match, err); on panic, err is non-nil and match is false.
+func safeNativeCompare(clientAuthData, hash, salt []byte) (match bool, err error) {
```
lance6716: After change to `HashedPasswords []HashedPassword`, no need to introduce new safeXXX function?

server/auth.go
```diff
@@ -130,6 +187,30 @@ func (c *Conn) compareSha256PasswordAuthData(clientAuthData []byte, credential C
...
+			// Stored hashes registered via AddUserWithHashedPassword have
+			// already been shape-checked by validateHashedPassword, so
+			// reaching this branch means either a malformed hash was
+			// installed by constructing Credential directly (which can
+			// trigger an upstream slice-out-of-bounds panic that
+			// safeSha256Check converts to an error), or the upstream
+			// verifier changed format expectations. Log so the silent
+			// skip is auditable; auth still falls through to the plaintext
+			// loop and ultimately ErrAccessDenied if nothing matches.
+			log.Printf("server: sha256_password hash compare error for user %q: %v", c.user, err)
```
lance6716: this function always prints to stdout. Please use a configurable logger.

server/auth.go
takaidohigasi: Right, removed in 0d8f39f. With each entry validated at construction (`NewHashedPassword`), no malformed bytes can reach the verifier — `mysql.CompareNativePassword` already returns `false` on empty stored, and the other two verifiers `bytes.Split` first so they fail safely on garbage input. The `safeXXX` recover wrappers became dead weight; inlined the verifier calls.

server/auth.go
takaidohigasi: Good point — added `Server.SetLogger(*slog.Logger)` in 0d8f39f (matching `canal`'s existing pattern) and routed the new hash-compare error logs through it via an internal `logger()` accessor that falls back to `slog.Default()`. Left the existing `log.Printf` calls in `command.go` alone since wiring those through is separate scope; happy to follow up if you'd like that as part of this PR.

server/authentication_handler.go
```diff
@@ -119,6 +204,214 @@ func (h *InMemoryAuthenticationHandler) AddUser(username, password string, optio
...
+// AppendUserHashed adds another hash to an existing user's credential,
+// rather than replacing it. The intended use case is password rotation:
+// install the new hash alongside the old, let in-flight clients drain on
+// the old credential, then prune the old hash.
+//
+// The user must already exist (returns an error otherwise — use
+// AddUserHashed for first-time registration), and hp's auth plugin must
+// match the existing credential's AuthPluginName so a single verifier
+// flow handles every entry.
+func (h *InMemoryAuthenticationHandler) AppendUserHashed(username string, hp HashedPassword) error {
```
lance6716, line 286: instead of providing 2 functions, I think we can use `SetUserHashed(username string, hps ...HashedPassword)`

lance6716 (approved):
Can you provide an example demo of how this will be used? So we can check the new features more end-to-end.

rest lgtm

takaidohigasi:
> Can you provide an example demo of how this will be used? So we can check the new features more end-to-end.

Sure — happy to add context and an end-to-end example.

**The use case in one line:** let the server-side auth handler accept the same stored hash format that's already in `mysql.user.authentication_string`, so a proxy can authenticate clients against upstream's existing credentials without ever holding their plaintext.

**Industry precedent — ProxySQL.** ProxySQL has supported this for years. Its `mysql_users.password` column accepts *either* the plaintext or the native_password hash form (`*ABC123…` 41-char hex). Operators commonly copy `mysql.user.authentication_string` straight from upstream into ProxySQL without ever touching the cleartext, and frontend handshakes still succeed because the comparison happens against the stored hash. This PR brings the equivalent capability to go-mysql's server-side auth handler — and broadens it to `caching_sha2_password` / `sha256_password`, not just native_password.

**Concrete scenario — mysql-interceptor.** I'm landing this so I can use it in [`mysql-interceptor`](https://github.com/takaidohigasi/mysql-interceptor), a MySQL proxy I'm building. Today its `proxy.users` config requires plaintext:

```yaml
proxy:
  users:
    - username: app_user
      password: "app_pw"   # plaintext today
```

After this PR lands I plan to add a `password_hash` alternative so operators can mirror upstream's `authentication_string` directly — same pattern ProxySQL operators already use:

```yaml
proxy:
  users:
    - username: app_user
      password_hash: "*6BB4837EB74329105EE4568DDA7DC67ED2CA2AD9"   # native_password
    - username: legacy_user
      password_hash: "$A$005$..."                                   # caching_sha2_password
```

…which under the hood translates to `mysql.DecodePasswordHex` + `server.NewHashedPassword(plugin, bytes)` + `handler.AddUserHashed(username, hp)` — the exact public surface this PR adds. Combined with `AppendUserHashed`, operators also get a clean rotation path: install the new hash alongside the old, drain in-flight clients on the old credential, then `AddUserHashed`-replace with just the new hash.

**End-to-end demo.** Here's a runnable example inline so it's visible without checking out a branch. ~150 lines, self-contained, runs against an in-process `server.NewDefaultServer` + `client.Connect` handshake on a random port.

<details>
<summary><code>examples/server_hashed_password/main.go</code></summary>

```go
// Demo for the HashedPassword API added in PR #1129.
//
// This program shows how to authenticate a MySQL client when the server
// only has the *stored-form* hash, not the plaintext — the use case of a
// MySQL proxy that rehomes credentials from another server's mysql.user
// table. The plaintext never touches the server side; the handshake
// hashes it client-side and the verifier compares against the stored
// SHA1(SHA1(plaintext)) value.
//
// Walkthrough:
//
//  1. Compute (or, in real use, load) the stored hash for a user.
//  2. Wrap it in a HashedPassword via NewHashedPassword.
//  3. Register the user via AddUserHashed.
//  4. Connect from client.Connect with the plaintext password — the
//     handshake succeeds without the server ever seeing it.
//  5. Demonstrate password rotation via AppendUserHashed: install the
//     new password's hash alongside the old, verify both authenticate
//     during the grace window, then drop the old hash by re-registering
//     with just the new one.
//
// Run: go run ./examples/server_hashed_password
package main

import (
	"fmt"
	"log"
	"net"

)

// minimalHandler embeds EmptyHandler so we only have to override the
// handful of commands the demo actually exercises (USE / SELECT).
type minimalHandler struct {
	server.EmptyHandler
}

func (h *minimalHandler) HandleQuery(query string) (*mysql.Result, error) {
	rs, err := mysql.BuildSimpleTextResultset(
		[]string{"answer"},
		[][]any{{int64(42)}},
	)
	if err != nil {
		return nil, err
	}
	return &mysql.Result{Resultset: rs}, nil
}

func main() {
	const (
		username     = "alice"
		oldPlaintext = "old-secret"
		newPlaintext = "new-secret"
	)

	// 1. Compute the stored hash. In real use this would come from your
	//    source of truth — e.g. the mysql.user row of the upstream
	//    server you're proxying — and you'd never see the plaintext at
	//    all. We synthesize it here so the demo stays self-contained.
	oldStoredHash := mysql.NativePasswordHash([]byte(oldPlaintext))
	log.Printf("step 1: stored hash = % x  (sha1(sha1(%q)))", oldStoredHash, oldPlaintext)

	// 2. Build an authentication handler and wrap the hash.
	authHandler := server.NewInMemoryAuthenticationHandler(mysql.AUTH_NATIVE_PASSWORD)
	oldHP, err := server.NewHashedPassword(mysql.AUTH_NATIVE_PASSWORD, oldStoredHash)
	if err != nil {
		log.Fatalf("NewHashedPassword: %v", err)
	}

	// 3. Register the user. The server now knows how to verify alice
	//    without holding her plaintext.
	if err := authHandler.AddUserHashed(username, oldHP); err != nil {
		log.Fatalf("AddUserHashed: %v", err)
	}
	log.Printf("step 3: registered %q via AddUserHashed (no plaintext on the server)", username)

	// 4. Spin up a server on a random port and try a real handshake.
	ln, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		log.Fatalf("listen: %v", err)
	}
	defer ln.Close()
	addr := ln.Addr().String()

	srv := server.NewDefaultServer()
	cmdHandler := &minimalHandler{}
	go acceptLoop(ln, srv, authHandler, cmdHandler)

	// runOnce dials, runs SELECT 42, and reports auth + query success.
	runOnce := func(label, user, pass string) error {
		c, err := client.Connect(addr, user, pass, "")
		if err != nil {
			return fmt.Errorf("%s: connect: %w", label, err)
		}
		defer c.Close()
		r, err := c.Execute("SELECT 42")
		if err != nil {
			return fmt.Errorf("%s: execute: %w", label, err)
		}
		got, _ := r.GetIntByName(0, "answer")
		if got != 42 {
			return fmt.Errorf("%s: expected 42, got %d", label, got)
		}
		return nil
	}

	if err := runOnce("step 4 (old password)", username, oldPlaintext); err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("step 4: ✔ authenticated with plaintext %q against the stored hash", oldPlaintext)

	// 5a. Password rotation: install a new hash alongside the old one.
	//     The old hash stays installed so in-flight clients still
	//     authenticate during the rotation window.
	newStoredHash := mysql.NativePasswordHash([]byte(newPlaintext))
	newHP, err := server.NewHashedPassword(mysql.AUTH_NATIVE_PASSWORD, newStoredHash)
	if err != nil {
		log.Fatalf("NewHashedPassword (new): %v", err)
	}
	if err := authHandler.AppendUserHashed(username, newHP); err != nil {
		log.Fatalf("AppendUserHashed: %v", err)
	}
	log.Printf("step 5a: appended a second hash — both passwords accept now")

	if err := runOnce("step 5a (old)", username, oldPlaintext); err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("step 5a: ✔ old password still works (rotation grace window)")
	if err := runOnce("step 5a (new)", username, newPlaintext); err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("step 5a: ✔ new password also works")

	// 5b. Once clients have moved over, drop the old hash by
	//     re-registering with just the new one. AddUserHashed replaces
	//     the entire credential, so the old hash is gone.
	if err := authHandler.AddUserHashed(username, newHP); err != nil {
		log.Fatalf("AddUserHashed (replace): %v", err)
	}
	log.Printf("step 5b: re-registered %q with only the new hash — rotation complete", username)

	if err := runOnce("step 5b (old, must fail)", username, oldPlaintext); err == nil {
		log.Fatalf("BUG: old password authenticated after rotation completed")
	}
	log.Printf("step 5b: ✔ old password correctly rejected")
	if err := runOnce("step 5b (new)", username, newPlaintext); err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("step 5b: ✔ new password authenticates")

	fmt.Println("OK — full hashed-password lifecycle demonstrated end-to-end.")
}

// acceptLoop runs the server side of the handshake for each incoming
// connection and then drives the command loop until the client closes.
func acceptLoop(ln net.Listener, srv *server.Server, ah server.AuthenticationHandler, h server.Handler) {
	for {
		conn, err := ln.Accept()
		if err != nil {
			return
		}
		go func(c net.Conn) {
			defer c.Close()
			sc, err := server.NewCustomizedConn(c, srv, ah, h)
			if err != nil {
				log.Printf("handshake: %v", err)
				return
			}
			for sc.HandleCommand() == nil {
			}
		}(conn)
	}
}
```

</details>

Expected output when run (`go run ./examples/server_hashed_password`):

```
step 1: stored hash = 41 c9 7f 0e 87 85 a0 e0 14 4d 14 30 ec 33 c0 f0 8f 33 18 1d  (sha1(sha1("old-secret")))
step 3: registered "alice" via AddUserHashed (no plaintext on the server)
step 4: ✔ authenticated with plaintext "old-secret" against the stored hash
step 5a: appended a second hash — both passwords accept now
step 5a: ✔ old password still works (rotation grace window)
step 5a: ✔ new password also works
step 5b: re-registered "alice" with only the new hash — rotation complete
handshake: ERROR 1045 (28000): Access denied for user 'alice'@...
step 5b: ✔ old password correctly rejected
step 5b: ✔ new password authenticates
OK — full hashed-password lifecycle demonstrated end-to-end.
```

(The `ERROR 1045` is the *server* logging the deliberate access-denied probe in step 5b — that's the demo proving rotation actually drops the old hash, not a failure.)

### Quick scoping question on the demo

I've kept this PR focused on the API and committed the demo only locally. How would you prefer it to land?

1. **Push it into this PR** as `examples/server_hashed_password/` — lives next to the existing `examples/stream_result/` precedent, runnable directly with `go run ./examples/server_hashed_password`.
2. **Push it on a separate branch** in my fork — keeps this PR scoped to the API change; reviewers who want to run it `git fetch` it.
3. **Leave it inline here** in this comment — read-only, no extra commits.

Happy with any of the three; just want to match the project's preference for example-code placement.

server/authentication_handler.go
takaidohigasi: > instead of providing 2 functions, I think we can use `SetUserHashed(username string, hps ...HashedPassword)`

Agreed. Locking in a few design points before re-rolling:

1. **Zero args** → error (`"at least one HashedPassword is required"`); explicit user deletion deserves its own method.
2. **Plugin consistency** → all variadic hashes must share the same plugin, stored as `Credential.AuthPluginName`. Preserves the one-credential-one-plugin invariant the verifier flows assume.
3. **`AddUserWithHashedPassword`** — should this also go, or stay as a raw-bytes shim around `NewHashedPassword` + `SetUserHashed`? Dropping it leaves the smallest possible API (`NewHashedPassword` + `SetUserHashed`); keeping it spares callers who haven't moved to the typed form.

Will push as a follow-up commit — flag if any of (1)–(3) should land differently.

lance6716: 1. For args, maybe

```
SetUserHashed(username string, hp HashedPassword, hps ...HashedPassword)
```

So golang can make sure the call site has at lease 1 HashedPassword

2. Plugin consistency

If the inner implementation has such restriction, I think it's OK.

3. AddUserWithHashedPassword

I'm neutral about it.

PR #1131, opened 2026-04-30
Clear stale `*Field` entries on `Resultset.Reset`

ramnes:
Right now `Resultset.Reset` re-slices `r.Fields` within the pooled backing array without clearing entries.

So builders that only fill nil slots would reuse a previous call's `Name` / `Charset` / `Flag` from whatever `Resultset` `sync.Pool` happened to return, which would corrupt column metadata across pool reuse.

Let's clear after the re-slice so that the slot is always nil before the builder fills it.

Note: This could be seen as a contract change, because external callers pooling `Resultset` instances and re-using the same `*Field` pointers across `Reset` calls would now have to repopulate `r.Fields` afterwards.

### DIFF
--- a/mysql/resultset.go
+++ b/mysql/resultset.go
@@ -53,6 +53,8 @@ func (r *Resultset) returnToPool() {
 	resultsetPool.Put(r)
 }
 
+// Reset prepares the resultset for reuse. After Reset, all r.Fields entries
+// are nil and the caller must repopulate them.
 func (r *Resultset) Reset(fieldsCount int) {
 	r.RawPkg = r.RawPkg[:0]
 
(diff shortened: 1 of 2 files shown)


lance6716:
Can you explain your usage? I see some caller will reuse `Fields` struct and overwrite it instead of creating a new one.

https://github.com/go-mysql-org/go-mysql/blob/961379bc91cae8f54c85979747ef2c747da61f52/client/resp.go#L422

So for performance consideration, the old behavior makes sense.

ramnes:
Context was that I was preparing a follow-up PR to #1132, in which I ended up silently reusing stale field metadata from the pooled `Resultset`, so I thought it would be a good idea to fix it here directly for everyone.

But yeah, this PR would indeed break the client as-is, good catch. And probably other places? I can move the logic in the new PR instead, and it probably makes sense to favor performance over defensing programming here... Going ahead and closing this one, sorry for the noise. :)

PR #1132, opened 2026-04-30
Type-aware encoder for streaming binary rows

ramnes:
Currently, the streaming binary path produces malformed wire bytes for several column types:

* narrow integers (`TINY`, `SHORT`, `INT24`, `LONG`) get 8 bytes regardless of declared width
* `FLOAT` is encoded as `Float64bits` truncated to 4 bytes, which is wrong
* variable-length types other than `VAR_STRING` get unframed payloads
* temporals are sent as parser-formatted ASCII instead of the packed binary form

This PR adds a new `EncodeBinaryFieldValue` function and routes `writeStreamBinaryRows` through it. This new encoder honors the declared column type for width, framing, and packing.

Out-of-range and malformed inputs are now rejected loudly instead of silently corrupting the wire:

* per-type integer widths with range checks
* `Float32bits` for `FLOAT`, with overflow rejection
* length-encoded variable-width types
* packed temporals, re-packed from formatted strings as needed
* `MYSQL_TYPE_YEAR` forced unsigned
* `bool` only for `MYSQL_TYPE_TINY`
* empty strings, trailing whitespace and dots, microseconds over 6 digits, and `TIME` hours over 838 are rejected

`BuildSimpleBinaryResultset` is untouched, but consolidating it onto the new encoder is coming in a follow-up PR.

### DIFF
--- a/mysql/resultset_helper.go
+++ b/mysql/resultset_helper.go
@@ -137,7 +137,11 @@ func toBinaryDate(t time.Time) ([]byte, error) {
 	if t.IsZero() {
 		return []byte{0}, nil
 	}
-	return packDate(uint16(t.Year()), uint8(t.Month()), uint8(t.Day())), nil
+	year := t.Year()
+	if year < 0 || year > 9999 {
+		return nil, errors.Errorf("year %d out of range (0-9999)", year)
+	}
+	return packDate(uint16(year), uint8(t.Month()), uint8(t.Day())), nil
 }
 
 // toBinaryTime encodes a time.Time as a length-prefixed packed binary TIME.
@@ -159,8 +163,12 @@ func toBinaryDateTime(t time.Time) ([]byte, error) {
 	if t.IsZero() {
 		return []byte{0}, nil
 	}
+	year := t.Year()
+	if year < 0 || year > 9999 {
+		return nil, errors.Errorf("year %d out of range (0-9999)", year)
+	}
 	return packDateTime(
-		uint16(t.Year()), uint8(t.Month()), uint8(t.Day()),
+		uint16(year), uint8(t.Month()), uint8(t.Day()),
 		uint8(t.Hour()), uint8(t.Minute()), uint8(t.Second()),
 		uint32(t.Nanosecond()/1000),
 	), nil
--- a/mysql/resultset_helper.go
+++ b/mysql/resultset_helper.go
@@ -1,10 +1,10 @@
 package mysql
 
 import (
-	"bytes"
-	"encoding/binary"
+	"database/sql/driver"
 	"math"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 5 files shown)


dveeden:
For the failing check:

```
dvaneeden@dve-carbon:~/dev/go-mysql$ golangci-lint run
mysql/resultset_helper_test.go:547:1: File is not properly formatted (gofumpt)

^
1 issues:
* gofumpt: 1
dvaneeden@dve-carbon:~/dev/go-mysql$ gofumpt -w mysql/resultset_helper_test.go
dvaneeden@dve-carbon:~/dev/go-mysql$ git diff
```
```patch
diff --git a/mysql/resultset_helper_test.go b/mysql/resultset_helper_test.go
index bc91a63..fa86d08 100644
--- a/mysql/resultset_helper_test.go
+++ b/mysql/resultset_helper_test.go
@@ -544,4 +544,3 @@ func TestEncodeBinaryFieldValueTypedNilSlice(t *testing.T) {
                require.Equal(t, "after", string(row[1].AsString()))
        })
 }
-
```

mysql/resultset_helper.go
```diff
@@ -51,45 +51,603 @@ func FormatTextValue(value any) ([]byte, error) {
...
+		hour, minute, sec,
+	}
+}
+
+// toBinaryDate encodes a time.Time as a length-prefixed packed binary DATE.
+func toBinaryDate(t time.Time) ([]byte, error) {
 	if t.IsZero() {
+		return []byte{0}, nil
+	}
+	return packDate(uint16(t.Year()), uint8(t.Month()), uint8(t.Day())), nil
```
dveeden: We should check the range of the year here. Both for too high and negative values.

ramnes: Good catch, thanks! Fixed.

ramnes:
@dveeden Thanks for the review, both fixed!

mysql/resultset_helper.go
```diff
@@ -51,45 +51,611 @@ func FormatTextValue(value any) ([]byte, error) {
...
+
+	case MYSQL_TYPE_DECIMAL, MYSQL_TYPE_NEWDECIMAL,
+		MYSQL_TYPE_VARCHAR, MYSQL_TYPE_VAR_STRING, MYSQL_TYPE_STRING,
+		MYSQL_TYPE_TINY_BLOB, MYSQL_TYPE_MEDIUM_BLOB, MYSQL_TYPE_LONG_BLOB, MYSQL_TYPE_BLOB,
+		MYSQL_TYPE_BIT, MYSQL_TYPE_ENUM, MYSQL_TYPE_SET,
+		MYSQL_TYPE_VECTOR, MYSQL_TYPE_GEOMETRY, MYSQL_TYPE_JSON:
+		switch v := value.(type) {
+		case []byte:
+			if v == nil {
+				// Typed-nil []byte → NULL. The top-level nil check doesn't
```
lance6716: [P1] `[]byte(nil)` is treated as NULL here, but `RowData.ParseBinary` can produce `[]byte(nil)` for *empty* length-encoded strings (len=0) via `append(dst[:0], v...)` when `v` is empty. In a proxy flow (parse → `FieldValue.Value()` → encode), empty strings can therefore be re-encoded as NULL (NULL-bitmap set, no payload). Consider preserving empty-string vs NULL by ensuring the parser returns a non-nil zero-length slice for empty strings, or by encoding `[]byte(nil)` as length=0 for string-like column types and reserving NULL for nil interface / `driver.Valuer`→nil / `MYSQL_TYPE_NULL`.

(AI review)

lance6716:
Left an inline note on a potential empty-string vs NULL re-encoding edge case. I’ll do a full human review of the PR soon.

mysql/resultset_helper.go
```diff
@@ -51,45 +51,611 @@ func FormatTextValue(value any) ([]byte, error) {
...
+	}
+}
+
+// toBinaryDate encodes a time.Time as a length-prefixed packed binary DATE.
+func toBinaryDate(t time.Time) ([]byte, error) {
+	if t.IsZero() {
+		return []byte{0}, nil
+	}
+	year := t.Year()
+	if year < 0 || year > 9999 {
```
lance6716, line 143: Can you explain how this range limit is chosen? Maybe it's from the 4 digits of YEAR type?

ref https://dev.mysql.com/doc/refman/8.0/en/year.html

mysql/resultset_helper.go
```diff
@@ -51,45 +51,611 @@ func FormatTextValue(value any) ([]byte, error) {
...
+		return nil, errors.Errorf("year %d out of range (0-9999)", year)
+	}
+	return packDate(uint16(year), uint8(t.Month()), uint8(t.Day())), nil
+}
 
+// toBinaryTime encodes a time.Time as a length-prefixed packed binary TIME.
+// For negative or >23h values, pass a string instead.
```
lance6716: > pass a string

Do you mean `parseTimeString`?

mysql/resultset_helper.go
```diff
@@ -51,45 +51,611 @@ func FormatTextValue(value any) ([]byte, error) {
...
+	}
+	negative := strings.HasPrefix(s, "-")
+	if negative {
+		s = s[1:]
+	}
+	hms, frac, hasDot := strings.Cut(s, ".")
+	if hasDot && frac == "" {
+		return nil, errors.Errorf("invalid TIME %q: trailing dot without fractional digits", s)
+	}
+	parts := strings.Split(hms, ":")
```
lance6716: Seems can reuse parseHMSFraction

mysql/resultset_helper.go
```diff
@@ -51,45 +51,611 @@ func FormatTextValue(value any) ([]byte, error) {
...
+	}
+	return uint8(hour), uint8(minute), uint8(sec), micro, nil
+}
+
+func parseMicroseconds(s string) (uint32, error) {
+	if len(s) > 6 {
+		// MySQL fractional seconds top out at 6 digits.
+		return 0, errors.Errorf("invalid microseconds %q: more than 6 digits", s)
+	}
+	for len(s) < 6 {
```
lance6716: nit: maybe * power of 10 after ParseUint

lance6716 (commented):
will review soon

mysql/resultset_helper.go
ramnes: Valid issue. A nil `[]byte` for a string column now encodes as an empty length-encoded string. Added a regression test covering the parse -> encode round-trip.

mysql/resultset_helper.go
ramnes: It's not the YEAR type, it's the DATE/DATETIME range. MySQL caps those at year 9999 and the wire year is a 2-byte unsigned int, hence 0–9999. Added a comment with the doc link.

mysql/resultset_helper.go
ramnes: Yep, reworded.

mysql/resultset_helper.go
ramnes: Done. Generalized `parseHMSFraction` to return the hour unbounded so both DATETIME and TIME go through it, each applying its own max (23 and 838).

mysql/resultset_helper.go
ramnes: done

singulasar:
Running into issues that this PR would resolve, any timeline on this?

lance6716:
Sorry I've been busy these days and I forget this PR 😢 maybe next week...

lance6716:
Hi @ramnes can you merge master? This is necessary for the future merging of PR if I finish the review

ramnes:
@lance6716 rebased

lance6716 (approved):

PR #1144, opened 2026-05-20
canal: add transactional position sync

### LINKED ISSUE #1090: OnPosSynced() called on BEGIN may cause reader to skip a transaction
Canal invokes `OnPosSynced()` on every Query event. That’s generally fine with SBR events, since readers probably don't care where they pause and resume from within an SBR stream.

However, MySQL also uses BEGIN - which is a Query event too - to mark the start of RBR events. Readers may actually have reason to care about processing these transactions atomically. Unfortunately, `OnPosSynced()` provides no context about why it was triggered, leaving the reader no choice but to sync and advance the position.

This becomes especially problematic for GTID‑based readers. Calling `OnPosSynced()` on the BEGIN event advances the GTID set by one entire transaction. If the reader crashes before reaching the next `OnPosSynced()` call - one triggered by the subsequent XID event - it will restart from the next transaction, silently skipping the previous one, or part of it.

A straightforward fix could be passuig `force=true` to `OnPosSynced()` invoked on XID events.

lance6716:
Fixes #1090.

## Summary
- Add `Config.TransactionalPosSync` as an opt-in mode for crash-safe Canal position sync semantics.
- In that mode, skip `OnPosSynced` for `BEGIN` QueryEvent so the GTID set is not advanced before the transaction is processed.
- Report XID position sync with `force=true`, and avoid the final forced `OnPosSynced` from `Canal.Close()`.
- Add offline unit tests for legacy behavior, transactional behavior, XID force sync, close behavior, and `BEGIN` detection.

## Validation
- `go test ./canal -run 'TestIsBeginQuery|TestHandleEventBeginPositionSync|TestHandleEventXIDForcePositionSync|TestClosePositionSync|TestGetShowBinaryLogQuery'`
- `go test ./canal -run '^$'`
- `go test ./... -run '^$'`
- `git diff --check`

Full `go test ./canal` requires a local MySQL instance; without one it fails at `127.0.0.1:3306: connect: connection refused`.

### DIFF
--- a/canal/config.go
+++ b/canal/config.go
@@ -97,8 +97,7 @@ type Config struct {
 	DisableFlushBinlogWhileWaiting bool `toml:"disable_flush_binlog_while_waiting"`
 
 	// TransactionalPosSync avoids syncing positions that may belong to an unfinished transaction.
-	// When enabled, BEGIN QueryEvent does not call OnPosSynced, XIDEvent calls it with force=true,
-	// and Close does not force a final OnPosSynced call.
+	// When enabled, commit events call OnPosSynced with force=true.
 	TransactionalPosSync bool `toml:"transactional_pos_sync"`
 
 	// Set TLS config
--- a/canal/config.go
+++ b/canal/config.go
@@ -96,6 +96,11 @@ type Config struct {
 	// and requires additional privileges.
 	DisableFlushBinlogWhileWaiting bool `toml:"disable_flush_binlog_while_waiting"`
 
+	// TransactionalPosSync avoids syncing positions that may belong to an unfinished transaction.
+	// When enabled, BEGIN QueryEvent does not call OnPosSynced, XIDEvent calls it with force=true,
+	// and Close does not force a final OnPosSynced call.
+	TransactionalPosSync bool `toml:"transactional_pos_sync"`
+
 	// Set TLS config
 	TLSConfig *tls.Config
 
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -13,7 +13,10 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/ast"
 )
 
-var beginQuery = []byte("BEGIN")
+var (
+	beginQuery  = []byte("BEGIN")
+	commitQuery = []byte("COMMIT")
+)
 
 func (c *Canal) startSyncer() (*replication.BinlogStreamer, error) {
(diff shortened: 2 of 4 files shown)


canal/config.go
```diff
@@ -96,6 +96,11 @@ type Config struct {
 	// and requires additional privileges.
 	DisableFlushBinlogWhileWaiting bool `toml:"disable_flush_binlog_while_waiting"`
 
+	// TransactionalPosSync avoids syncing positions that may belong to an unfinished transaction.
+	// When enabled, BEGIN QueryEvent does not call OnPosSynced, XIDEvent calls it with force=true,
```
lance6716: I think here we only need to mention on XIDEvent which means txn commit it will call force=true. No need to bother say other cases.

canal/sync.go
```diff
@@ -185,6 +194,13 @@ func (c *Canal) handleEvent(ev *replication.BinlogEvent) error {
 	return nil
 }
 
+func isBeginQuery(query []byte) bool {
```
lance6716: I think the event is always "BEGIN", so such variant. Please simplify

lance6716 (commented):
1. And for non-transactional engines like MyISAM, COMMIT query event is used to mean transaction commit. Should also support it.
2. Remove the new unit test which is too trival.

lance6716:
ptal @dbnski @dveeden

lance6716:
@serprex I have merged #1153. Can you take a look at this PR? For example, do you think it's good to add a new config item, the default behaviour will not run this logic?

```
			switch stmt.(type) {
			case *ast.BeginStmt, *ast.SavepointStmt:
				// transaction not yet complete; checkpointing here would skip it on GTID resume
				continue
			}
```

serprex:
imo you should drop this PR & take #1153 as the correct way about things

but take my opinion with a grain of salt, peerdb doesn't use canal, so I have no context on its use / design

lance6716:
> imo you should drop this PR & take #1153 as the correct way about things
> 
> but take my opinion with a grain of salt, peerdb doesn't use canal, so I have no context on its use / design

We also don't use canal API, so I can't decide if it's worth adding an option to keep wrong but more compatible behaviour.

@refat75 @dbnski @dveeden WDYT

dbnski:
I can't imagine circumstances in which the old behavior would be useful, nor the new one breaking enough to need have a choice between them.

PR #1145, opened 2026-05-20
canal: fix ignore_tables separator and accept quoted identifiers

### LINKED ISSUE #1143: DumpConfig  配置的解析问题
配置说明是用点，但是代码解析用的是逗号
// Ignore table format is db.table
IgnoreTables []string `toml:"ignore_tables"`

解析的部分代码
for _, ignoreTable := range c.cfg.Dump.IgnoreTables {
		if seps := strings.Split(ignoreTable, ","); len(seps) == 2 {
			c.dumper.AddIgnoreTables(seps[0], seps[1])
		}
	}

dveeden:
The DumpConfig.IgnoreTables docs say entries use the form db.table, but the parser split on "," instead of ".", so the setting never took effect. Parse entries with a small MySQL identifier parser so the documented db.table form works, along with backtick-quoted identifiers like `db`.`table` that may contain dots or other special characters.

Fixes #1143

### DIFF
--- a/canal/canal.go
+++ b/canal/canal.go
@@ -181,7 +181,7 @@ func (c *Canal) prepareDumper() error {
 		if db, table, ok := splitDBTable(ignoreTable); ok {
 			c.dumper.AddIgnoreTables(db, table)
 		} else {
-			slog.Warn("failed to parse ignored tables", "ignoreTable", ignoreTable)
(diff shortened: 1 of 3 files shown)


dveeden:
@gemini could you review again?

lance6716 (commented):
I prefer https://github.com/go-mysql-org/go-mysql/pull/1146 😄

PR #1147, opened 2026-05-22
Add option RenderJSONAsMySQLText

### LINKED ISSUE #1142: replication: JSON decoder loses type information on round-trip back to MySQL
The JSON binary decoder in `replication/json_binary.go` walks JSONB into Go values (`float64`, `shopspring/decimal`, `map[string]any`) and then `json.Marshal`s the tree. That step drops the original MySQL JSON type tags, so when the decoded string is fed back into a MySQL JSON column the stored binary changes:

```
JSONB_DOUBLE 1.0           -> float64(1.0) -> "1"       (MySQL: JSON INTEGER)
JSONB_OPAQUE/NEWDECIMAL 1.0 -> Go string   -> "\"1.0\""  (MySQL: JSON STRING)
```

This breaks spirit, which uses replication to capture row changes and re-apply them to a shadow table. See: https://github.com/block/spirit/issues/842 (Note: there was no corruption here: the discrepancy was picked up by a checksum we use).

We are currently using a tracking fork of go-mysql, but are looking to upstream our change, which was to add an opt in `BinlogSyncerConfig.RenderJSONAsMySQLText bool` that emits text directly from the JSONB byte stream.

Note that there are some existing flags here, but they only partially fix this:

* `UseFloatWithTrailingZero=true` covers the DOUBLE case.
* `UseDecimal=true` doesn't help — `shopspring/decimal.Decimal` marshals as a quoted string by default, and `Decimal.String()` trims trailing zeros.

Happy to open a PR and contribute my changes (https://github.com/morgo/go-mysql/tree/mysql-text-json-rendering) or discuss an alternative fix.

morgo:
Fixes https://github.com/go-mysql-org/go-mysql/issues/1142

The implementation is different from the draft I proposed when creating the initial issue, and it re-uses the existing json_binary decoder.

### DIFF
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -141,7 +141,10 @@ func jsonbGetValueEntrySize(isSmall bool) int {
 // common JSON encoding data. When RenderJSONAsMySQLText is set on the
 // parent RowsEvent the decoder wraps leaf values in MySQL-text marshalers
 // (see json_mysql_text.go) so json.Marshal produces MySQL's textual JSON
-// form, preserving each JSONB value's original type tag.
+// form, faithful to each JSONB value's original type tag where the JSON
+// text grammar can express it. NEWDECIMAL is the one tag that cannot be
+// preserved on text round-trip (no decimal literal in JSON); see
+// BinlogSyncerConfig.RenderJSONAsMySQLText for the full caveat list.
 func (e *RowsEvent) decodeJSONBinary(data []byte) ([]byte, error) {
 	d := jsonBinaryDecoder{
 		useDecimal:               e.useDecimal,
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -138,7 +138,20 @@ func jsonbGetValueEntrySize(isSmall bool) int {
 
 // decodeJSONBinary decodes the JSON binary encoding data and returns
 // the common JSON encoding data.
+//
+// When RenderJSONAsMySQLText is set on the parent RowsEvent the bytes are
+// rendered directly to MySQL-style JSON text via the renderer in
+// json_mysql_text.go, preserving each JSONB value's original type tag
+// (INT vs DOUBLE-with-trailing-zero vs OPAQUE/NEWDECIMAL etc.). Otherwise
+// the legacy path runs: decode to a Go value tree, then json.Marshal.
+// The two paths produce different byte sequences for the same input (the
+// renderer adds spaces after ':' and ',', preserves "1.0" doubles, etc.)
+// -- callers comparing outputs byte-for-byte should pick one and stick.
 func (e *RowsEvent) decodeJSONBinary(data []byte) ([]byte, error) {
+	if e.renderJSONAsMySQLText {
+		return renderJSONAsMySQLText(data, e.ignoreJSONDecodeErr)
+	}
+
 	d := jsonBinaryDecoder{
 		useDecimal:               e.useDecimal,
 		useFloatWithTrailingZero: e.useFloatWithTrailingZero,
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -151,7 +151,7 @@ func (e *RowsEvent) decodeJSONBinary(data []byte) ([]byte, error) {
 	}
 
 	if d.isDataShort(data, 1) {
-		if d.mysqlTextMode && d.ignoreDecodeErr {
+		if d.ignoreDecodeErr {
 			return []byte("null"), nil
 		}
 		return nil, d.err
@@ -159,7 +159,7 @@ func (e *RowsEvent) decodeJSONBinary(data []byte) ([]byte, error) {
 
 	v := d.decodeValue(data[0], data[1:])
 	if d.err != nil {
-		if d.mysqlTextMode && d.ignoreDecodeErr {
+		if d.ignoreDecodeErr {
 			return []byte("null"), nil
 		}
 		return nil, d.err
@@ -488,7 +488,7 @@ func (d *jsonBinaryDecoder) decodeOpaque(data []byte) any {
 	case mysql.MYSQL_TYPE_DATE:
 		// Historically dates have been decoded the same as datetime (including the time portion).
 		// This is mostly harmless, but in text-mode we want to ensure that
-		// the time portion is emitted.
+		// the time portion is omitted.
 		return d.decodeDateTime(data, d.mysqlTextMode)
 	case mysql.MYSQL_TYPE_DATETIME, mysql.MYSQL_TYPE_TIMESTAMP:
 		return d.decodeDateTime(data, false)
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -1,6 +1,7 @@
 package replication
 
 import (
+	"encoding/base64"
 	"fmt"
 	"math"
 	"strconv"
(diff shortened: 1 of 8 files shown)


replication/json_binary.go
```diff
@@ -136,21 +137,31 @@ func jsonbGetValueEntrySize(isSmall bool) int {
...
 func (e *RowsEvent) decodeJSONBinary(data []byte) ([]byte, error) {
 	d := jsonBinaryDecoder{
 		useDecimal:               e.useDecimal,
 		useFloatWithTrailingZero: e.useFloatWithTrailingZero,
 		ignoreDecodeErr:          e.ignoreJSONDecodeErr,
+		mysqlTextMode:            e.renderJSONAsMySQLText,
 	}
 
 	if d.isDataShort(data, 1) {
+		if d.mysqlTextMode && d.ignoreDecodeErr {
```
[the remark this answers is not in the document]
morgo: Good catch — applied. Dropped the `mysqlTextMode` guard so `IgnoreJSONDecodeError` behaves consistently for top-level short payloads in both modes (a51b652).

replication/json_binary.go
```diff
@@ -136,21 +137,31 @@ func jsonbGetValueEntrySize(isSmall bool) int {
...
+			return []byte("null"), nil
+		}
 		return nil, d.err
 	}
 
 	v := d.decodeValue(data[0], data[1:])
 	if d.err != nil {
+		if d.mysqlTextMode && d.ignoreDecodeErr {
```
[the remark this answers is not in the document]
morgo: Applied, same as the sibling thread above (a51b652).

replication/json_binary.go
```diff
@@ -459,20 +485,47 @@ func (d *jsonBinaryDecoder) decodeOpaque(data []byte) any {
 		return d.decodeDecimal(data)
 	case mysql.MYSQL_TYPE_TIME:
 		return d.decodeTime(data)
-	case mysql.MYSQL_TYPE_DATE, mysql.MYSQL_TYPE_DATETIME, mysql.MYSQL_TYPE_TIMESTAMP:
-		return d.decodeDateTime(data)
+	case mysql.MYSQL_TYPE_DATE:
+		// Historically dates have been decoded the same as datetime (including the time portion).
+		// This is mostly harmless, but in text-mode we want to ensure that
+		// the time portion is emitted.
```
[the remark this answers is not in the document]
morgo: Right, the comment was inverted — fixed to say "omitted" (a51b652).

lance6716:
I found two issues worth addressing before merge:

1. `NEWDECIMAL` round-tripping through JSON text does not actually preserve the original JSONB type tag.

The new comments say that `RenderJSONAsMySQLText` preserves each JSONB value's original type tag and that re-inserting the rendered text into a MySQL JSON column reproduces the original JSONB binary. That is true for the `DOUBLE 1.0` case, but it does not hold for `JSONB_OPAQUE/MYSQL_TYPE_NEWDECIMAL`.

The renderer currently emits a decimal opaque value as an unquoted JSON number, e.g. `1.0`. MySQL's JSON text parser handles that through RapidJSON's numeric callbacks and creates a JSON double/int/uint value, not a `Json_decimal`. `Json_decimal` is what later serializes back to `JSONB_OPAQUE` with `MYSQL_TYPE_NEWDECIMAL`. So this avoids turning the value into a JSON string, but it still changes the stored JSONB type tag/binary representation for decimal opaque values.

I think the implementation or the public documentation should be adjusted here. At minimum, the current wording in `BinlogSyncerConfig.RenderJSONAsMySQLText` and `json_mysql_text.go` should not promise original JSONB binary reproduction for `NEWDECIMAL`.

Relevant MySQL paths:
- JSON text parse uses RapidJSON default number handling: `json_dom.cc` `Json_dom::parse`.
- Numeric text values become `Json_double` / `Json_int` / `Json_uint` through the parser callbacks.
- Only `Json_decimal` serializes as opaque `MYSQL_TYPE_NEWDECIMAL` in `json_binary.cc`.

2. `RenderJSONAsMySQLText` is not propagated into decoded transaction payload events.

`BinlogParser.newRowsEvent` copies `p.renderJSONAsMySQLText` into ordinary `RowsEvent`s, but `TRANSACTION_PAYLOAD_EVENT` decoding creates a fresh parser inside `TransactionPayloadEvent.decodePayload()`. That inner parser only receives the copied format description with checksum disabled; it does not receive parser options such as `renderJSONAsMySQLText`.

As a result, when MySQL binlog transaction compression is enabled, rows decoded from the compressed payload silently fall back to the legacy JSON rendering path even if the outer `BinlogSyncerConfig.RenderJSONAsMySQLText` flag is set. The fix should probably carry the relevant parser options into `TransactionPayloadEvent` and apply them to the inner parser before parsing the decompressed events.

Disclosure: This review comment was generated with OpenAI GPT-5 via Codex.

morgo:
> 1. `NEWDECIMAL` round-tripping through JSON text does not actually preserve the original JSONB type tag.
..
> I think the implementation or the public documentation should be adjusted here. At minimum, the current wording in `BinlogSyncerConfig.RenderJSONAsMySQLText` and `json_mysql_text.go` should not promise original JSONB binary reproduction for `NEWDECIMAL`.

The documentation has been updated for this.

> 2. `RenderJSONAsMySQLText` is not propagated into decoded transaction payload events.

Support for this has been added.

replication/json_binary.go
```diff
@@ -459,20 +488,47 @@ func (d *jsonBinaryDecoder) decodeOpaque(data []byte) any {
...
+	case mysql.MYSQL_TYPE_DATE:
+		// Historically dates have been decoded the same as datetime (including the time portion).
+		// This is mostly harmless, but in text-mode we want to ensure that
+		// the time portion is omitted.
+		return d.decodeDateTime(data, d.mysqlTextMode)
+	case mysql.MYSQL_TYPE_DATETIME, mysql.MYSQL_TYPE_TIMESTAMP:
+		return d.decodeDateTime(data, false)
 	default:
+		if d.mysqlTextMode {
+			return "base64:type" + strconv.Itoa(int(tp)) + ":" + base64.StdEncoding.EncodeToString(data)
```
lance6716, line 500: Better to add a unit test to guard all known types should / should not use this fallback way (I'll check if we already have it)

replication/json_mysql_text.go
```diff
@@ -0,0 +1,157 @@
...
+//     covered here do reproduce the original JSONB binary on re-insert.
+
+// jsonString carries a JSONB string payload as raw bytes so MarshalJSON
+// can pass non-ASCII bytes through verbatim. MySQL JSON is byte-
+// transparent, so bytes >= 0x20 (other than '"' and '\\') are written
+// without UTF-8 validation -- unlike the default encoding/json path which
+// replaces invalid UTF-8 with U+FFFD.
+type jsonString string
+
+func (s jsonString) MarshalJSON() ([]byte, error) {
```
lance6716, line 37: please check `"` escaping (`writeJSONString`?)

replication/json_mysql_text.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			continue
+		}
+		buf.WriteByte(c)
+	}
+}
+
+// renderJSONAsMySQLText is a thin entry point used by tests. Production
+// code reaches this path through RowsEvent.decodeJSONBinary.
+func renderJSONAsMySQLText(data []byte, ignoreDecodeErr bool) ([]byte, error) {
```
lance6716: seems we can move this definition to `_test.go` files

lance6716 (commented):
some human review. Will review hours later

replication/json_binary.go
morgo: I've added TestRenderJSONAsMySQLText_OpaqueDispatch for this.

replication/json_mysql_text.go
morgo: I've added TestRenderJSONAsMySQLText_StringEscaping for this.

replication/transaction_payload_event_test.go
```diff
@@ -2,10 +2,64 @@ package replication
...
+	require.False(t, inner.verifyChecksum)
+	// tables must be a fresh map, not shared with the outer parser.
+	require.NotNil(t, inner.tables)
+	require.NotSame(t, &outer.tables, &inner.tables)
+}
+
+// TestNewTransactionPayloadEvent_InheritsParent verifies that events
+// created via the parser path carry a reference to it, so that
+// decodePayload picks up the parser options.
+func TestNewTransactionPayloadEvent_InheritsParent(t *testing.T) {
```
lance6716: please update the function naming style like `TestNewTransactionPayloadEventInheritsParent`

replication/json_mysql_text_test.go
```diff
@@ -0,0 +1,574 @@
...
+	"math"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/stretchr/testify/require"
+)
+
+// renderJSONAsMySQLText is a thin entry point used by tests. Production
```
lance6716: Production code will not reach it?

lance6716 (commented):
rest lgtm

replication/transaction_payload_event_test.go
morgo: Renamed to `TestNewTransactionPayloadEventInheritsParent` (e1e9ce5).

replication/json_mysql_text_test.go
morgo: Correct — production code never calls this helper; it goes through `BinlogParser.newRowsEvent` → `RowsEvent.decodeJSONBinary`. The helper just synthesises a minimal `RowsEvent` so unit tests can drive `jsonBinaryDecoder` directly with `mysqlTextMode` on. Reworded the comment to make that explicit (e1e9ce5).

replication/transaction_payload_event_test.go
lance6716: Oh, sorry I mean all new unit tests in this PR should be renamed

morgo: 🤦 Makes sense, fixing the others.

lance6716 (approved):

PR #1150, opened 2026-06-11
compress: return Reset error from pooled zlib reader

dtunikov:
## Problem

`GetPooledZlibReader` swallows the error from the pooled reader's `Reset` call and returns `(nil, nil)`:

```go
if rc.(zlib.Resetter).Reset(src, nil) != nil {
    return nil, err // err is always nil at this point
}
```

`err` is still `nil` here (it is only assigned in the `else` branch that calls `zlib.NewReader`). So when `Reset` fails — e.g. the source does not begin with a valid zlib header — the caller receives a **nil `io.ReadCloser` together with a nil error**.

That return value is not benign downstream. In `packet.Conn.newCompressedPacketReader` the result is stored directly as the active compressed reader:

```go
c.compressedReader, err = c.newCompressedPacketReader()
```

A `nil` reader with a `nil` error means `c.compressedReader == nil`, and `currentPacketReader()` then falls back to reading the **raw, still-compressed** connection stream as if it were uncompressed:

```go
func (c *Conn) currentPacketReader() io.Reader {
    if c.Compression == MYSQL_COMPRESS_NONE || c.compressedReader == nil {
        return c.reader
    }
    return c.compressedReader
}
```

So a recoverable `Reset` failure is converted into silent, cascading stream corruption that never surfaces to the caller as an error (and `Reset` has already consumed bytes from the underlying stream before failing).

## Fix

Capture and return the actual `Reset` error instead of the always-nil `err`.

## Test

Adds `TestGetPooledZlibReaderResetError`, which primes the pool and then requests a reader over a source with an invalid zlib header, asserting the error is propagated and the returned reader is nil. Verified that it **passes with this fix and fails against the current code** (`got nil`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/compress/zlib.go
+++ b/compress/zlib.go
@@ -57,7 +57,7 @@ func GetPooledZlibReader(src io.Reader) (io.ReadCloser, error) {
 
 	if r := zlibReaderPool.Get(); r != nil {
 		rc = r.(io.ReadCloser)
-		if rc.(zlib.Resetter).Reset(src, nil) != nil {
+		if err = rc.(zlib.Resetter).Reset(src, nil); err != nil {
 			return nil, err
 		}
 	} else {
--- a/compress/zlib_test.go
+++ b/compress/zlib_test.go
@@ -1,0 +1,49 @@
+package compress
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


lance6716 (approved):
Thanks!

PR #1151, opened 2026-06-12
packet: bound uncompressed compressed-protocol frames to the frame length

dtunikov:
## Problem

In the compressed protocol, when a frame stores its payload **verbatim** (length-before-compression `== 0`, which the server uses for small or incompressible chunks), `newCompressedPacketReader` returns a `nil` reader:

```go
if uncompressedLength > 0 {
    limitedReader := io.LimitReader(c.reader, int64(compressedLength))
    switch c.Compression { ... }
}
return nil, nil   // <- uncompressedLength == 0
```

With `compressedReader == nil`, `currentPacketReader()` falls back to the **raw, unbounded** `c.reader`. So `copyN` no longer stops at the frame boundary.

If a MySQL packet spans *out of* such an uncompressed frame, `copyN` reads straight through the **following frame's 7-byte header** as if it were payload, desyncing the compressed stream. And because the reads come from the raw connection (which never returns EOF at the boundary), the `EOF → advance frame` path that increments `CompressedSequence` never runs.

The corruption typically surfaces later, on an unrelated subsequent read, as:

- `invalid compressed sequence X != Y`, or
- `zlib: invalid header`

This is hard to reproduce in the wild because it only triggers when a packet happens to straddle the boundary of a stored-uncompressed frame that is followed by more frames — but under heavy traffic it does happen.

## Fix

Bound the uncompressed payload to the frame with an `io.LimitReader`, mirroring the compressed path, so `copyN` hits EOF at the frame boundary and advances to the next compressed packet.

## Tests

The compressed read path had no unit coverage, so this adds `packet/conn_test.go`:

- `TestReadPacketPacketSpanningUncompressedFrames` — regression test for this bug. It **fails without the fix** (the reassembled payload contains the next frame's header bytes verbatim) and passes with it.
- `TestReadPacketPacketSpanningZlibFrames` — guards the already-working compressed multi-frame path.
- `TestReadPacketSingleUncompressedFrame` — guards the common single-frame small-response case.

### DIFF
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -50,8 +50,6 @@ type Conn struct {
 	compressedHeader [7]byte
 
 	compressedReader io.Reader
-
-	compressedReaderActive bool
 }
 
 func NewConn(conn net.Conn) *Conn {
@@ -107,19 +105,10 @@ func (c *Conn) ReadPacketReuseMem(dst []byte) ([]byte, error) {
 	}()
 
 	if c.Compression != mysql.MYSQL_COMPRESS_NONE {
-		// it's possible that we're using compression but the server response with a compressed
-		// packet with uncompressed length of 0. In this case we leave compressedReader nil. The
-		// compressedReaderActive flag is important to track the state of the reader, allowing
-		// for the compressedReader to be reset after a packet write. Without this flag, when a
-		// compressed packet with uncompressed length of 0 is read, the compressedReader would
-		// be nil, and we'd incorrectly attempt to read the next packet as compressed.
-		if !c.compressedReaderActive {
-			var err error
-			c.compressedReader, err = c.newCompressedPacketReader()
-			if err != nil {
-				return nil, err
-			}
-			c.compressedReaderActive = true
+		var err error
+		c.compressedReader, err = c.newCompressedPacketReader()
+		if err != nil {
+			return nil, err
 		}
 	}
 
@@ -336,7 +325,6 @@ func (c *Conn) WritePacket(data []byte) error {
 			return errors.Wrapf(mysql.ErrBadConn, "Write failed. only %v bytes written, while %v expected", n, len(data))
 		}
 
-		c.compressedReaderActive = false
 		if c.compressedReader != nil {
 			if _, ok := c.compressedReader.(io.ReadCloser); ok {
 				_ = c.compressedReader.(io.ReadCloser).Close()
--- a/packet/conn.go
+++ b/packet/conn.go
@@ -168,8 +168,12 @@ func (c *Conn) newCompressedPacketReader() (io.Reader, error) {
 
 	compressedLength := int(uint32(c.compressedHeader[0]) | uint32(c.compressedHeader[1])<<8 | uint32(c.compressedHeader[2])<<16)
 	uncompressedLength := int(uint32(c.compressedHeader[4]) | uint32(c.compressedHeader[5])<<8 | uint32(c.compressedHeader[6])<<16)
+
+	// Always bound reads to this frame's payload (compressedLength bytes on the wire).
+	// copyN relies on hitting EOF at the frame boundary to advance CompressedSequence
+	// and move on to the next compressed packet.
+	limitedReader := io.LimitReader(c.reader, int64(compressedLength))
 	if uncompressedLength > 0 {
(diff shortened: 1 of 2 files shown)


dveeden:
### The problem, as a diagram

```mermaid
flowchart TD
    A["newCompressedPacketReader():<br/>read 7-byte frame header →<br/>compressedLength, CompressedSequence,<br/>uncompressedLength"]
    A --> C{"length-before-compression<br/>(uncompressedLength)"}

    C -->|"&gt; 0<br/>payload is zlib/zstd"| D["reader = zlib over<br/>io.LimitReader(conn, compressedLength)<br/>— bounded to the frame"]
    C -->|"== 0<br/>payload stored verbatim"| G["return nil, nil →<br/>compressedReader = nil"]

    D --> E["packet spans out of the frame:<br/>copyN hits EOF at the frame boundary"]
    E --> F["CompressedSequence++ →<br/>read next frame's header →<br/>packet continues in next frame ✓"]

    G --> H["currentPacketReader() falls back<br/>to the raw, unbounded c.reader"]
    H --> I["packet spans out of the frame:<br/>no EOF — copyN reads the next frame's<br/>7-byte header as packet payload"]
    I --> J["payload corrupted,<br/>CompressedSequence never advanced,<br/>stream position desynced"]
    J --> K["surfaces later, on an unrelated read:<br/>'invalid compressed sequence X != Y'<br/>or 'zlib: invalid header'"]

    FIX["fix (this PR): bound verbatim frames too —<br/>reader = io.LimitReader(conn, compressedLength)"]
    FIX -.-> G

    classDef neutral fill:#e7f5ff,stroke:#1971c2,color:#111
    classDef good fill:#d3f9d8,stroke:#2f9e44,color:#111
    classDef bad fill:#ffe3e3,stroke:#e03131,color:#111
    classDef fix fill:#fff9db,stroke:#f08c00,color:#111
    class A,C neutral
    class D,E,F good
    class G,H,I,J,K bad
    class FIX fix
```

On the wire (the exact layout of the regression test — a 260-byte MySQL packet split
at byte 100 across two verbatim-stored frames):

```text
┌─ frame 1 header ─┬─ frame 1 payload ────────┬─ frame 2 header ─┬─ frame 2 payload ──┐
│ 7 B              │ 100 B, stored verbatim   │ 7 B              │ 160 B, verbatim    │
│ len=100, seq=0   │  pkt header (4 B)        │ len=160, seq=1   │  body[96:256]      │
│ uncompLen=0      │  + body[0:96]            │ uncompLen=0      │                    │
└──────────────────┴──────────────────────────┴──────────────────┴────────────────────┘
                    ◄──── pkt[0:100] ────────►  not packet data!  ◄── pkt[100:260] ──►

copyN consuming the 256-byte body, BEFORE the fix (reader = raw conn, no frame bound):

  consumed as packet body (256 B):                          left behind in the stream:
  ┌────────────┬──────────────────────┬───────────────┐     ┌───────────────┐
  │ body[0:96] │ frame-2 header (7 B) │ body[96:249]  │     │ body[249:256] │ …
  └────────────┴──────────────────────┴───────────────┘     └───────────────┘
   from frame 1  ▲ 7 stray header       from frame 2          next frame-header read
                   bytes spliced into                         starts here, mid-payload
                   the payload                                → desync

AFTER the fix (reader = io.LimitReader(conn, 100), bounded to the frame):

  body[0:96] → EOF at frame boundary → CompressedSequence++ →
  frame-2 header consumed as a header → body[96:256]   ✓ packet intact, stream in sync
```

---

🤖 *This comment was AI-generated with [Claude Code](https://claude.com/claude-code).*

dveeden (approved):
This looks fine on first glance.

So this happens when the protocol switches on compression and there are packets that span multiple compressed packets but that are not actually using a compressed payload?

lance6716 (commented):
Seems we wont' return nil reader, and the nilness of reader has the same value of caller's compressedReaderActive. So Conn.compressedReaderActive can be removed

dtunikov:
> Seems we wont' return nil reader, and the nilness of reader has the same value of caller's compressedReaderActive. So Conn.compressedReaderActive can be removed

makes sense, removed compressedReaderActive flag
can we merge? or is there anything else i should look into?

lance6716 (approved):

PR #1152, opened 2026-06-12
feat(cmd): support gtid subscription

fansehep:
add cmd `gtid` flags to support go-canal `gtid` subscription.

### DIFF
--- a/cmd/go-canal/main.go
+++ b/cmd/go-canal/main.go
@@ -35,6 +35,8 @@ var (
 	startName = flag.String("bin_name", "", "start sync from binlog name")
 	startPos  = flag.Uint("bin_pos", 0, "start sync from binlog position of")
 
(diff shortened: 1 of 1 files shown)


fansehep:
@lance6716 Can you take a look?

cmd/go-canal/main.go
```diff
@@ -86,13 +88,23 @@ func main() {
...
-
 	go func() {
-		err = c.RunFrom(startPos)
+		var err error
+		if len(*gtid) > 0 {
+			var gset mysql.GTIDSet
+			gset, err = mysql.ParseGTIDSet(*flavor, *gtid)
+			if err != nil {
+				fmt.Printf("parse GTID set err %v\n", err)
+				return
```
lance6716: we should check the gtid at main goroutine? If error happens here, this goroutine will exit, and the main goroutine still blocks.

fansehep: +1

lance6716 (approved):

PR #1153, opened 2026-06-12
fix: make compressed transaction payloads checkpointable

serprex:
binlog_transaction_compression=ON wraps each transaction's Query/TableMap/Rows/XID events in TRANSACTION_PAYLOAD_EVENT. Inner events are copied from transaction cache before position fixup, arriving with LogPos=0, & GTID set attachment only happened on outer XID/Query events. Consequence: canal persisted Pos=0 after every compressed transaction & never called UpdateGTIDSet during streaming since inner XIDEvent.GSet was always nil

- parser: stamp inner event positions after payload decode, payload start for mid-transaction events (resume replays whole transaction, matching uncompressed semantics), payload end for final XID/COMMIT (resume skips transaction)
- binlogsyncer: attach current GTID set to XID/Query events decoded from payload, same as uncompressed counterparts; safe since GTID event precedes payload uncompressed

### DIFF
--- a/canal/sync.go
+++ b/canal/sync.go
@@ -145,10 +145,13 @@ func (c *Canal) handleEvent(ev *replication.BinlogEvent) error {
 			c.cfg.Logger.Error("error parsing query, will skip this event", slog.String("query", string(e.Query)), slog.Any("error", err))
 			return nil
 		}
-		if len(stmts) > 0 {
-			savePos = true
-		}
 		for _, stmt := range stmts {
+			switch stmt.(type) {
+			case *ast.BeginStmt, *ast.SavepointStmt:
+				// transaction not yet complete; checkpointing here would skip it on GTID resume
+				continue
+			}
+			savePos = true
 			nodes := parseStmt(stmt)
 			for _, node := range nodes {
 				if node.db == "" {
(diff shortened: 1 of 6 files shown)


lance6716:
My codex has one comment. I'll do a human review later

---

Review result: one blocking issue.

**[P1] `BEGIN` inside compressed payload can checkpoint GTID too early**

In `replication/binlogsyncer.go:1020`, the new code attaches the current GTID set to every inner `QueryEvent` in a `TransactionPayloadEvent`. Compressed row transactions commonly include an inner `QueryEvent` for `BEGIN` before the row events and final `XID`.

`canal.handleEvent` treats parsable `QueryEvent`s as saveable and updates the GTID set when `GSet` is non-nil. That means `BEGIN` can persist the transaction GTID before the rows and `XID` are processed. If the process crashes after handling `BEGIN` but before the inner `XID`, restart from that GTID can skip the transaction.

Fix direction: only attach `GSet` to terminal transaction events, or make canal ignore `BEGIN`/transaction-control-only `QueryEvent`s for GTID checkpointing.

serprex:
@lance6716 that appears to be an existing issue in uncompressed too, added commit with fix for canal, seems like you already found this in #1144

lance6716 (approved):

PR #1155, opened 2026-06-18
server: advertise CLIENT_MULTI_RESULTS, CLIENT_PS_MULTI_RESULTS in default capabilities

### LINKED ISSUE #1154: server: default capability flags missing CLIENT_MULTI_RESULTS, CLIENT_PS_MULTI_RESULTS, CLIENT_SESSION_TRACK
## Problem

The go-mysql `server` package is commonly used to build MySQL proxies and middleware. During the connection handshake, the server advertises its capability flags; the client then negotiates capabilities as the intersection of what both sides support (`client_caps &= server_caps` in `client/auth.go`).

`NewDefaultServer` and `NewServerWithAuth` currently omit three capability flags that a real MySQL 8.0 server advertises:

| Flag | What it enables |
|------|-----------------|
| `CLIENT_MULTI_RESULTS` | Multiple result sets (required for `CALL` stored procedures) |
| `CLIENT_PS_MULTI_RESULTS` | Multi-result prepared statements and OUT parameters |
| `CLIENT_SESSION_TRACK` | Session-state-change data in OK packets |

### Stored procedure `CALL` fails with Error 1312

Per the [MySQL stored-program protocol](https://dev.mysql.com/doc/dev/mysql-server/8.0.46/page_protocol_command_phase_sp.html):

> The client has to announce that it wants multi-resultsets by either setting the CLIENT_MULTI_RESULTS or CLIENT_PS_MULTI_RESULTS capability flags.

When a go-mysql server acts as a proxy frontend, clients cannot negotiate `CLIENT_MULTI_RESULTS` because the server does not advertise it. If the proxy forwards `CALL sp_that_returns_resultsets()` to a real MySQL upstream, MySQL returns:

```
ERROR 1312 (0A000): PROCEDURE sp_that_returns_resultsets can't return a result set in the given context
```

This affects JDBC Connector/J (`CallableStatement`), the C API, and any client that relies on stored procedures.

### Session tracking silently lost

MySQL 8.0+ can include session-state-change metadata (e.g. `gtid_executed`, charset changes) in OK packets when `CLIENT_SESSION_TRACK` is negotiated. Without advertising this flag, proxies cannot transparently relay that metadata.

## Reproducer

See the PR for a self-contained Go test (`TestNegotiatedMultiResultsCapability`) that starts a go-mysql server, connects a go-mysql client, and shows that `CLIENT_MULTI_RESULTS` is not negotiated without this fix.

With a real MySQL 8.0 backend:

```sql
CREATE PROCEDURE test_multi()
BEGIN
    SELECT 1 AS a;
    SELECT 2 AS b;
END;
```

```bash
# Through go-mysql proxy (before fix):
mysql -h 127.0.0.1 -P <proxy_port> -u root -e "CALL test_multi()"
# ERROR 1312 (0A000): PROCEDURE test.test_multi can't return a result set in the given context

# Direct to MySQL (works):
mysql -h 127.0.0.1 -P 3306 -u root -e "CALL test_multi()"
```

## Proposed fix

Add `CLIENT_MULTI_RESULTS`, `CLIENT_PS_MULTI_RESULTS`, and `CLIENT_SESSION_TRACK` to the default capability bitmask in `NewDefaultServer` and `NewServerWithAuth`. This matches what MySQL 8.0 advertises and only enables negotiation — it does not change wire-format handling.

nzin-alayacare:
## Summary

Advertise `CLIENT_MULTI_RESULTS`, `CLIENT_PS_MULTI_RESULTS`, and `CLIENT_SESSION_TRACK` in the default server capability flags for `NewDefaultServer` and `NewServerWithAuth`.

Closes #1154

## Why

The go-mysql `server` package is commonly used to build MySQL proxies. During handshake, clients negotiate capabilities as the intersection of server-advertised and client-requested flags (`c.capability &= capability` in `client/auth.go`).

Without these flags, clients like JDBC Connector/J cannot negotiate multi-result support. When a proxy forwards `CALL sp_that_returns_resultsets()` to a real MySQL upstream, MySQL returns:

```
ERROR 1312 (0A000): PROCEDURE sp can't return a result set in the given context
```

See [MySQL stored-program protocol](https://dev.mysql.com/doc/dev/mysql-server/8.0.46/page_protocol_command_phase_sp.html).

`CLIENT_SESSION_TRACK` is also advertised by MySQL 8.0 and required for proxies to relay session-state-change metadata in OK packets.

## What changed

- Add three capability flags to `NewDefaultServer` and `NewServerWithAuth` in `server/server_conf.go`
- Add `server/server_conf_test.go` with:
  - `TestDefaultServerCapabilities` / `TestNewServerCapabilities` — verify flags are advertised
  - `TestNegotiatedMultiResultsCapability` — reproduces the stored-procedure CALL scenario: server + client handshake, verifies `CLIENT_MULTI_RESULTS` is negotiated, executes `CALL get_report()`

## Reproducer

`TestNegotiatedMultiResultsCapability` is a self-contained test (no external MySQL required) that:

1. Starts a go-mysql server with a handler simulating `CALL get_report()`
2. Connects a client requesting `CLIENT_MULTI_RESULTS` (like JDBC)
3. Verifies the flag is negotiated and the CALL returns a result set

With a real MySQL 8.0 backend, the same missing flag causes Error 1312 when routing `CALL` through an unpatched go-mysql proxy.

## Test plan

- [x] `go test ./server -run 'TestDefaultServerCapabilities|TestNewServerCapabilities|TestNegotiatedMultiResultsCapability'`

### DIFF
--- a/server/server_conf.go
+++ b/server/server_conf.go
@@ -55,7 +55,7 @@ func NewDefaultServer() *Server {
 		capability: mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_LONG_FLAG | mysql.CLIENT_CONNECT_WITH_DB | mysql.CLIENT_PROTOCOL_41 |
 			mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_SECURE_CONNECTION | mysql.CLIENT_PLUGIN_AUTH | mysql.CLIENT_SSL |
 			mysql.CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA | mysql.CLIENT_CONNECT_ATTRS |
-			mysql.CLIENT_MULTI_RESULTS | mysql.CLIENT_PS_MULTI_RESULTS | mysql.CLIENT_SESSION_TRACK,
+			mysql.CLIENT_MULTI_RESULTS | mysql.CLIENT_PS_MULTI_RESULTS,
 		collationID:       mysql.DEFAULT_COLLATION_ID,
 		defaultAuthMethod: mysql.AUTH_NATIVE_PASSWORD,
 		rsaPrivateKey:     rsaPrivateKey,
@@ -102,7 +102,7 @@ func NewServerWithAuth(serverVersion string, collationID uint8, defaultAuthMetho
 	capFlag := mysql.CLIENT_LONG_PASSWORD | mysql.CLIENT_LONG_FLAG | mysql.CLIENT_CONNECT_WITH_DB | mysql.CLIENT_PROTOCOL_41 |
 		mysql.CLIENT_TRANSACTIONS | mysql.CLIENT_SECURE_CONNECTION | mysql.CLIENT_PLUGIN_AUTH | mysql.CLIENT_CONNECT_ATTRS |
 		mysql.CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA |
-		mysql.CLIENT_MULTI_RESULTS | mysql.CLIENT_PS_MULTI_RESULTS | mysql.CLIENT_SESSION_TRACK
+		mysql.CLIENT_MULTI_RESULTS | mysql.CLIENT_PS_MULTI_RESULTS
 	if tlsConfig != nil {
 		capFlag |= mysql.CLIENT_SSL
 	}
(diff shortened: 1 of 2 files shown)


nzin-alayacare:
If needed, I can also bring a full java client <-> proxy <-> mysql8 example code (with the current issue).
Let me know!

lance6716:
Hi, I have triggerd CI for this PR. Please fix linter error

nzin-alayacare:
> Hi, I have triggerd CI for this PR. Please fix linter error

Hi. @lance6716 , I guess you need to restart the CI? :-(

lance6716:
This library didn't implement CLIENT_SESSION_TRACK, so I think we should not turn it on by default. Is it necessary in your use case?

nzin-alayacare:
> This library didn't implement CLIENT_SESSION_TRACK, so I think we should not turn it on by default. Is it necessary in your use case?

True. I can try to create a separate follow-up PR: add session tracking support (writeOK serialization + AppendOKSessionTrackSuffix), then advertise CLIENT_SESSION_TRACK alongside that implementation.

-> https://github.com/go-mysql-org/go-mysql/pull/1158

lance6716 (approved):

PR #1158, opened 2026-06-21
Add server-side session tracking support

### LINKED ISSUE #1159: Server does not emit session tracking data in OK packets
## Problem

The go-mysql **server** does not support [MySQL session state tracking](https://dev.mysql.com/doc/internals/en/session-state-change-packets.html). When a client negotiates the `CLIENT_SESSION_TRACK` capability during the handshake, the server should append a session-tracking suffix to OK packets whenever the `SERVER_SESSION_STATE_CHANGED` status flag is set.

Without this support:

- Clients such as **mysql-connector-j** cannot observe session changes (schema switches, system variable updates, GTIDs, transaction state, etc.) from OK responses.
- Applications that rely on `trackSessionState` or equivalent driver APIs receive no `SESSION_TRACK_*` data from a go-mysql server.
- The server does not advertise `CLIENT_SESSION_TRACK` in its default capability flags, so capable clients may not even attempt to use the feature.

### Expected behavior

When `CLIENT_SESSION_TRACK` is negotiated and a result sets `SERVER_SESSION_STATE_CHANGED`, the OK packet should include a session-tracking block containing one or more `SESSION_TRACK_*` entries:

| Type | Purpose |
|------|---------|
| `SESSION_TRACK_SYSTEM_VARIABLES` | Changed system variables (name/value pairs) |
| `SESSION_TRACK_SCHEMA` | Current default database |
| `SESSION_TRACK_STATE_CHANGE` | Session state changed flag (`"1"` / `"0"`) |
| `SESSION_TRACK_GTIDS` | GTID set for the session |
| `SESSION_TRACK_TRANSACTION_CHARACTERISTICS` | Transaction characteristics |
| `SESSION_TRACK_TRANSACTION_STATE` | Transaction state |

Payload lengths must use **length-encoded integers** per the MySQL protocol. In particular, `SESSION_TRACK_STATE_CHANGE` carries a **single raw byte** (`0x31` for `"1"`) inside the entry wrapper—not a nested length-encoded string.

### Current behavior

The server writes a plain OK packet with no session-tracking suffix, regardless of client capabilities or result flags.

## Proposed solution

PR #1158 adds server-side session tracking support:

- Serialize all `SESSION_TRACK_*` payload types via `AppendOKSessionTrackSuffix` in the `mysql` package.
- Extend `writeOK` to emit the suffix when `CLIENT_SESSION_TRACK` is negotiated and `SERVER_SESSION_STATE_CHANGED` is set.
- Advertise `CLIENT_SESSION_TRACK` in default server capability flags.
- Align encoder/decoder wire format with MySQL (length-encoded entry lengths; single-byte `STATE_CHANGE` payload).

## References

- MySQL internals: [Session State Change Packets](https://dev.mysql.com/doc/internals/en/session-state-change-packets.html)
- Solution: #1158

nzin-alayacare:
Fixes #1159

## Summary
- Add `AppendOKSessionTrackSuffix` and session tracking serialization in the `mysql` package so OK packets can carry all `SESSION_TRACK_*` payload types.
- Extend server `writeOK` to emit the session tracking suffix when `CLIENT_SESSION_TRACK` is negotiated and `SERVER_SESSION_STATE_CHANGED` is set on the result.
- Advertise `CLIENT_SESSION_TRACK` in default server capability flags and add unit tests for encoding and OK packet writing.

## Test plan
- [x] `go test ./mysql/... ./server/...`
- [x] Verify `TestConnWriteOKWithSessionTrack` and `TestEncodeSessionTrackingRoundTrip`


nzin-alayacare:
I can provide a fake component test (docker-compose) if needed

lance6716 (commented):
Codex review. I'll do a human review later

server/resp.go
```diff
@@ -28,6 +28,11 @@ func (c *Conn) writeOK(r *mysql.Result) error {
 		data = append(data, byte(r.Warnings), byte(r.Warnings>>8))
 	}
 
+	if c.capability&mysql.CLIENT_SESSION_TRACK > 0 &&
```
lance6716: [P2] This gates the whole CLIENT_SESSION_TRACK OK suffix on SERVER_SESSION_STATE_CHANGED, but the length-encoded info/status message field belongs to every OK packet once CLIENT_SESSION_TRACK is negotiated; only the session-state-changes field is gated by SERVER_SESSION_STATE_CHANGED. As written, a Result with StatusMessage but no session-state change silently drops the message for clients that requested session tracking. Split the helper or gate only the session-change block.

client/resp.go
```diff
@@ -82,7 +82,8 @@ func decodeSessionTracking(data []byte) (s *mysql.SessionTrackingInfo, err error
 	for pos < len(data) {
 		sessionTrackingChangeType := data[pos]
 		pos++ // session tracking type
-		pos++ // length of session tracking data, unused
+		_, _, n := mysql.LengthEncodedInt(data[pos:])
```
lance6716, line 85: [P2] The entry length is now length-encoded and the new encoder writes nested fields with appendLenEncString, but the type-specific decoder below still reads var/schema/GTID/transaction lengths as a single byte. Values >=251 bytes, especially GTID sets or variable values, will be misdecoded or make the remaining bytes look like another session-track entry. Use mysql.LengthEncodedString for those nested fields and add a long-value test.

nzin-alayacare:
Hi @lance6716 I (think) I addressed Codex comments.

I also did a kind of component testing with a (fake) java app using mysql-connector-j

client/resp.go
lance6716: the definition of `err` should be put closer to its usage.

lance6716: Although you didn't change it, I just restart review for this part. As https://dev.mysql.com/doc/dev/mysql-server/8.4.10/page_protocol_basic_ok_packet.html said, the logic should be

```go
if c.capability&mysql.CLIENT_SESSION_TRACK > 0 {
  // line 53 to line 60, read StatusMessage
  if c.status&mysql.SERVER_SESSION_STATE_CHANGED > 0 {
    // line 62 to line 69, read SessionTracking
  }
}
```

Can you help to recheck it?

client/resp.go
```diff
@@ -52,17 +52,17 @@ func (c *Conn) handleOKPacket(data []byte) (*mysql.Result, error) {
...
+		statusMessageLength, _, n := mysql.LengthEncodedInt(data[pos:])
+		pos += n
 		if statusMessageLength > 0 {
-			r.StatusMessage = utils.ByteSliceToString(data[pos : pos+statusMessageLength])
-			pos += statusMessageLength
+			r.StatusMessage = utils.ByteSliceToString(data[pos : pos+int(statusMessageLength)])
+			pos += int(statusMessageLength)
 		}
 
-		sessionTrackingChangeLength := int(data[pos])
```
lance6716, line 62: I think the old code is correct

https://dev.mysql.com/doc/dev/mysql-server/8.4.10/page_protocol_basic_ok_packet.html#sect_protocol_basic_ok_packet_sessinfo

lance6716 (commented):
rest file may also be updated to align with protocol

nzin-alayacare:
Hi @lance6716, addressed all three comments:

**Comment 1 — `var err error` placement:** moved inside the inner `SERVER_SESSION_STATE_CHANGED` block, right before its use.

**Comment 2 — read order:** restructured `handleOKPacket` to match the protocol spec exactly:
```go
if c.capability&mysql.CLIENT_SESSION_TRACK > 0 {
    // always read info string
    if c.status&mysql.SERVER_SESSION_STATE_CHANGED > 0 {
        // also read session_state_info
    }
}
```
Added a test for the `CLIENT_SESSION_TRACK`-only case (no `SERVER_SESSION_STATE_CHANGED`) to cover the new path.

**Comment 3 — `sessionTrackingChangeLength` using `LengthEncodedInt`:** I kept `LengthEncodedInt` here. The MySQL OK packet spec (https://dev.mysql.com/doc/dev/mysql-server/8.4.10/page_protocol_basic_ok_packet.html) describes `session state info` as `string<lenenc>`, which means the outer block length is a length-encoded integer — not a fixed single byte. Our encoder also writes it with `PutLengthEncodedInt`, so the decoder must match. The single-byte approach (`int(data[pos])`) would silently corrupt packets with a session tracking block ≥ 251 bytes. Happy to discuss if you see it differently.

Commit: `1a88fff`

lance6716:
Comment 3 : You are right. I mistakenly see the layout of `sessionTrackingChangeType`. Reviewing.

lance6716 (approved):

PR #1164, opened 2026-07-02
client: add ExecQueryRelayLocalInfile for proxy-friendly LOAD DATA LOCAL INFILE

### LINKED ISSUE #1161: client: add ExecQueryRelayLocalInfile for proxy-friendly LOAD DATA LOCAL INFILE
## Summary

Add `(*Conn).ExecQueryRelayLocalInfile` to the `client` package — a variant of `Execute` that handles the `LOCAL INFILE` protocol handshake via a callback instead of reading a local file. This enables transparent proxying of `LOAD DATA LOCAL INFILE` without the proxy needing filesystem access.

## Problem

The existing `client.Conn.Execute` follows the standard `LOCAL INFILE` flow defined in `client/auth.go` (`handleAuthResult` / local file reading). This works when the go-mysql client is the origin of the file, but **not** when the client is a proxy that must relay the file from another party (e.g., an end-user client connected to the proxy's server side).

There is currently no API to intercept the `0xfb LOCAL INFILE` response from MySQL and hand off the remainder of the exchange to caller-supplied logic.

## Proposed API

```go
// ExecQueryRelayLocalInfile sends COM_QUERY and handles the LOCAL INFILE protocol
// when the server responds with a 0xfb packet. relayFile receives the full 0xfb
// request payload; the implementation must forward the request to the application
// client, read file-data packets from that client, and write each to this Conn via
// WritePacket (4-byte header + payload) until an empty packet signals EOF.
// The final OK or ERR from the upstream server is read and returned.
func (c *Conn) ExecQueryRelayLocalInfile(
    query string,
    relayFile func(localInfileRequestPayload []byte) error,
) (*mysql.Result, error)
```

A companion helper is needed for protocol correctness when `relayFile` returns an error mid-transfer — the connection must be re-synchronized before it can be reused:

```go
// cleanupLocalInfileAfterRelayError sends an empty payload (EOF signal) and reads
// the server's OK/ERR to resynchronize the connection state.
func (c *Conn) cleanupLocalInfileAfterRelayError() error
```

## Implementation Sketch

```go
func (c *Conn) cleanupLocalInfileAfterRelayError() error {
    if err := c.WritePacket(make([]byte, 4)); err != nil {
        return err
    }
    resp, err := c.ReadPacket()
    if err != nil {
        return err
    }
    switch resp[0] {
    case mysql.OK_HEADER:
        _, err := c.handleOKPacket(resp)
        return err
    case mysql.ERR_HEADER:
        _ = c.handleErrorPacket(resp)
        return nil
    default:
        return errors.Errorf("unexpected packet during local infile cleanup: 0x%x", resp[0])
    }
}

func (c *Conn) ExecQueryRelayLocalInfile(query string, relayFile func([]byte) error) (*mysql.Result, error) {
    if err := c.execSend(query); err != nil {
        return nil, errors.Trace(err)
    }
    data, err := c.ReadPacket()
    if err != nil {
        return nil, errors.Trace(err)
    }
    switch data[0] {
    case mysql.OK_HEADER:
        return c.handleOKPacket(data)
    case mysql.ERR_HEADER:
        return nil, c.handleErrorPacket(data)
    case mysql.LocalInFile_HEADER:
        if err := relayFile(data); err != nil {
            if cleanupErr := c.cleanupLocalInfileAfterRelayError(); cleanupErr != nil {
                return nil, errors.Errorf("relay failed: %v; cleanup failed: %v", err, cleanupErr)
            }
            return nil, errors.Trace(err)
        }
        data2, err := c.ReadPacket()
        if err != nil {
            return nil, errors.Trace(err)
        }
        switch data2[0] {
        case mysql.OK_HEADER:
            return c.handleOKPacket(data2)
        case mysql.ERR_HEADER:
            return nil, c.handleErrorPacket(data2)
        default:
            return nil, errors.Errorf("unexpected packet after LOCAL INFILE: 0x%x", data2[0])
        }
    default:
        return nil, errors.Errorf("unexpected response to COM_QUERY: 0x%x", data[0])
    }
}
```

## Use Case

A database jump-host proxy (like the one we are developing internally) sits between a MySQL client (DBeaver, JDBC) and the real MySQL server. When the client issues `LOAD DATA LOCAL INFILE`:

1. The proxy's server-side connection (`server.Conn`) receives the query.
2. The proxy calls `targetConn.ExecQueryRelayLocalInfile(query, func(req []byte) error { ... })`.
3. Inside the callback, the proxy forwards the `0xfb` request to its server-side client, then pipes file-data packets in both directions until the empty EOF packet.
4. The final result (OK or ERR) is returned and forwarded to the original client.

## Notes

- Pure additive change — no breaking API changes.
- This complements issue #1160 (`CLIENT_LOCAL_FILES` server capability) which is needed so clients connecting to the proxy-as-server negotiate the flag in the first place.
- Tested in production at alayacare/castledm.

nzin-alayacare:
## Summary

Add `(*Conn).ExecQueryRelayLocalInfile` and the companion helper `cleanupLocalInfileAfterRelayError` to `client/conn.go`.

Closes #1161

## Motivation

The existing `Execute` / `exec` path (and the LOCAL INFILE handling in `client/auth.go`) reads the file from the local filesystem. In a proxy scenario the file lives on the *end-user's* machine, not on the proxy. The proxy must relay the `0xfb` request and the subsequent file-data packets between the upstream MySQL server and the downstream client.

There is no API today to intercept the `0xfb LOCAL INFILE` response and hand off the transfer to a callback. This PR fills that gap.

## New API

```go
// ExecQueryRelayLocalInfile sends COM_QUERY and handles the LOCAL INFILE protocol
// when the server responds with a 0xfb packet. relayFile receives the full 0xfb
// request payload; the implementation must forward the request to the application
// client, read file-data packets from that client, and write each to this
// connection via WritePacket (4-byte header + payload) until an empty packet
// signals EOF. The final OK or ERR from the upstream server is read and returned.
//
// If relayFile returns an error the connection is resynchronized before the error
// is returned so the connection can be reused.
func (c *Conn) ExecQueryRelayLocalInfile(
    query string,
    relayFile func(localInfileRequestPayload []byte) error,
) (*mysql.Result, error)
```

`cleanupLocalInfileAfterRelayError` is unexported; it sends the empty EOF packet and reads the server's OK/ERR to bring the connection back to a usable state when `relayFile` fails mid-transfer.

## Notes

- Pure additive change — no existing API is modified.
- Requires `CLIENT_LOCAL_FILES` to be set on both the server side (see #1163) and the client side (via `SetCapability`).
- Tested in production at [alayacare/castledm](https://github.com/alayacare/castledm).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/client/conn.go
+++ b/client/conn.go
@@ -392,38 +392,35 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
 	return c.readResultStreaming(false, result, perRowCallback, perResultCallback)
 }
 
-// cancelLocalInfileRequest sends an empty LOCAL INFILE payload packet and reads the server's
-// OK or ERR so the connection can be reused when relayFile failed before any file data was sent.
-func (c *Conn) cancelLocalInfileRequest() error {
-	data := make([]byte, 4)
-	if err := c.WritePacket(data); err != nil {
-		return errors.Trace(err)
+// prepareLocalInfileReader fully validates that r can be read without error before any data is
+// sent upstream, per the LOCAL INFILE protocol requirement that on error the server must see
+// only the terminal empty packet, never a truncated file. If r implements io.Seeker (as returned
+// by callers per the documented example: *bytes.Reader, *os.File) it is drained and seeked back
+// to the start so large files are not fully buffered in memory; otherwise the content is read
+// fully into memory with io.ReadAll.
+func prepareLocalInfileReader(r io.Reader) (io.Reader, error) {
+	if r == nil {
+		return bytes.NewReader(nil), nil
 	}
-	resp, err := c.ReadPacket()
+	if seeker, ok := r.(io.Seeker); ok {
+		if _, err := io.Copy(io.Discard, r); err != nil {
+			return nil, err
+		}
+		if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+			return nil, err
+		}
+		return r, nil
+	}
+	content, err := io.ReadAll(r)
 	if err != nil {
-		return errors.Trace(err)
+		return nil, err
 	}
-	if len(resp) == 0 {
-		return errors.New("unexpected empty packet during local infile cancel")
-	}
-	switch resp[0] {
-	case mysql.OK_HEADER:
-		_, err := c.handleOKPacket(resp)
-		return errors.Trace(err)
-	case mysql.ERR_HEADER:
-		_ = c.handleErrorPacket(resp)
-		return nil
-	default:
-		return errors.Errorf("unexpected packet during local infile cancel: 0x%x", resp[0])
-	}
+	return bytes.NewReader(content), nil
 }
 
-// sendLocalInfileContent reads file bytes from r and sends them as LOCAL INFILE data packets,
+// streamLocalInfileChunks sends already-validated file bytes as LOCAL INFILE data packets,
 // followed by the terminal empty packet.
-func (c *Conn) sendLocalInfileContent(r io.Reader) error {
-	if r == nil {
-		r = bytes.NewReader(nil)
-	}
+func (c *Conn) streamLocalInfileChunks(r io.Reader) error {
 	buf := make([]byte, defaultBufferSize)
 	for {
 		n, err := r.Read(buf)
@@ -447,13 +444,58 @@ func (c *Conn) sendLocalInfileContent(r io.Reader) error {
 	return nil
 }
 
+// sendLocalInfileContentAndAwaitResult validates and sends LOCAL INFILE content, then reads the
+// server's OK or ERR. When relayErr is non-nil, or when reader validation fails, only the empty
+// terminator packet is sent (no file data).
+func (c *Conn) sendLocalInfileContentAndAwaitResult(reader io.Reader, relayErr error) (*mysql.Result, error) {
+	content := io.Reader(bytes.NewReader(nil))
+	if relayErr == nil {
+		validated, err := prepareLocalInfileReader(reader)
+		if err != nil {
+			relayErr = err
+		} else {
+			content = validated
+		}
+	}
+	if err := c.streamLocalInfileChunks(content); err != nil {
+		return nil, errors.Trace(err)
+	}
+	resp, err := c.ReadPacket()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if len(resp) == 0 {
+		return nil, errors.New("unexpected empty packet after LOCAL INFILE")
+	}
+	var result *mysql.Result
+	var respErr error
+	switch resp[0] {
+	case mysql.OK_HEADER:
+		result, respErr = c.handleOKPacket(resp)
+	case mysql.ERR_HEADER:
+		respErr = c.handleErrorPacket(resp)
+	default:
+		respErr = errors.Errorf("unexpected packet after LOCAL INFILE: 0x%x", resp[0])
+	}
+	if relayErr != nil {
+		if respErr != nil {
+			return nil, errors.Errorf("local infile relay failed: %v; server response: %v", relayErr, respErr)
+		}
+		return nil, errors.Trace(relayErr)
+	}
+	return result, errors.Trace(respErr)
+}
+
 // ExecQueryRelayLocalInfile sends COM_QUERY and handles the LOCAL INFILE protocol when the server
 // responds with a 0xfb packet. relayFile receives that full request payload and must return a
-// reader over the complete file content (or nil for an empty file). The library sends the data
-// packets and terminal empty packet; the final OK or ERR from the upstream server is returned.
+// reader over the complete file content (or nil for an empty file). The library validates the
+// reader, sends data packets and the terminal empty packet, then returns the final OK or ERR.
 //
-// If relayFile returns an error, no file data is sent upstream; cancelLocalInfileRequest sends
-// only the empty packet so the connection can be reused.
+// If relayFile returns an error, or the reader cannot be fully read, no file data is sent
+// upstream; only the empty terminator packet is sent so the connection can be reused.
+//
+// OK and ERR responses before the 0xfb packet are handled directly when the server rejects the
+// LOAD DATA statement (e.g. local_infile disabled, missing privileges).
 //
 // This is the proxy-friendly counterpart to the local-file reading in client/auth.go. It does not
 // access the filesystem; ownership of the file transfer is delegated entirely to the caller.
@@ -497,34 +539,11 @@ func (c *Conn) ExecQueryRelayLocalInfile(query string, relayFile func(localInfil
 	case mysql.ERR_HEADER:
 		return nil, c.handleErrorPacket(data)
 	case mysql.LocalInFile_HEADER:
-		reader, err := relayFile(data)
-		if err != nil {
-			if cancelErr := c.cancelLocalInfileRequest(); cancelErr != nil {
-				return nil, errors.Errorf("local infile relay failed: %v; cancel failed: %v", err, cancelErr)
-			}
-			return nil, errors.Trace(err)
-		}
+		reader, relayErr := relayFile(data)
 		if closer, ok := reader.(io.Closer); ok && reader != nil {
 			defer closer.Close()
 		}
-		if err := c.sendLocalInfileContent(reader); err != nil {
-			return nil, errors.Trace(err)
-		}
-		data2, err := c.ReadPacket()
-		if err != nil {
-			return nil, errors.Trace(err)
-		}
-		if len(data2) == 0 {
-			return nil, errors.New("unexpected empty packet after LOCAL INFILE")
-		}
-		switch data2[0] {
-		case mysql.OK_HEADER:
-			return c.handleOKPacket(data2)
-		case mysql.ERR_HEADER:
-			return nil, c.handleErrorPacket(data2)
-		default:
-			return nil, errors.Errorf("unexpected packet after LOCAL INFILE: 0x%x", data2[0])
-		}
+		return c.sendLocalInfileContentAndAwaitResult(reader, relayErr)
 	default:
 		return nil, errors.Errorf("unexpected response to COM_QUERY (expected OK, ERR, or LOCAL INFILE): 0x%x", data[0])
 	}
--- a/client/conn.go
+++ b/client/conn.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"maps"
 	"math/bits"
 	"net"
@@ -391,9 +392,9 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
 	return c.readResultStreaming(false, result, perRowCallback, perResultCallback)
 }
 
-// cleanupLocalInfileAfterRelayError sends an empty payload packet (EOF for LOCAL INFILE) and reads
-// the server's OK or ERR so the client connection is resynchronized after relayFile failed mid-transfer.
-func (c *Conn) cleanupLocalInfileAfterRelayError() error {
+// cancelLocalInfileRequest sends an empty LOCAL INFILE payload packet and reads the server's
+// OK or ERR so the connection can be reused when relayFile failed before any file data was sent.
+func (c *Conn) cancelLocalInfileRequest() error {
 	data := make([]byte, 4)
 	if err := c.WritePacket(data); err != nil {
 		return errors.Trace(err)
@@ -403,7 +404,7 @@ func (c *Conn) cleanupLocalInfileAfterRelayError() error {
 		return errors.Trace(err)
 	}
 	if len(resp) == 0 {
-		return errors.New("unexpected empty packet during local infile cleanup")
+		return errors.New("unexpected empty packet during local infile cancel")
 	}
 	switch resp[0] {
 	case mysql.OK_HEADER:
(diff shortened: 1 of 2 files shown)


lance6716:
Oh for previous PR codex found some problems:

[P1] server/resp.go (line 31): when CLIENT_SESSION_TRACK is negotiated but SERVER_SESSION_STATE_CHANGED is not set, writeOK still appends a zero-length session-state block via AppendOKSessionTrackSuffix. In the MySQL OK-packet format, that field is only present when the status bit is set; otherwise the packet should stop after the length-encoded info string. This makes auth OK / normal OK packets malformed for strict clients.

[P2] server/server_conf.go (line 137): the new SetCapability / UnsetCapability API lets callers flip any server capability bit, but some of those bits are not independently configurable. A concrete example is CLIENT_SSL: enabling it on a server without tlsConfig will advertise TLS in the greeting and then fail the SSLRequest path during handshake. The setter should either validate a safe whitelist or tie each flag to the underlying server config.

Can you create a new PR to fix it?

nzin-alayacare:
> [P1] server/resp.go (line 31): when CLIENT_SESSION_TRACK is negotiated but SERVER_SESSION_STATE_CHANGED is not set, writeOK still appends a zero-length session-state block via AppendOKSessionTrackSuffix. In the MySQL OK-packet format, that field is only present when the status bit is set; otherwise the packet should stop after the length-encoded info string. This makes auth OK / normal OK packets malformed for strict clients.
> 
> [P2] server/server_conf.go (line 137): the new SetCapability / UnsetCapability API lets callers flip any server capability bit, but some of those bits are not independently configurable. A concrete example is CLIENT_SSL: enabling it on a server without tlsConfig will advertise TLS in the greeting and then fail the SSLRequest path during handshake. The setter should either validate a safe whitelist or tie each flag to the underlying server config.

sure -> https://github.com/go-mysql-org/go-mysql/pull/1167

client/conn.go
```diff
@@ -391,6 +391,86 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+	if len(data) == 0 {
+		return nil, errors.New("unexpected empty packet from server")
+	}
+	switch data[0] {
+	case mysql.OK_HEADER:
+		return c.handleOKPacket(data)
+	case mysql.ERR_HEADER:
+		return nil, c.handleErrorPacket(data)
+	case mysql.LocalInFile_HEADER:
+		if err := relayFile(data); err != nil {
```
lance6716: This function signature does not clearly say how to write file content back. (from unit tests I know we can use receiver.) Maybe change it to `func(localInfileRequestPayload []byte, c *Conn) error`?

client/conn.go
```diff
@@ -391,6 +391,86 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+			if cleanupErr := c.cleanupLocalInfileAfterRelayError(); cleanupErr != nil {
```
lance6716: I'm not sure about the behaivour.

As https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response_local_infile_data.html said

"If the file is empty or there is a error while reading the file only the empty packet is sent."

Does it mean we should not send partial file content?

client/conn.go
nzin-alayacare: Good point — updated in e5a1daf. The callback is now `func(localInfileRequestPayload []byte, c *Conn) error` so callers can write file data back via `c.WritePacket` without relying on closure capture.

client/conn.go
nzin-alayacare: The spec sentence applies to the **originating client** (the party that owns the file), not to forbidding partial relay in a proxy.

**Success path:** `relayFile` must send all file-data packets and the terminal empty packet (4-byte header, zero payload) before returning `nil`. This matches the normal LOCAL INFILE flow.

**Error path:** If `relayFile` fails after forwarding some chunks, we cannot unsend data already written to the upstream connection. What the protocol requires on error is that **only the empty packet** is sent as the terminator — no further data after the failure. `cleanupLocalInfileAfterRelayError` sends that empty packet, consumes the server's OK/ERR, and leaves the connection reusable. The original relay error is still returned to the caller.

So partial content may already be on the wire when an error occurs (unavoidable in a relay), but we always terminate with the empty packet as required. Documented this contract in the godoc in e5a1daf.

lance6716: > On error, relayFile may return after forwarding partial data; the empty packet
> // is then sent by cleanupLocalInfileAfterRelayError so the connection can be resynchronized and
> // reused. 

I don't think this is a good behaivour. The real backend MySQL server will not know the data is not completed. It only sees data and an empty packet which is the same as a completed LOAD DATA.

How about changing the function signature to return an input filename to caller and require a complete file content from caller? In other words

```
relayFile func(localInfileRequestPayload []byte) (fullFileContent []byte, err error)
```

So we constrain the caller to only have one chance to return the full file content, and let this library send the content at protocol level.

nzin-alayacare: Agree that partial data + empty packet is indistinguishable from success on the server, and that the callback should not stream via `WritePacket`.

Our internal proxy solution collects the full file from the downstream JDBC client before the library sends upstream. For files over 1 MB we spill to a local temp file during collection to avoid holding the full payload in proxy memory.

Would `func(localInfileRequestPayload []byte) (io.Reader, error)` work so large files can be returned from disk without a final `[]byte` allocation? The library would read from the reader and send protocol packets only after the callback returns successfully — nothing sent upstream on callback error. If `[]byte` is required we can still adopt it with the temp-file collection path.

Pushed an update on our branch implementing the `io.Reader` variant.

client/conn.go
```diff
@@ -391,6 +392,144 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+	}
+	data, err := c.ReadPacket()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if len(data) == 0 {
+		return nil, errors.New("unexpected empty packet from server")
+	}
+	switch data[0] {
+	case mysql.OK_HEADER:
```
lance6716, line 559: should we only treat `LocalInFile_HEADER` as expected? This function seems a dedicated execution path of LOAD DATA, the query should be a LOAD DATA and `relayFile` is always expected to be called?

client/conn.go
```diff
@@ -391,6 +392,144 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+	if r == nil {
+		r = bytes.NewReader(nil)
+	}
+	buf := make([]byte, defaultBufferSize)
+	for {
+		n, err := r.Read(buf)
+		if n > 0 {
+			pkt := make([]byte, 4+n)
+			copy(pkt[4:], buf[:n])
+			if err := c.WritePacket(pkt); err != nil {
```
lance6716: we still need to io.ReadAll to ensure not sending partial file content?

client/conn.go
```diff
@@ -391,6 +392,144 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
 	return c.readResultStreaming(false, result, perRowCallback, perResultCallback)
 }
 
+// cancelLocalInfileRequest sends an empty LOCAL INFILE payload packet and reads the server's
+// OK or ERR so the connection can be reused when relayFile failed before any file data was sent.
+func (c *Conn) cancelLocalInfileRequest() error {
```
lance6716: I think the logic of sending empty packet and verifying the response is OK can be unified with other path.

step 1: try to read all data and send to real backend sever
step 2: send empty packet as ending mark
step 3: verify server responds with OK

nzin-alayacare: Done in 3935302. Replaced `cancelLocalInfileRequest` and the inline OK/ERR tail with a single `sendLocalInfileContentAndAwaitResult` that always follows your three steps: validate/prepare content, stream chunks + empty terminator, read server OK/ERR once. Both the relayFile error path and the success path now share this helper.

client/conn.go
nzin-alayacare: Yes — added `prepareLocalInfileReader` in 3935302. It fully validates the reader before any packet is sent: for `io.Seeker` implementations (`*bytes.Reader`, `*os.File`) it drains to `io.Discard` and seeks back (no full-file memory buffer); for non-seekable readers it falls back to `io.ReadAll`. If validation fails, only the empty terminator is sent upstream. Added `TestExecQueryRelayLocalInfile_RelayReadError` to assert a mid-read failure sends no file data.

client/conn.go
nzin-alayacare: Good question. `relayFile` is only called when the server responds with `LocalInFile_HEADER` (0xfb). The OK/ERR branches handle the case where MySQL rejects the LOAD DATA statement before ever sending 0xfb — e.g. `local_infile` disabled, missing privileges, or syntax error. This matches the OK/ERR/LocalInFile_HEADER switch used elsewhere in the client package (`readResultStreaming`, `ExecuteMultiple`). For a dedicated LOAD DATA call we expect 0xfb on success, but handling direct ERR keeps the connection usable when the server refuses the statement.

client/conn.go
```diff
@@ -391,6 +392,163 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+}
+
+// streamLocalInfileChunks sends already-validated file bytes as LOCAL INFILE data packets,
+// followed by the terminal empty packet.
+func (c *Conn) streamLocalInfileChunks(r io.Reader) error {
+	buf := make([]byte, defaultBufferSize)
+	for {
+		n, err := r.Read(buf)
+		if n > 0 {
+			pkt := make([]byte, 4+n)
```
lance6716: We don't need the extra buffer? I think we can read into `buf[4:]` and pass `buf[:4+n]` to `WritePacket`

client/conn.go
```diff
@@ -391,6 +392,163 @@ func (c *Conn) ExecuteSelectStreaming(command string, result *mysql.Result, perR
...
+			copy(pkt[4:], buf[:n])
+			if err := c.WritePacket(pkt); err != nil {
+				return errors.Trace(err)
+			}
+		}
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
```
lance6716, line 445: If error happens halfway, I think we still need to write empty packet to avoid blocking this connection, just be defensive.

Maybe log something in this case.

client/conn.go
lance6716: Oh I didn't think of this before. Could you write it down in the notes? Just to remind future readers.

client/local_infile_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+func readServerCOMQuery(t *testing.T, pc *packet.Conn) {
+	t.Helper()
+	pkt, err := pc.ReadPacket()
+	require.NoError(t, err)
+	require.NotEmpty(t, pkt)
+	require.Equal(t, mysql.COM_QUERY, pkt[0])
+}
+
+func TestExecQueryRelayLocalInfile_DirectOK(t *testing.T) {
```
lance6716:
```suggestion
func TestExecQueryRelayLocalInfileDirectOK(t *testing.T) {
```

This naming style is more common. Please also refine other testing functions. Maybe use sub test cases?

client/local_infile_test.go
```diff
@@ -0,0 +1,201 @@
...
+	c := newLocalInfileTestConn(clientConn)
+	var requestPayload []byte
+	result, err := c.ExecQueryRelayLocalInfile("LOAD DATA LOCAL INFILE 'test.csv'", func(payload []byte) (io.Reader, error) {
+		requestPayload = append([]byte(nil), payload...)
+		return bytes.NewReader([]byte("row1\n")), nil
+	})
+	require.NoError(t, err)
+	require.NotNil(t, result)
+	require.Equal(t, mysql.LocalInFile_HEADER, requestPayload[0])
+	require.Equal(t, "test.csv", string(requestPayload[1:]))
```
lance6716: We should strip the first byte so caller can directly use filename 🤔

lance6716 (commented):
rest lgtm

client/conn.go
nzin-alayacare: when you say write it down in the notes, you mean in the godoc documentation of the function, right?

client/local_infile_test.go
nzin-alayacare: Done in 5001f50: consolidated into `TestExecQueryRelayLocalInfile` with subtests `DirectOK`, `DirectErr`, `RelaySuccess`, `RelayError`, `RelayReadError`, and `StreamWriteError`.

client/local_infile_test.go
nzin-alayacare: Done in 5001f50: `relayFile` now receives `filename []byte` (0xfb stripped). The godoc example re-wraps with `append([]byte{mysql.LocalInFile_HEADER}, filename...)` when forwarding to the downstream client.

client/conn.go
nzin-alayacare: Done in 5001f50: `streamLocalInfileChunks` now reuses a single `buf := make([]byte, 4+defaultBufferSize)` buffer, reads into `buf[4:]`, and passes `buf[:4+n]` to `WritePacket`.

client/conn.go
nzin-alayacare: Done in 5001f50: extracted `writeLocalInfileTerminator()`; on `WritePacket` failure during streaming (or read error after data was sent) we best-effort send the empty terminator and log via `log.Printf`. `sendLocalInfileContentAndAwaitResult` no longer returns early on stream errors — it still reads the server OK/ERR so the connection stays usable. Added `StreamWriteError` subtest.

client/conn.go
nzin-alayacare: Done in 5001f50: added a **Notes** section to `ExecQueryRelayLocalInfile` godoc covering LOAD DATA LOCAL INFILE intent, when `relayFile` is invoked (0xfb only), and why direct OK/ERR-before-0xfb is handled (keeps the connection usable, consistent with `readResultStreaming` / `ExecuteMultiple`).

lance6716 (approved):

PR #1165, opened 2026-07-02
client: add Stmt.ExecuteProcedureMultiResults for multi-result stored procedures

### LINKED ISSUE #1162: client: add Stmt.ExecuteProcedureMultiResults for draining stored procedure result sets
## Summary

Add `(*Stmt).ExecuteProcedureMultiResults` to the `client` package — a prepared-statement variant of `ExecuteMultiple` that drains all result sets produced by a `CALL` statement (stored procedures with `OUT` parameters and/or multiple `SELECT` result sets) and forwards each to a caller-supplied callback.

## Problem

The existing `(*Stmt).Execute` reads a single result via `readResult(true)`. When a stored procedure returns multiple result sets (e.g., `CALL my_proc(?)` where `my_proc` has `OUT` params or issues multiple `SELECT`s), the server sets `SERVER_MORE_RESULTS_EXISTS` in the status flags of intermediate results. `Execute` stops after the first result and leaves subsequent packets in the connection buffer, corrupting the connection state.

The text-protocol equivalent (`Conn.ExecuteMultiple`) handles this correctly by looping until `SERVER_MORE_RESULTS_EXISTS` is clear, but no equivalent exists for prepared statements (`COM_STMT_EXECUTE`).

This matters especially for proxy implementations: the proxy must read **and forward** every result set before the connection can be reused.

## Proposed API

```go
// StmtProcedureMultiResultForward is called once per logical result returned by
// COM_STMT_EXECUTE. When err is non-nil, res is nil. Implementations typically
// forward each result to a proxy client via server.Conn.WriteValue.
type StmtProcedureMultiResultForward func(res *mysql.Result, err error) error

// ExecuteProcedureMultiResults runs COM_STMT_EXECUTE and reads every response
// until SERVER_MORE_RESULTS_EXISTS is clear (CALL / stored procedures with
// multiple result sets). forward is invoked for each result; the loop continues
// draining even after forward returns non-nil, matching ExecuteMultiple
// semantics for proxies that must not leave unread data on the wire.
func (s *Stmt) ExecuteProcedureMultiResults(
    args []any,
    forward StmtProcedureMultiResultForward,
) (*mysql.Result, error)
```

## Implementation Sketch

```go
func (s *Stmt) ExecuteProcedureMultiResults(args []any, forward StmtProcedureMultiResultForward) (*mysql.Result, error) {
    if forward == nil {
        return nil, errors.New("forward callback cannot be nil")
    }
    if err := s.write(args...); err != nil {
        return nil, errors.Trace(err)
    }
    var forwardErr error
    for {
        res, err := s.conn.readResult(true)
        if forwardErr == nil {
            if err != nil {
                forwardErr = forward(nil, err)
            } else if res != nil {
                forwardErr = forward(res, nil)
            }
        }
        if err != nil {
            break
        }
        if res == nil || res.Status&mysql.SERVER_MORE_RESULTS_EXISTS == 0 {
            break
        }
    }
    if forwardErr != nil {
        return nil, forwardErr
    }
    rs := mysql.NewResultset(1)
    rs.Streaming = mysql.StreamingMultiple
    rs.StreamingDone = true
    return mysql.NewResult(rs), nil
}
```

## Use Case

A database jump-host proxy (like the one we are developing internally) must support Connector/J clients calling stored procedures with `OUT` parameters via prepared statements. The JDBC driver uses `COM_STMT_EXECUTE` for `CALL` and expects all result sets (including the implicit `OUT` param result set) to be forwarded correctly.

Without this API, the proxy would only forward the first result set, breaking Connector/J's `CallableStatement`.

## Relation to Existing API

- `Conn.ExecuteMultiple` (text protocol) — same semantics, different protocol.
- `(*Stmt).Execute` — single result; does not drain `SERVER_MORE_RESULTS_EXISTS`.
- `(*Stmt).ExecuteSelectStreaming` — streaming single result set.

This is the prepared-statement equivalent of `ExecuteMultiple` with a forwarding callback.

## Notes

- Pure additive change — no breaking API changes.
- Requires `CLIENT_MULTI_RESULTS` and `CLIENT_PS_MULTI_RESULTS` to be negotiated (already included in the default server capabilities since PR #1155).
- Tested in production at alayacare/castledm.

nzin-alayacare:
## Summary

Add `(*Stmt).ExecuteProcedureMultiResults` and the `StmtProcedureMultiResultForward` callback type to `client/stmt.go`.

Closes #1162

## Motivation

`(*Stmt).Execute` reads a single result via `readResult(true)` and stops. When a stored procedure returns multiple result sets (e.g., a `CALL` with `OUT` parameters or multiple `SELECT` statements inside), the server sets `SERVER_MORE_RESULTS_EXISTS` in intermediate result statuses. Calling `Execute` in this case leaves subsequent packets buffered on the connection, corrupting its state.

The text-protocol equivalent `Conn.ExecuteMultiple` already handles this correctly by looping until the flag is clear. This PR adds the same capability for prepared statements (`COM_STMT_EXECUTE`), which is the path taken by Connector/J and other JDBC drivers for `CallableStatement`.

## New API

```go
// StmtProcedureMultiResultForward is called once per logical result returned by
// COM_STMT_EXECUTE. When err is non-nil, res is nil.
type StmtProcedureMultiResultForward func(res *mysql.Result, err error) error

// ExecuteProcedureMultiResults runs COM_STMT_EXECUTE and reads every response
// until SERVER_MORE_RESULTS_EXISTS is clear.  forward is called for each
// result; the loop keeps draining even after forward returns non-nil so that
// unread packets do not corrupt the connection state.
func (s *Stmt) ExecuteProcedureMultiResults(
    args []any,
    forward StmtProcedureMultiResultForward,
) (*mysql.Result, error)
```

## Design notes

- **Drain-on-error**: the loop does not stop when `forward` returns an error because leaving unread results on the wire would corrupt the connection. Instead the forward error is saved and returned only after all results have been consumed — identical to `ExecuteMultiple` semantics.
- **Synthetic return value**: a `StreamingMultiple / StreamingDone` result is returned on success so callers can pass it to `server.Conn.WriteValue` without additional handling.
- **Nil check on callback**: returns an immediate error rather than panic.

## Notes

- Pure additive change — no existing API is modified.
- Requires `CLIENT_PS_MULTI_RESULTS` to be negotiated (included in server default capabilities since #1155).
- Tested in production at [alayacare/castledm](https://github.com/alayacare/castledm).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -548,7 +548,19 @@ END`)
 		require.NoError(s.T(), dropErr)
 	}()
 
-	stmt, err := s.c.Prepare("CALL " + procName + "()")
+	// A dedicated connection with CLIENT_MULTI_RESULTS and CLIENT_PS_MULTI_RESULTS
+	// is required; these are optional capabilities not set on the shared suite conn.
+	addr := fmt.Sprintf("%s:%s", *test_util.MysqlHost, s.port)
+	mc, err := Connect(addr, *testUser, *testPassword, *testDB, func(conn *Conn) error {
+		if err := conn.SetCapability(mysql.CLIENT_MULTI_RESULTS); err != nil {
+			return err
+		}
+		return conn.SetCapability(mysql.CLIENT_PS_MULTI_RESULTS)
+	})
+	require.NoError(s.T(), err)
+	defer mc.Close()
+
+	stmt, err := mc.Prepare("CALL " + procName + "()")
 	require.NoError(s.T(), err)
 	defer stmt.Close()
 
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -531,7 +531,7 @@ INSERT INTO field_value_test VALUES (
 	}
 }
 
-func (s *clientTestSuite) TestStmt_ExecuteProcedureMultiResults() {
+func (s *clientTestSuite) TestStmt_ProcedureMultiResult() {
 	const procName = "go_mysql_test_multi_result"
 
 	_, err := s.c.Execute("DROP PROCEDURE IF EXISTS " + procName)
@@ -565,7 +565,7 @@ END`)
 	defer stmt.Close()
 
 	var resultSets int
-	result, err := stmt.ExecuteProcedureMultiResults(func(res *mysql.Result, err error) error {
+	err = stmt.ExecuteProcedureMultiResults(func(res *mysql.Result, err error) error {
 		require.NoError(s.T(), err)
 		if res != nil && res.HasResultset() {
 			resultSets++
@@ -577,9 +577,6 @@ END`)
 		return nil
 	})
 	require.NoError(s.T(), err)
-	require.NotNil(s.T(), result)
-	require.True(s.T(), result.StreamingDone)
-	require.Equal(s.T(), mysql.StreamingMultiple, result.Streaming)
 	require.Equal(s.T(), 2, resultSets)
 }
 
@@ -596,8 +593,8 @@ func (s *clientTestSuite) TestLongPassword() {
 	require.NoError(s.T(), err)
 }
 
-func TestExecuteProcedureMultiResults_NilCallback(t *testing.T) {
+func TestStmtProcedureMultiResultNilForward(t *testing.T) {
 	s := &Stmt{conn: &Conn{}}
-	_, err := s.ExecuteProcedureMultiResults(nil)
+	err := s.ExecuteProcedureMultiResults(nil)
 	require.ErrorContains(t, err, "forward callback cannot be nil")
 }
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -34,13 +34,7 @@ func (s *clientTestSuite) SetupSuite() {
 	s.c, err = Connect(addr, *testUser, *testPassword, "", func(conn *Conn) error {
 		// test the collation logic, but this is essentially a no-op since
 		// the collation set is the default value
-		if err := conn.SetCollation(mysql.DEFAULT_COLLATION_NAME); err != nil {
-			return err
-		}
-		if err := conn.SetCapability(mysql.CLIENT_MULTI_RESULTS); err != nil {
-			return err
-		}
-		return conn.SetCapability(mysql.CLIENT_PS_MULTI_RESULTS)
+		return conn.SetCollation(mysql.DEFAULT_COLLATION_NAME)
 	})
 	require.NoError(s.T(), err)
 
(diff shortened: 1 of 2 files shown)


client/client_test.go
```diff
@@ -531,6 +531,58 @@ INSERT INTO field_value_test VALUES (
 	}
 }
 
+func (s *clientTestSuite) TestStmt_ExecuteProcedureMultiResults() {
```
lance6716: Update the naming style as previous PRs.

client/stmt.go
```diff
@@ -49,6 +49,53 @@ func (s *Stmt) ExecuteSelectStreaming(result *mysql.Result, perRowCb SelectPerRo
...
+			break
+		}
+	}
+	if forwardErr != nil {
+		return nil, forwardErr
+	}
+	rs := mysql.NewResultset(0)
+	rs.Streaming = mysql.StreamingMultiple
+	rs.StreamingDone = true
+	return mysql.NewResult(rs), nil
```
lance6716: Does the dummy result is needed in your application? Maybe we don't need that return value

client/client_test.go
```diff
@@ -531,6 +531,55 @@ INSERT INTO field_value_test VALUES (
 	}
 }
 
+func (s *clientTestSuite) TestStmt_ProcedureMultiResult() {
```
lance6716: I mean TestStmtProcedureMultiResult 😂

client/stmt.go
```diff
@@ -49,6 +49,49 @@ func (s *Stmt) ExecuteSelectStreaming(result *mysql.Result, perRowCb SelectPerRo
...
+			}
+		}
+		if err != nil {
+			break
+		}
+		if res == nil || res.Status&mysql.SERVER_MORE_RESULTS_EXISTS == 0 {
+			break
+		}
+	}
+	if forwardErr != nil {
```
lance6716: 1. This function will not return error if `forward` accepts the `err` but return a nil `forwardErr`. Not sure if this is expected. Maybe add this behaviour to function commment
2. 

```
	if forwardErr != nil {
		return forwardErr
	}
	return nil
```

can be simplified to `return forwardErr`

lance6716 (commented):
rest lgtm

client/stmt.go
nzin-alayacare: :-), let me address that

lance6716 (approved):

PR #1166, opened 2026-07-03
Fix and test out of range

desdic:
When sending a pluginname, username without a null termination it panics like
panic: runtime error: slice bounds out of range [58:57]

### DIFF
--- a/server/handshake_resp.go
+++ b/server/handshake_resp.go
@@ -28,7 +28,9 @@ func (c *Conn) readHandshakeResponse() error {
 		return err
 	}
 
-	pos = c.readPluginName(data, pos)
+	if pos, err = c.readPluginName(data, pos); err != nil {
+		return err
+	}
 
 	cont, err := c.handleAuthMatch()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/handshake_resp.go
```diff
@@ -109,9 +111,13 @@ func (c *Conn) decodeFirstPart(data []byte) (newData []byte, pos int, err error)
 }
 
 func (c *Conn) readUserName(data []byte, pos int) (int, error) {
+	idx := bytes.IndexByte(data[pos:], 0x00)
+	if idx < 0 {
+		return 0, mysql.NewDefaultError(mysql.ER_HANDSHAKE_ERROR)
```
lance6716, line 119: Hi, did you check the official mysql-server use this kind of error code?

lance6716 (commented):
rest lgtm

server/handshake_resp.go
desdic: No :( I assumed it was the same as decode error due to not maintaining protocol

desdic: It seems newer MySQL code sometimes uses ER_MALFORMED_PACKET (1835, "Malformed communication packet") for length/parse errors. It's defensible, but within this library
  ER_HANDSHAKE_ERROR is the convention for handshake-response parse failures so that is why I think its ok .. if I'm wrong I'll change it

lance6716: If ER_HANDSHAKE_ERROR is the old version behaviour, I think it's OK. Will check mysql-server code soon. (Can you provide some reference from https://github.com/mysql/mysql-server ❤️ )

server/handshake_resp.go
```diff
@@ -109,9 +111,13 @@ func (c *Conn) decodeFirstPart(data []byte) (newData []byte, pos int, err error)
 }
 
 func (c *Conn) readUserName(data []byte, pos int) (int, error) {
+	idx := bytes.IndexByte(data[pos:], 0x00)
```
lance6716, line 117: [P1] `readUserName` still slices `data[pos:]` without checking `pos <= len(data)`. `decodeFirstPart` can return `pos == 32` even for packets shorter than that fixed header length, and `readHandshakeResponse` then calls `readUserName(data, pos)`, so a truncated handshake response can still panic here instead of returning `ER_HANDSHAKE_ERROR`.

Disclosure: I am Codex, an AI coding agent, not a human reviewer.

desdic: fixed

server/handshake_resp.go
desdic: found several places in https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sql_authentication.cc when doing auth

lance6716 (approved):

PR #1173, opened 2026-07-29
replication: Fix RenderJSONAsMySQLText floating point corruption

morgo:
In https://github.com/go-mysql-org/go-mysql/pull/1147 an option was introduced (by me) to render JSON as text to avoid round-tripping through golang and corrupting JSON values.

That PR has a corruption bug itself: floating point values get rendered in a spelling MySQL would never produce, and MySQL re-reads some of those as a different value.

```sql
-- The same JSON document, stored two ways. These SHOULD be identical:
SELECT
  CRC32(CAST(CAST('{"v": 602214075999999987023872.0}' AS JSON) AS CHAR)) AS via_go_mysql_text, -- 4146900613
  CRC32(CAST(JSON_OBJECT('v', 6.02214076e23)          AS CHAR)) AS via_sql_double;             -- 1416476187

-- Why: go-mysql renders the double as a 24-digit expansion, which MySQL re-reads as a different value:
SELECT CAST(CAST('{"v": 602214075999999987023872.0}' AS JSON) AS CHAR);
--> {"v": 6.0221407600000005e23}    -- MySQL itself stores {"v": 6.02214076e23}
```

The fix is to render doubles exactly the way MySQL does. That's a somewhat complicated fix because it requires bug compatibility with MySQL itself.

The significant digits are the shortest round-trip form, which golang's `strconv` already produces identically. What differs is the fixed-vs-scientific choice and the exponent spelling (`1.5e-5` vs golang's `1.5e-05`). The rule is derived in the comment on `formatMySQLDouble`, and I've checked it byte-for-byte against 8.0, 8.4 and 9.x, so it isn't specific to one version.

Scope: this makes our rendering match the server, which is what stops us from introducing differences of our own. It does not make text-mediated round-trips lossless, because MySQL's JSON text parser misrounds some 16-17 significant-digit doubles by 1 ulp regardless of how they're spelled (bugs [#116160](https://bugs.mysql.com/bug.php?id=116160), [#112904](https://bugs.mysql.com/bug.php?id=112904)):

```sql
SELECT CAST(CAST('{"v": 1.0000013331624825e-276}' AS JSON) AS CHAR);
--> {"v": 1.0000013331624823e-276}   -- last digit changed, and this is MySQL's own rendering
```

I went with matching the server anyway — the alternative is emitting exact decimal expansions, which is worse across the board and which MySQL rejects outright at the top of the range ("Number too big to be stored in double" for `DBL_MAX`).

There is a second corruption bug in #1147 affecting temporal values, which I'll send separately as it touches code outside of the text rendering path.

How we detected this:

We use the go-mysql library to perform schema changes with Spirit. When making changes containing JSON documents we would sometimes see checksum mismatch errors. The checksum process can also isolate the specific rows that mismatch, leading to these discoveries.

After applying the fixes, we've been running without checksum mismatches again.

### DIFF
--- a/replication/json_mysql_text.go
+++ b/replication/json_mysql_text.go
@@ -16,10 +16,9 @@ import (
 // stays unquoted; etc.) and preserves the JSONB key order.
 //
 // Caveats:
-//   - The output is type-faithful, not byte-identical to MySQL's
-//     "SELECT json_col" form. Inter-token whitespace is compact (no
-//     space after ',' or ':') and floating-point text differs in some
-//     exponent/precision corner cases (see jsonMySQLDouble).
+//   - Inter-token whitespace is compact (no space after ',' or ':'),
+//     unlike MySQL's "SELECT json_col" form. DOUBLE scalars are
+//     byte-identical to MySQL's own text (see jsonMySQLDouble).
 //   - NEWDECIMAL is the one tag that cannot be preserved on text
 //     round-trip: MySQL's JSON text grammar has no decimal literal, so
 //     re-inserting the unquoted number yields a JSON DOUBLE, not the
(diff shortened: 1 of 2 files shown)


replication/json_mysql_text.go
```diff
@@ -99,16 +95,70 @@ func (o jsonObject) MarshalJSON() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// formatMySQLDouble returns the bytes MySQL emits for a JSON DOUBLE.
+// The rule below was derived empirically and checked byte-for-byte
+// against MySQL 8.0, 8.4 and 9.x over a corpus spanning the format
```
lance6716, line 100: (will review later)

You mean all tried MySQL versions have this behaivour, or the old versions are not so it applies for partial version?

morgo: I tried all MySQL major versions and they are all broken.

(They are not always the latest point release, but I have an 8.0, 8.4, 9.7 installed locally and also use a matrix of versions in Spirit's CI).

replication/json_mysql_text.go
```diff
@@ -51,20 +50,17 @@ func (n jsonRawNumber) MarshalJSON() ([]byte, error) {
...
 //
-// The output is NOT guaranteed to be byte-identical to MySQL's
-// my_gcvt-formatted text: Go's 'g' verb emits exponents as e.g.
-// "1.5e-05" where MySQL emits "1.5e-5", and the integer/scientific
-// crossover threshold differs. Binary round-trip through a MySQL JSON
-// column is unaffected (the same float64 produces the same JSONB
-// DOUBLE bytes); only the visible text form may differ.
+// This does not make the round-trip lossless. MySQL's JSON text parser
+// misrounds some 16-17 significant-digit doubles by 1 ulp whatever the
+// spelling (MySQL bugs #116160 and #112904), including values MySQL
```
lance6716, line 61: Just in case in the future MySQL will fix these bugs, we can use the server version to switch the behavior then.

lance6716 (approved):
(I'll merge it after My local AI review finished)

PR #1174, opened 2026-07-29
replication: render zero JSON temporal opaques with 6-digit fractions

morgo:
This is the second of the two corruption bugs in https://github.com/go-mysql-org/go-mysql/pull/1147 (an option I added to render JSON as text to avoid round-tripping through golang and corrupting JSON values). The floating point one is https://github.com/go-mysql-org/go-mysql/pull/1173; this one is temporal values, and it's a straight forward fix by removing code that was incorrect.

```sql
-- The same JSON document, stored two ways. These SHOULD be identical:
SELECT
  CRC32(CAST(CAST('{"t": "00:00:00"}' AS JSON) AS CHAR))      AS via_go_mysql_text, -- 2609006976
  CRC32(CAST(JSON_OBJECT('t', CAST('00:00:00' AS TIME)) AS CHAR)) AS via_sql_time;  -- 2560132030

-- Why: MySQL gives a zero TIME the same 6-digit fraction as any other value:
SELECT CAST(JSON_OBJECT('t', CAST('00:00:00' AS TIME)) AS CHAR);
--> {"t": "00:00:00.000000"}     -- go-mysql rendered {"t": "00:00:00"}
```

`decodeTime` and `decodeDateTime` had a `v == 0` shortcut that returned a fraction-free string. Every non-zero value already fell through to the general path and picked up a 6-digit fraction, so zero was the only value rendered inconsistently, and the shortcut disagreed with the server:

| | MySQL | go-mysql before |
|---|---|---|
| `00:00:00` | `"00:00:00.000000"` | `"00:00:00"` |
| `01:02:03` | `"01:02:03.000000"` | `"01:02:03.000000"` |
| `0000-00-00 00:00:00` | `"0000-00-00 00:00:00.000000"` | `"0000-00-00 00:00:00"` |
| `0000-00-00` | `"0000-00-00"` | `"0000-00-00"` |

Deleting the shortcuts is the whole fix - the general formatting paths already produce the server's exact output for `v == 0`.

One thing to flag: the shortcut wasn't behind the `RenderJSONAsMySQLText` flag, so this also changes the legacy render of zero temporals for anyone not using the option (`"00:00:00"` becomes `"00:00:00.000000"`). I think that's the right call, since it's the same string the legacy path already produced for every non-zero value and it's what the server itself returns, but I'm happy to gate it on the flag if you'd rather not touch existing behaviour.

For testing I added `TestJSONTemporalServerParity`, which inserts zero and non-zero temporals into a JSON column, reads the rows back through the binlog, and compares our rendering against the server's own `CAST(c AS CHAR)` for the same row. It fails without this fix -- and can be reproduced on 8.0, 8.4, 9.x.

How we detected this:

We use the go-mysql library to perform schema changes with Spirit. When making changes containing JSON documents we would sometimes see checksum mismatch errors. The checksum process can also isolate the specific rows that mismatch, leading to these discoveries.

After applying the fixes, we've been running without checksum mismatches again.

### DIFF
--- a/replication/json_binary.go
+++ b/replication/json_binary.go
@@ -535,10 +535,6 @@ func (d *jsonBinaryDecoder) decodeDecimal(data []byte) any {
 func (d *jsonBinaryDecoder) decodeTime(data []byte) any {
 	v := d.decodeInt64(data)
 
-	if v == 0 {
-		return "00:00:00"
-	}
-
 	sign := ""
 	if v < 0 {
 		sign = "-"
@@ -556,12 +552,6 @@ func (d *jsonBinaryDecoder) decodeTime(data []byte) any {
 
 func (d *jsonBinaryDecoder) decodeDateTime(data []byte, isDate bool) any {
 	v := d.decodeInt64(data)
-	if v == 0 {
-		if isDate {
-			return "0000-00-00"
-		}
-		return "0000-00-00 00:00:00"
-	}
 
 	// handle negative?
 	if v < 0 {
--- a/replication/json_mysql_text_test.go
+++ b/replication/json_mysql_text_test.go
@@ -2,13 +2,19 @@ package replication
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
+	"fmt"
 	"math"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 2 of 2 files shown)


lance6716 (approved):
>  I think that's the right call, since it's the same string the legacy path already produced for every non-zero value

Agreed.

PR #1181, opened 2026-08-05
replication: tag unmarshalable BinlogSyncerConfig fields with json:"-"

### LINKED ISSUE #1180: BinlogSyncer logs its config as a JSON error placeholder
**Version:** go-mysql v1.16.0 (and earlier — the log call is unchanged for several releases), Go 1.25, MySQL 8.0.

`NewBinlogSyncer` logs the whole config with `slog.Any("config", cfg)` (`replication/binlogsyncer.go:232`). `BinlogSyncerConfig` holds func-typed fields — `Option`, `Dialer`, `RowsEventDecodeFunc` — and `encoding/json` rejects those **by type**, whether or not they are nil. So under a JSON handler every syncer start emits:

```json
{"level":"INFO","msg":"create BinlogSyncer","config":"!ERROR:json: unsupported type: func(*client.Conn) error"}
```

Two consequences: no configuration is actually logged, and the record carries an `ERROR` substring at INFO level, which trips alerting that scrapes logs for that string.

Incidentally, the `cfg.Password` clear-and-restore around the log call is currently doing nothing, because the marshal fails before any field is reached.

**What I'd propose** — `slog.LogValuer` on the config, which keeps the call site as it is and lets the password never enter the record at all rather than being blanked out around it:

```go
func (c BinlogSyncerConfig) LogValue() slog.Value {
	return slog.GroupValue(
		slog.Uint64("server_id", uint64(c.ServerID)),
		slog.String("flavor", c.Flavor),
		slog.String("host", c.Host),
		slog.Uint64("port", uint64(c.Port)),
		slog.String("user", c.User),
		slog.Bool("tls", c.TLSConfig != nil),
		slog.Duration("heartbeat_period", c.HeartbeatPeriod),
		slog.Duration("read_timeout", c.ReadTimeout),
		slog.Int("max_reconnect_attempts", c.MaxReconnectAttempts),
		// ... and the remaining operational fields
	)
}
```

Other options, if you'd rather not add a method to the exported type: `json:"-"` on the func fields, a `MarshalJSON`, or logging a handful of fields inline at the call site.

I have a working patch with a test (the test fails with exactly the placeholder above and passes with the method). Happy to send it, or to follow whichever approach you prefer — the field set in particular is your call.

kremniov:
Fixes #1180. Uses JSON annotations rather than a `LogValue` method, as @lance6716 asked for in the issue: a tag can't go stale when a new field is added.

`NewBinlogSyncer` logs the whole config with `slog.Any("config", cfg)`. `encoding/json` rejects func-typed fields by type, whether they are nil or not, so under a JSON handler every syncer start printed this and nothing else:

```json
{"level":"INFO","msg":"create BinlogSyncer","config":"!ERROR:json: unsupported type: func(*client.Conn) error"}
```

So no config was logged at all, and the record carried the word `ERROR` at INFO level, which is enough to set off alerting that greps logs for it.

Tagged fields: `Option`, `Dialer`, `RowsEventDecodeFunc`, `TableMapOptionalMetaDecodeFunc`, and two more that fail the same marshal for the same reason. `tls.Config` has func fields of its own, so a non-nil `TLSConfig` fails with `func() time.Time`. `SynchronousEventHandler` is an interface, so whatever the caller passes decides.

One side effect: the `cfg.Password` clear and restore around the log call now does something, because the marshal gets as far as that field.

Note that the record no longer says whether TLS is configured. Tell me if you want a bool for that at the call site.

Tested: the new test calls `NewBinlogSyncer` with all six of those fields set and checks that the record has no `!ERROR` placeholder, no password, and real values. It fails before the change and passes after. `./replication/ ./packet/ ./mysql/ ./compress/ ./utils/` pass. I did not run the suites that need MySQL on 127.0.0.1:3306 (`canal`, `client`, `server`, `dump`, `schema`); CI covers those.

### DIFF
--- a/replication/binlogsyncer.go
+++ b/replication/binlogsyncer.go
@@ -53,7 +53,7 @@ type BinlogSyncerConfig struct {
 	RawModeEnabled bool
 
 	// If not nil, use the provided tls.Config to connect to the database using TLS/SSL.
-	TLSConfig *tls.Config
+	TLSConfig *tls.Config `json:"-"`
 
 	// Use replication.Time structure for timestamp and datetime.
 	// We will use Local location for timestamp and UTC location for datetime.
@@ -143,17 +143,17 @@ type BinlogSyncerConfig struct {
 
 	// Option function is used to set outside of BinlogSyncerConfig， between mysql connection and COM_REGISTER_SLAVE
 	// For MariaDB: slave_gtid_ignore_duplicates、skip_replication、slave_until_gtid
-	Option func(*client.Conn) error
+	Option func(*client.Conn) error `json:"-"`
 
 	// Set Logger
 	Logger *slog.Logger
 
 	// Set Dialer
-	Dialer client.Dialer
+	Dialer client.Dialer `json:"-"`
 
-	RowsEventDecodeFunc func(*RowsEvent, []byte) error
+	RowsEventDecodeFunc func(*RowsEvent, []byte) error `json:"-"`
 
-	TableMapOptionalMetaDecodeFunc func([]byte) error
+	TableMapOptionalMetaDecodeFunc func([]byte) error `json:"-"`
 
 	DiscardGTIDSet bool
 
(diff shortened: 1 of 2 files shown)


lance6716 (approved):
I think we can display TLS status in config later. Currently seems no one is requiring it

PR #1182, opened 2026-08-11
Add mariadb integration test setup

nkondratyev:
- Add make targets to start, test, and stop a MariaDB 11.8.8 container.
- Pass MYSQL_FLAVOR=mariadb to the complete test suite while preserving
  the existing behavior when the variable is not set.
- Configure canal with the selected server flavor so it uses MariaDB GTID
  parsing instead of querying the MySQL-only GTID_EXECUTED variable.
- Run MariaDB replication tests on the configured test port and skip
  scenarios that explicitly require MySQL protocol behavior or GTIDs.
- Use SHOW BINLOG STATUS and SHOW SLAVE HOSTS for MariaDB because its
  status commands differ from recent MySQL versions.
- Skip MySQL-specific CAST(... AS JSON) expressions because MariaDB 11.8
  does not accept JSON as a CAST target and stores JSON as validated text.
- Skip the binary JSON parity test because it verifies MySQL's binary JSON
  representation, which MariaDB does not use.
- Always provide a scan destination for the MariaDB Ignored column
  returned by SHOW INDEX when the MySQL Visible column is absent.
- Normalize SHOW FULL COLUMNS Extra metadata before matching because
  MariaDB returns values such as "on update current_timestamp()" with
  different casing from MySQL.
- Adjust the default-expression expectation because MariaDB reports the
  expression in Default without adding MySQL's DEFAULT_GENERATED marker
  to Extra.
- Skip the multi-valued index test because this index syntax is specific
  to MySQL and is not supported by MariaDB.
- Add MariaDB test matrix to CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,39 @@ jobs:
           go test $(go list ./... | grep -v canal | grep -v dump)
           # go test $(go list ./... | grep canal | grep -v dump)
 
+  mariadbtest:
+    strategy:
+      fail-fast: false
+      matrix:
+        mariadb_version:
+          - 10.11.18
+          - 11.4.12
+          - 11.8.8
+    name: Tests with MariaDB ${{ matrix.mariadb_version }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: stable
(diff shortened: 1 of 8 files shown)


lance6716:
Hi, I have triggered a round of CI. Please check the result. I'll take a look later.

nkondratyev:
> Hi, I have triggered a round of CI. Please check the result. I'll take a look later.

Hi, I fixed the tests for MariaDB 10.11

lance6716 (approved):

PR #1183, opened 2026-08-17
mysql: distinguish empty text values from NULL in text resultsets

MOON-CLJ:
`BuildSimpleTextResultset` already calls `fieldType(value)` before encoding each cell. This is the semantic classification of the original value: `nil` maps to `MYSQL_TYPE_NULL`, while `string` and `[]byte` map to `MYSQL_TYPE_VAR_STRING`.

The previous implementation checked `b == nil` after calling `FormatTextValue`. At that point, an empty string may also be represented by a nil, zero-length byte slice, so `b == nil` is not sufficient to distinguish it from SQL `NULL`.

| Go value | Inferred MySQL type | Previous encoding | New encoding |
|---|---|---|---|
| `""` | `MYSQL_TYPE_VAR_STRING` | `0xfb` (NULL) | `0x00` (empty string) |
| `[]byte{}` | `MYSQL_TYPE_VAR_STRING` | `0x00` (empty string) | `0x00` (empty string) |
| `nil` | `MYSQL_TYPE_NULL` | `0xfb` (NULL) | `0xfb` (NULL) |

In the MySQL text protocol, `0xfb` is the NULL marker, while `0x00` is a length-encoded string with length zero.

The change therefore uses the already inferred per-value type:

```go
if typ == MYSQL_TYPE_NULL {
    row = append(row, 0xfb)
    continue
}

### DIFF
--- a/mysql/resultset_helper.go
+++ b/mysql/resultset_helper.go
@@ -212,17 +212,17 @@ func BuildSimpleTextResultset(names []string, values [][]any) (*Resultset, error
 					return nil, errors.Errorf("row types aren't consistent")
 				}
 			}
+			if typ == MYSQL_TYPE_NULL {
+				// NULL value is encoded as 0xfb here (without additional info about length)
+				row = append(row, 0xfb)
+				continue
+			}
+
 			b, err = FormatTextValue(value)
 			if err != nil {
 				return nil, errors.Trace(err)
 			}
-
-			if b == nil {
-				// NULL value is encoded as 0xfb here (without additional info about length)
-				row = append(row, 0xfb)
-			} else {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by lance6716 — COMMENTED** [review]
```
 Error: replication/binlogsyncer_test.go:90:1: File is not properly formatted (gofumpt)
  func (*deadlinelessConn) SetDeadline(time.Time) error { return errors.New("deadline not supported") }
  ^
  Error: mysql/errname.go:4:1: File is not properly formatted (goimports)
  	ER_HASHCHK:                                       "hashchk",
  ^
```

Seems there are some unrelated linter errors 🤔

lance6716 (approved):

PR #1185, opened 2026-08-19
Add boolean support to result set formatting

tablebird:
Add handling for bool values throughout result set formatting. Text formatting now uses utils.BoolToByteSlice ("true"/"false"). Binary formatting encodes bool as 1/0 via toBinaryBool and FormatBinaryValue. fieldType maps bool to MYSQL_TYPE_TINY and formatField sets appropriate charset/flags. Also add BoolToByteSlice helper in utils/zeroalloc.go.


lance6716 (commented):
The linter is run based on go 1.27, if you can't reproduce the errors, you can try to upgrade golang.

lance6716 (commented):
Hi, please add some unit test to cover this behaviour.

Rest lgtm

lance6716 (approved):

PR #1190, opened 2026-09-07
mysql: return nil from AsString for NULL fields

### LINKED ISSUE #1189: client.Execute  获得了一个不应该返回的值
版本 v1.16.0.  数据库: [MariaDb](mariadb:10.6)

Image

Image

"github.com/go-mysql-org/go-mysql/client" 的 Execute 和 database/sql  github.com/go-sql-driver/mysql 的结果一至

这是 DDL
create table act_prize
(
    id            bigint         default 0        not null comment '奖励ID'
        primary key,
    name          varchar(200)                    null comment '奖品名称',
    prize_type    tinyint        default 0        not null comment '奖品类型',
    prize_mode    tinyint        default 0        not null comment '奖励方式',
    prize_num     decimal(22, 6) default 0.000000 not null comment '奖品值',
    max_value     decimal(22, 6) default 0.000000 not null comment '最大值',
    flow_multiple float                           null comment '流水倍数',
    related_obj   bigint                          null comment '关联对象',
    prize_desc    bigint                          null comment '奖品描述',
    remark        varchar(512)                    null comment '备注',
    create_time   timestamp(3)                    null comment '创建时间',
    delete_time   timestamp(3)                    null comment '删除时间',
    is_deleted    tinyint(1)     default 0        not null comment '是否删除'
)
    comment '奖品配置';

获使用 client.Execute 获得一个错误的值
使用 sql.Query 没有返回值

这是源码
```
// GetPartitionType 动态查询 MySQL 上游表的 RANGE 分区字段
func (c *Canal) GetPartitionType(ctx context.Context, schema Schema, table Table) (string, error) {
	// 查询该表是否拥有 RANGE 类型的分区，并找出分区键 (COLUMN_NAME)
	query := `SELECT PARTITION_METHOD FROM INFORMATION_SCHEMA.PARTITIONS WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ? LIMIT 1;`

	result, err := c.conn.Execute(query, string(schema), string(table))
	if err != nil {
		return "", err
	}
	defer result.Close()

	// 如果没有查询到分区记录，说明是普通表，不启用分区
	if len(result.Values) == 0 {
		return "", nil
	}
	ret := string(result.Values[0][0].AsString())
	println("result", ret)

	////////////////////////////////////////////
	rows, err := c.sql.Query(query, string(schema), string(table))
	if err != nil {
		return "", err
	}
	defer rows.Close()
	for rows.Next() {
		var partitionMethod *string
		err := rows.Scan(&partitionMethod)
		if err != nil {
			return "", err
		}
		if partitionMethod != nil {
			ret = *partitionMethod // Unwrap the pointer
			println("rows", *partitionMethod)
		}
	}
	return ret, nil
}
```

jakezwang:
`FieldValue.AsString()` can return bytes from a previous row when a result set is reused and the new value is NULL. For example, reading `'id'` followed by NULL can leave `Value()` returning nil while `AsString()` still returns `"id"`.

Return nil from `AsString()` for `FieldValueTypeNull`, while retaining the underlying buffer for reuse. Add regression coverage for text and binary rows, including transitions through NULL and an empty string.

Verified with `go test -race ./mysql -count=1`. Also reproduced the original behavior with v1.16.0 and MariaDB 10.6.28, and verified the fix against both query protocols.

Fixes #1189.

### DIFF
--- a/mysql/field.go
+++ b/mysql/field.go
@@ -203,7 +203,11 @@ func (fv *FieldValue) AsFloat64() float64 {
 	return utils.Uint64ToFloat64(fv.value)
 }
 
+// AsString returns the string value as bytes, or nil for NULL.
 func (fv *FieldValue) AsString() []byte {
+	if fv.Type == FieldValueTypeNull {
+		return nil
+	}
 	return fv.str
 }
 
--- a/mysql/rowdata_test.go
+++ b/mysql/rowdata_test.go
@@ -1,0 +1,54 @@
+package mysql
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRowDataParseNullAfterString(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		binary bool
+		str    RowData
+		null   RowData
+		empty  RowData
+	}{
+		{
+			name:  "text",
+			str:   RowData{2, 'i', 'd'},
+			null:  RowData{0xfb},
+			empty: RowData{0},
+		},
+		{
+			name:   "binary",
+			binary: true,
+			str:    RowData{0, 0, 2, 'i', 'd'},
+			null:   RowData{0, 4}, // The first column uses bit 2 of the NULL bitmap.
+			empty:  RowData{0, 0, 0},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			fields := []*Field{{Type: MYSQL_TYPE_VAR_STRING}}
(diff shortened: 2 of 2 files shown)


jakezwang:
@lance6716 Could you take a look at this fix for #1189? It prevents `AsString()` from returning a previous row's bytes for NULL, with regression tests for both text and binary protocols. Local race tests pass. CI appears to be waiting for maintainer approval. Could you approve the workflow and review when you have a chance? Thanks!

lance6716 (approved):

PR #1193, opened 2026-09-22
Fix query attribute lifecycle and prepared statement encoding

IuraHD:
Query attributes set for one execution currently leave zeroed entries on the connection and mutate the caller's slice. Prepared statements also omit attribute values with zero or only NULL parameters, and advertise parameter-count support without negotiating it.

Reset the connection's attribute slice after execution, encode attributes when supported even without non-NULL parameters, and gate the parameter-count flag on the negotiated capability.

Six in-memory packet regressions fail on current master and pass with this change. Validation: `go test ./client -run TestQueryAttribute -count=1` and `go test ./mysql ./utils`, with networking disabled. The server-dependent client integration suite was not run.

Replaces [#1192](https://github.com/go-mysql-org/go-mysql/pull/1192); earlier review discussion remains there.


lance6716:
(I'll close and reopen to trigger CI)

client/conn.go
```diff
@@ -824,7 +824,8 @@ func (c *Conn) StatusString() string {
 
 // SetQueryAttributes sets the query attributes to be send along with the next query
```
lance6716: I think we can add the clone behaviour to function comments, and remove the comment at line 827

lance6716 (commented):
please check the linter CI and fix it.

rest lgtm

